PR #25, opened 2015-02-18
decoding html entities in 'title' attribute value

hutnikau:
In the value of the attribute (`title` in this case) is not allowed to use ampersand.


jbout:
Please do not use htmlentities to encode xml attribute values.

As per specifications: http://www.xml.com/axml/target.html#NT-AttValue the only characters not allowed in an XML attribute value are the quotation marks, & and <. You could consider using DOMDocument to encode the attribute values

hutnikau:
Now the XML test file is created using `DOMDocument`. Also I have created the unit test for validate that attribute values will be properly encoded.

jbout:
Looking good, but still requires the cleanup of the created resource

hutnikau:
Done.

PR #36, opened 2015-03-25
Fix rubric block editor

krampstudio:
- replace item's html editor by only ckeditor
 - image management is still not supported


ssipasseuth:
It works fine !

However I have a few comments on the style of the rubric block editor :
* the list element has a weird background (that disappear in the delivery)
* the section of rubricBlock is marked in red, looks a bit like an error (won't a lighter $info blue be better?)
* after reloading the test, previously created rubricBlock are hidden by default (I thought that my rubric block didn't save for a while)

If it is not too complicated, is it worth considering these changes ?

bugalot:
If you need any help to find out why it "disappears in the delivery", just let me know!

ssipasseuth:
Sorry, I meant the background color only is not there in the delivery (although it is in the authoring) :)
So I believe it is just a css issue.

bugalot:
Thanks for the info :) :+1:

krampstudio:
No it's not difficult to change. I'll do it.

bugalot:
It's working pretty well guys! The only glitch is on my side were the base.css stylesheets for rubric blocks does not really align with what the authoring tool shows.

I will make some changes on the delivery side to make it more homogeneous.

Cheers,
Jérôme

PR #54, opened 2015-04-27
Change way to hide item iframe

hutnikau:
If Iframe has `display: none;` css rule then for the nested elements can't be calculated it dimensions (width, height). But for select2 plugin is necessary to get width of initial `<select>` element to create widget with appropriate width.


bugalot:
Good catch! OK for me!

PR #68, opened 2015-06-11
modify report data treatment

antoinerobin:


antoinerobin:
https://github.com/oat-sa/extension-tao-itemqti/pull/296

jbout:
currently too much code redundancy with QTI item import, should not require adaption for shared stimulus

PR #86, opened 2015-07-01
Tao 862 test taker review screen

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-862

An update is needed to be able to use the new test taker review screen.
But, you will need to manually enable the option in the file `config/taoQtiTest/testRunner.conf.php`.

There are some interesting options you can change:
- **progress-indicator** : allows to change the behavior of the progress bar (can be `percentage`or `position`
- **test-taker-review** : enables/disables the test taker review screen
- **test-taker-review-region** : set the region on which display the review screen (can be `left`or `right`)
- **test-taker-review-section-only** : when set to `true`, the review screen only display the current section
- **test-taker-review-prevents-unseen** : when set to `true`, only the viewed item can be accessed, the unseen items are simply disabled

On the review screen you can freely navigate between parts, sections and items, but only in non linear test.

To mark an item for review, just click on the icon before its name in the review screen.


actions/class.TestRunner.php
```diff
@@ -289,6 +294,61 @@ public function index() {
...
+     *
+     */
+    public function jumpTo() {
+        $this->beforeAction();
+        $session = $this->getTestSession();
+
+        try {
+            $session->jumpTo(intval($this->getRequestParameter('position')));
+
+            if ($session->isRunning() === true && taoQtiTest_helpers_TestRunnerUtils::isTimeout($session) === false) {
```
krampstudio: is triple = comparison required ?
JS habits ;)

views/js/controller/runtime/testRunner.js
```diff
@@ -98,6 +99,36 @@ define([
                 iframeNotifier.parent('unloading');
             },
 
+            jump: function(position) {
```
krampstudio: jsDoc is missing

views/js/controller/runtime/testRunner.js
```diff
@@ -98,6 +99,36 @@ define([
...
+                this.disableGui();
+                var that = this;
```
krampstudio: Our convention is to use `self`

views/js/controller/runtime/testRunner.js
```diff
@@ -98,6 +99,36 @@ define([
                 iframeNotifier.parent('unloading');
             },
 
+            jump: function(position) {
+                this.disableGui();
```
krampstudio: variable to be declared first (we want hoisting to be reflected)

actions/class.TestRunner.php
jsconan: This part of code was copied from other navigation methods, and I assumed this kind of comparison was needed in that case

views/js/runner/testReview.js
```diff
@@ -0,0 +1,610 @@
...
+
+            // build the component structure and inject it into the DOM
+            this.$container = $(element);
+            insertMethod = this.$container[insertMethod];
+            if (insertMethod) {
+                insertMethod.call(this.$container, navigatorTpl({
+                    region: putOnRight ? 'right' : 'left'
+                }));
+            } else {
+                throw "Unable to inject the component structure into the DOM";
```
krampstudio: Please throw using `new Error('..')`

views/js/runner/testReview.js
```diff
@@ -0,0 +1,610 @@
...
+            // access to the panel displayed when a linear part is reached
+            this.$linearState = this.$component.find('#qti-navigator-linear');
+        },
+
+        /**
+         * Installs the event handlers on the underlying DOM elements
+         * @private
+         */
+        _initEvents: function() {
+            var that = this;
```
krampstudio: we use `self` by convetion

views/js/runner/testReview.js
```diff
@@ -0,0 +1,610 @@
...
+         * Initializes the component
+         * @param {String|jQuery|HTMLElement} element The element on which install the component
+         * @param {Object} [options] A list of extra options
+         * @param {String} [options.region] The region on which put the component: left or right
+         * @param {Boolean} [options.sectionOnly] Limit the review screen to the current test section only
+         * @param {Boolean} [options.preventsUnseen] Prevents the test taker to access unseen items
+         * @returns {testReview}
+         */
+        init: function init(element, options) {
+            this.options = _.isObject(options) && options || {};
```
krampstudio: relfeclt hoisting in variable delcaration

views/templates/test_runner.tpl
```diff
@@ -1,6 +1,14 @@
...
 use oat\tao\helpers\Template;
 use oat\tao\helpers\Layout;
+
+$contentPanelSetup = '';
+if (get_data('review_screen')) {
+    $contentPanelSetup .= ' data-review-screen="1"';
+}
+if (get_data('review_region')) {
+    $contentPanelSetup .= ' data-review-region="' . get_data('review_region') . '"';
+}
```
krampstudio: There shouldn't be any construction in the template expect var,  if/then/else and for loops,

krampstudio:
I've an issue with all the ids in the template. 
I suggest you to use only one id (your top container or none if the container is given to factory) and then uses only the classes to access the elements.

krampstudio:
Last but not least, the testReview should be tested.

views/scss/navigator.scss
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+
+    .qti-navigator-counter {
+        text-align: right;
+        float: right;
+        @include font-size(12);
+    }
+
+    .qti-navigator-message {
```
krampstudio: empty selector

krampstudio:
Non blocking comments :
- use a media query to resize the the navigation screen
- in low resolution, modal feedback appears behind the navigation screen (z-index issue)
- message when trying to exit can be confusing
- the current item in the navigation should be highlighted (here the gray seems to be not enough)

PR #101, opened 2015-07-10
added the option order in button config

ssipasseuth:


views/js/testRunner/actionBarHook.js
```diff
@@ -62,13 +63,18 @@ define([
     function initQtiTool($toolsContainer, id, toolconfig, assessmentTestContext){
 
         if(isValid(toolconfig)){
-            
+
+            var order = _.parseInt(toolconfig.order);
+            if(isNaN(order)){
```
krampstudio: Use `_.isNan` in some platforms the behavior can be different (see https://github.com/lodash/lodash/blob/2.4.1/doc/README.md#_isnanvalue)

views/js/testRunner/actionBarHook.js
```diff
@@ -88,13 +94,56 @@ define([
...
+     * The orders are provided by data-order attribute set to the $button
+     * 
+     * @param {JQuery} $container
+     * @param {JQuery} $button
+     */
+    function _appendInOrder($container, $button){
+        
+        var order = $button.data('order');
+        if(order){
+            var $after, $before;
```
krampstudio: move them at the function level to reflect hoisting

views/js/testRunner/actionBarHook.js
```diff
@@ -88,13 +94,56 @@ define([
...
+     * Append a dom element $button to a $container in a specific order
+     * The orders are provided by data-order attribute set to the $button
+     * 
+     * @param {JQuery} $container
+     * @param {JQuery} $button
+     */
+    function _appendInOrder($container, $button){
+        
+        var order = $button.data('order');
+        if(order){
```
krampstudio: if order is 0, then you'll be in the else block and the button will be appended. Is it what you'd like to achieve ? Otherwise check if order is a number

ssipasseuth:
I have implemented the suggested changes

views/js/testRunner/actionBarHook.js
ssipasseuth: Yes, that means that the order is not defined, so it is appended in the end.

PR #102, opened 2015-07-13
Tao 875 section buttons

ssipasseuth:
The modification here allow adding button on a item or section basis, as requested here https://oat-sa.atlassian.net/browse/TAO-874 and here https://oat-sa.atlassian.net/browse/TAO-875

To test this, please use the sample qti test package available at https://oat-sa.atlassian.net/browse/TAO-875


ssipasseuth:
THis requires https://github.com/oat-sa/extension-tao-act/pull/10

views/js/controller/runtime/testRunner.js
```diff
@@ -288,6 +288,14 @@ define([
                     $controls.$skip.hide();
                     $controls.$skipEnd.hide();
                 }
+                
+                var config = module.config();
```
krampstudio: Could you move it to the top of the function

views/js/controller/runtime/testRunner.js
```diff
@@ -288,6 +288,14 @@ define([
...
+				var $toolsContainer = $('.tools-box-list');
```
krampstudio: You may only need to select the element in the `if` condition.

ssipasseuth:
suggested changes implemented

PR #108, opened 2015-07-15
Test review scope

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-1054 

The test taker review screen can be scoped for:
- the whole test
- the current test part
- the current test section

This scope is set up by the `test-taker-review-scope` config entry, which can take one of these values:
- `test`
- `testPart`
- `testSection`

This option entry is stored inside the `testRunner.conf.php` file

### DIFF
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -63,10 +63,13 @@ return array(
     'test-taker-review-region' => 'left',
 
     /**
-     * Limits the test taker review screen to the current test section.
-     * @type boolean
+     * Limits the test taker review screen to a particular scope. Can be:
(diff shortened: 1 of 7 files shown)


jsconan:
The last commit (`a2aebee`) fixes a bug that has also been fixed by another PR (https://github.com/oat-sa/extension-tao-testqti/pull/113).

The reason why I fixed it on my side is because I also changed the name of the variable, which was inconsistent with other names. In the test assessment context object, all the variables referring to items marked review are named `flagged`, with some prefix. The changed variable was named `numberReview` before, and now it is `numberFlagged`. And obviously I discovered the related bug, so I fixed it immediately.

PR #109, opened 2015-07-15
Fix/tao 999 multiple test type import

bursonic:
Depends on: oat-sa/extension-tao-itemqti#363


bugalot:
To avoid problematic conflicts (no re-base), I created an alternate PR doing basically the same thing.
https://github.com/oat-sa/extension-tao-testqti/pull/122

PR #119, opened 2015-07-21
Jump to next available item after a timeout is reached

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-864

When a item, a test section or a test part is timed out, the next button jump to the next runnable item.


jbout:
Will need to work with previous as well and should take into account https://github.com/oat-sa/qti-sdk/blob/legacy/qtism/runtime/tests/AssessmentTestSession.php#L1410

PR #132, opened 2015-07-29
PATCH. DO NOT MERGE. Story/tao 1051 close section on next

bursonic:
This PR is for Review only. The story should be delivered as a patch.


bursonic:
Reviewed and fixed at oat-sa/extension-tao-testqti/138

PR #156, opened 2015-08-13
TAO 1146:  fixed current item status in section exit message

bursonic:


ssipasseuth:
Sorry but this is not related to https://oat-sa.atlassian.net/browse/TAO-1146, is it ?

PR #165, opened 2015-08-24
Feature/actionbar refactoring

jsconan:
Updated behavior for the QTI Tools:
- action bar buttons can now be used as menus
- buttons are now normalized and should extend the base object actionBar/button
- base button behavior is provided by the base object (initializing, rendering and events processing)
- better positioning rules: the position can be either a number, or a string containing `first` or `last`
- can now use translations for the labels
- can display a group of buttons

Required by:
- https://github.com/oat-sa/extension-tao-mp/pull/4
- https://github.com/oat-sa/extension-tao-act/pull/63


jsconan:
This PR can now be merged

views/js/testRunner/actionBar/button.js
```diff
@@ -0,0 +1,558 @@
...
+        itemLoaded : function() {
+            // just a template method to be overloaded
+        }
+    };
+
+    /**
+     * Builds a button instance
+     * @returns {button}
+     */
+    var buttonFactory = function buttonFactory() {
```
krampstudio: Could you modify the factory in order to have an API that is more _straight forward_. Instead of using it like : 

```javascript
_.assign(buttonFactory(), {
       isVisible : function(){ return true; }
});
```

It would be better to do 

```javascript
buttonFactory({
       isVisible : function(){ return true; }
});
```

`_.defaults` or `_.merge` should be sufficient

views/js/test/actionBar/button/test.html
```diff
@@ -0,0 +1,43 @@
...
+        <base href="../../../../../../../../tao/views/" />
+        <link rel="stylesheet" type="text/css" href="js/lib/qunit/qunit.css">
+        <script type="text/javascript" src="js/lib/require.js"></script>
+        <script type="text/javascript" src="js/lib/qunit/qunit.js"></script>
+        <script type="text/javascript" src="js/lib/qunit/qunit-parameterize.js"></script>
+        <script type="text/javascript" src="js/lib/blanket/blanket.min.js" data-cover-only="actionBar/button.js"></script>
+
+        <script  type="text/javascript">
+            var taobase_www = '../../../../../../../../../tao/views/';
+            var base_lang = 'en';
```
krampstudio: Why are global variables needed ?

krampstudio: I'll remove them on a separate PR because it affects all the unit tests of the extension

PR #184, opened 2015-09-02
add script runner

hutnikau:
To run script execute the following command:
`php testCommand.php --action=endExpiredTests`

Tested on `KVDeliveryExecution` and `OntologyDeliveryExecution`

related on https://github.com/oat-sa/extension-tao-delivery/pull/101


jbout:
code needs to be moved into model and needs to respect dependencies which will require some major refactoring

PR #191, opened 2015-09-11
Added initialisation for variable subReport that used without declaring

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1340


antoinerobin:
thanks, saw that but forgot to fix it

PR #193, opened 2015-09-11
Fix: Forward Slashes Not Being Removed Causing Test Import to Fail

kendaop:
Fixed Windows issue with some forward slashes not being stripped by regex.
TESTED IN WINDOWS ONLY.


jbout:
Tested using Linux, Windows and Mac using NginX and Apache.

bugalot:
Thanks a lot for this contribution!!!

PR #195, opened 2015-09-14
Fixed style for top action bar with title

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1243

Requires https://github.com/oat-sa/tao-core/pull/539


Alroniks:
Hi Bertrand (@krampstudio). Dieter on holiday, so please see my PR or reassign to somebody else.

views/scss/test-runner.scss
```diff
@@ -50,14 +50,41 @@
...
             }
+            &.top-action-bar > .control-box {
+                position: relative;
+                & .control-box-group {
+                    position: absolute;
+                    top: 0;
+                    right: -5px;
+                    padding-right: 5px;
+
+                    background: -moz-linear-gradient(left,  transparent 0%, whiten($info, .1) 30px);
```
krampstudio: please use the `linear-gradient` mixin in `tao/views/scss/in/_functions.scss` (and vendor prefixes might not be necessary as the browsers we support doesn't need them)

Alroniks: `linear-gradient` count percentage automatically as I understood. It gets list of colors and creates gradient, but I need create gradient only for a piece of block - from 0 till 30px.

Alroniks: Also mixin `linear-gradient` use `background: linear-gradient( #{$css} );` for setting gradient, that not works in Chrome, because Chrome require -webkit prefix. I guess it should be solved by autoprefixer, but it isn't in dependencies.

Alroniks:
@krampstudio Fixed

Alroniks:
@krampstudio Tried solve issue using only CSS, but not found a solution. Added method to javascript, that check the width of the right block with controls and then calculates width for the title.

krampstudio:
I've made an experiment using flex boxes and a pure CSS solution seems to be working. 
I'll create a PR from this branch so you can work on top if it. 
Please ensure it works also with other browsers and with different configuration (I've only tested against FF and with the use case where timers are presents `.has-timer` CSS class).

Alroniks:
Thank you @krampstudio I've merged your suggestions, removed js-fallback and aded some fixed for support flexbox in Safari. Also I've checked in Chrome, Firefox and Safari on Mac, it's work.

Please see and check and give me feedback if it will be needed.

krampstudio:
Why do you have elements with a `table` display inside a `flex` element ?

Alroniks:
I use for fixing behavior of flex in Safari, but I will do some test, maybe it can be removed.

Alroniks:
Sorry @krampstudio Updated PR. Just left after experiments.

krampstudio:
I need to try it, but there's definitely something weird. If you define a container as display flex, you should define children with the flex attribute to define the base size, grow and shrink. If it works by using some default values, I would prefer to have those rules defined clearly.

Alroniks:
Defined flex attributes directly.

views/scss/test-runner.scss
```diff
@@ -50,14 +50,36 @@
...
+                }
             }
             .progressbar {
                 margin-top: 5px;
                 min-width: 150px;
+                max-width: 200px;
                 height: 0.6em;
             }
+
+            &.top-action-bar.has-timers > .control-box {
```
krampstudio: You might also need to remove the `.has-timers` because the progress bar isn't shown without a timer.
I suggest you to define also `flex-flow: row wrap` to ensure everything stays in the same line.

Alroniks:
@krampstudio Tested and updated PR

PR #196, opened 2015-09-15
Restore GUI before displaying the time limit message

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-1418

Fix the issue occurring when the test taker click outside the time limit message.

Need a time limited test. When the time limit is reached, a message is displayed. If you click on the Ok button, you are redirected to the next available item. But if you click outside, the modal will disappear leaving the GUI unusable unless you refresh the page. This patch fix this issue.

After merge into dev-act, you will also need to merge this branch into develop...


krampstudio:
Cherry picked to develop

PR #223, opened 2015-10-12
Fix the next section option

jsconan:
Use an item categorie as option to enable the next section button instead of the skip option.

To enable the next section button, you need to add one of these categories to each related items:
- `x-tao-option-nextSection`: enable the next section button
- `x-tao-option-nextSectionWarning`: enable the next section button and display a warning message when this button is clicked

### DIFF
--- a/views/js/controller/runtime/testRunner.js
+++ b/views/js/controller/runtime/testRunner.js
@@ -52,6 +52,7 @@ define([
         testMetaData,
         $doc = $(document),
         optionNextSection = 'x-tao-option-nextSection',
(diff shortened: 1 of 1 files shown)


views/js/controller/runtime/testRunner.js
```diff
@@ -51,6 +51,8 @@ define([
         timerIndex,
         testMetaData,
         $doc = $(document),
+        optionNextSection = 'x-tao-option-nextSection',
+        optionReviewScreen = 'x-tao-option-reviewScreen',
```
krampstudio: I have an issue keeping those codes directly in the test runner

PR #226, opened 2015-10-14
add timezone value

hutnikau:


views/js/controller/runtime/testRunner.js
```diff
@@ -372,7 +372,10 @@ define([
              */
             killItemSession : function (callback) {
                 testMetaData.addData({
-                    'ITEM' : {'ITEM_END_TIME_CLIENT' : Date.now() / 1000}
+                    'ITEM' : {
+                        'ITEM_END_TIME_CLIENT' : Date.now() / 1000,
+                        'ITEM_TIMEZONE' : moment().utcOffset(moment().utcOffset()).format('Z') + ' ' + /\((.*)\)/.exec(new Date().toString())[1]
+                    }
```
krampstudio: Do we really need the timezone abbreviation ? The form  `+02:00` or `+02` is ISO-8601

hutnikau: Thomas asked to include timezone name to this value. We are using `hh:mm` format in the ODS reports.

krampstudio: After checking, you cannot use new Date.toString to extract the TZ abbreviation. It's not consistently implemented accross browsers, for example IE11 write it as `(Pacific Dailight Time)` and Opera just ignore it and have only GMT+0200. 

The ODS isn't the subject of the TestRunner itself, and with UTC offset you can retrieve the abrreviation server side more consistently

hutnikau: Timezone name removed.

PR #227, opened 2015-10-15
TAO-1661 Support of assessmentSection->visible attribute in...

bugalot:
Test Runner!


bugalot:
@antoinerobin could you please recompile JS? (EXTRA THX)

antoinerobin:
no need to recompile JS, the testrunner is not compiled

PR #230, opened 2015-10-22
clear local storage

hutnikau:


views/js/testRunner/testMetaData.js
```diff
@@ -162,7 +162,27 @@ define([
...
+                            removed++;
+                        }
+                    }
+                    if (removed) {
+                        setLocalStorageData(val);
+                    } else {
+                        throw domException;
+                    }
+                }
+            }
```
krampstudio: you should aslo re-throw it if it is another type of exception (for example if the value isn't valid)

hutnikau: Done

krampstudio:
I'll reassign it to @jsconan for the testing. I'm agree with the implementation (at some point we will need a util module for storage that handle those edge cases, but for now it's sufficient).

jsconan:
Test OK for the immediate behavior: the localStorage is well cleaned at each transition.
And there is no more javascript error on tear down.

But, unfortunately the clean up does not work well when the storage is out of space.
Not all the keys are removed due to a misconception of your erase loop. By example you can read this article for some explanation about iterate and remove: http://billgonemad.com/blog/?p=34

Take also into account that erasing will take a while when the storage is full, and the browser will hang during this process.

Here is a more consistent erase loop:

```javascript
function cleanStorage(currentKey) {
    var key,
        removed = 0,
        i = localStorage.length;
    while (i) {
        key = localStorage.key(i--);
        if (/^testMetaData_.*/.test(key) && key !== currentKey) {
            window.localStorage.removeItem(key);
            removed++;
        }
    }
    return removed;
}
```

hutnikau:
@jsconan i just updated the code. Please review it again.

jsconan:
Ok, all is working.

PR #232, opened 2015-10-26
add a validator for the assessment test identifier

krampstudio:
add a specific validator 
as usual, opened files got jshint issues fixed

### DIFF
--- a/views/js/controller/creator/helpers/qtiTest.js
+++ b/views/js/controller/creator/helpers/qtiTest.js
@@ -33,7 +33,7 @@ define(['lodash', 'i18n'], function(_, __){
          * @param {Object} obj - the model to extract id from
          * @returns {Array} the extracted identifiers
          */
-        extractIdentifiers : function(obj){
+        extractIdentifiers : function extractIdentifiers(obj){
             var self = this;
             var identifiers = [];
             if(_.has(obj, 'identifier')){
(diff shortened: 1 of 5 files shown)


views/js/controller/creator/helpers/qtiTest.js
```diff
@@ -82,13 +82,29 @@ define(['lodash', 'i18n'], function(_, __){
             };
         },
 
+        /**
+         * Gives you a validator that check QTI id format of the test (it is different from the others...)
+         * @returns {Object} the validator
+         */
+        testidFormatValidator : function testidFormatValidator(){
+            var qtiTestIdPattern = /^\S+$/;
```
ssipasseuth: So this is is a valid identifier `Test-1ùµd*$}` ? This however is valid in the form, but an InvalidArgumentException is thrown by (vendor/qtism/qtism/qtism/data/AssessmentTest.php) at line 192 if I try to save the test. So something is wrong ...

krampstudio: I've changed the rule in order to conform to the xsd validation. Then it's a bug that need to be fixed on qtism.

ssipasseuth: I see... I believe the real bug actually is in the xsd but since we cannot wait for it to be officially modified, we would have to let this exception be.

ssipasseuth:
This fixes partially the issue https://oat-sa.atlassian.net/browse/TAO-1579. 
To make it complete, the fix must also be done in qtism : https://github.com/oat-sa/qti-sdk/blob/master/src/qtism/data/AssessmentTest.php#L180
@bugalot, this is a mission for you, if you accept !

PR #233, opened 2015-11-03
Keeping items time after resuming

boajer:


krampstudio:
@jsconan could you have a look to the client side part please

views/js/controller/runtime/testRunner.js
```diff
@@ -1078,6 +1106,9 @@ define([
                 TestRunner.beforeTransition();
                 TestRunner.testContext = testContext;
 
+                TestRunner.keepItemTimed(TestRunner.getSessionStateService().getDuration());
+                TestRunner.getSessionStateService().restart();
```
jsconan: I have a concern about this code, as the `keepItemTimed` will call the update method after the server has answered.

If you look the code above, there is already an explicit call to the update method, when the `onServiceApiReady` event is triggered.

If you trace the `update` method you will see that it is called twice on each page refresh. This is a waste of network resources, as this method triggers an item loading each time.

Maybe it can be better to place the call to the `keepItemTimed ` method in the `onServiceApiReady` event, instead of the `update` method. So this last will be called only once.

views/js/testRunner/ResumingStrategy/resetAfterResume.js
```diff
@@ -0,0 +1,43 @@
...
+         */
+        var sessionState = {
+
+            getDuration: function getDuration() {
+                return 0;
+            },
+
+            restart: function restart() {
+            }
+
```
jsconan: This code does not erase the local storage, so there may be still a value inside after reset. Is it intentional ?

boajer:
@jsconan  Could you take a look, i've updated PR

jsconan:
Ok, seems good now. Thanks for the change!

PR #236, opened 2015-11-05
force to register the validator again

krampstudio:
reapply hotfix from https://github.com/oat-sa/extension-tao-testqti/pull/235

### DIFF
--- a/views/js/controller/creator/creator.js
+++ b/views/js/controller/creator/creator.js
@@ -101,6 +101,7 @@ define([
             var $container = $('#test-creator');
             var $saver = $('#saver');
 
+            self.identifiers = [];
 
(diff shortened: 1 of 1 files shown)


ssipasseuth:
Side note : this PR has a small conflict due to https://github.com/oat-sa/extension-tao-testqti/pull/232, which fixes another issue reported by another client (https://oat-sa.atlassian.net/browse/TAO-1579 )
The related PR https://github.com/oat-sa/tao-core/pull/618 does not have any conflict because it does not benefit from the fix of made in  https://github.com/oat-sa/extension-tao-testqti/pull/232.
Let's keep in mind that the other issue (https://oat-sa.atlassian.net/browse/TAO-1579 ) may still happen in mpdeploy.

PR #244, opened 2015-11-19
Tao 1773 catch terminate state

jsconan:
Related to:
- https://oat-sa.atlassian.net/browse/TAO-1775
- https://oat-sa.atlassian.net/browse/TAO-1773

Requires: https://github.com/oat-sa/extension-tao-delivery/pull/133

Catch error messages and display them.

### DIFF
--- a/views/js/ResultServerApi.js
+++ b/views/js/ResultServerApi.js
@@ -37,7 +37,7 @@ define(['jquery', 'i18n', 'iframeNotifier'], function($, __, iframeNotifier) {
             //there is no error management, so doing an alert (an eval and I'll burn in hell...)
             //TODO manage errors during the delivery
             if (response && response.message) {
-                if (response.code === TEST_STATE_CLOSED && response.code === TEST_STATE_SUSPENDED) {
+                if (response.code === TEST_STATE_CLOSED || response.code === TEST_STATE_SUSPENDED) {
                     message = false;
                 } else {
                     message = response.message;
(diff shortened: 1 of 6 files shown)


views/js/ResultServerApi.js
```diff
@@ -22,11 +26,26 @@ define(['jquery', 'iframeNotifier'], function($, iframeNotifier) {
...
-
+            if (response && response.message) {
+                if (response.code === TEST_STATE_CLOSED || response.code === TEST_STATE_SUSPENDED) {
+                    message = false;
+                } else {
+                    message = response.message;
+                }
+            }
+            if (message) {
+                alert(message);
```
ssipasseuth: what about using your brand new dialogAlert here now ? :)

views/js/controller/runtime/testRunner.js
```diff
@@ -956,6 +994,35 @@ define([
...
+             */
+            processError : function processError(error) {
+                var self = this;
+
+                // keep disabled
+                this.hideGui();
+                this.beforeTransition();
+
+                // ask the parent to display a message
+                iframeNotifier.parent('alertMessage', {
```
ssipasseuth: Real question : I am not about the event naming convention, should we not avoid upper case ? should we have alert-message or message-alert instead ?

views/js/ResultServerApi.js
jsconan: Sure, it would be fine. Unfortunately I did not succeed to use this component in this context. The browser doesn't display it. I think it is because the stylesheets are not loaded in the frame. However I didn't spend too much time to find a solution, but it can be a good improvement.

PR #249, opened 2015-11-25
bug fix for taoQtiTest import

zShurick:
Windows 10 Pro x64
```
XmlDocument->getDomDocument()->documentUri
``` 
in path returns "file:/" instead of the expected "file:///" 

Test it doesn't break existing taoQtiTest import (for example QtiPackage.zip)
Test if the related unit tests are still running

### DIFF
--- a/helpers/class.Utils.php
+++ b/helpers/class.Utils.php
@@ -180,7 +180,7 @@ class taoQtiTest_helpers_Utils {
         $basePath = helpers_File::truePath($basePath);
         $basePath .= DIRECTORY_SEPARATOR;
         
-        $documentURI = preg_replace('/^file:\/{3}/', '', $test->getDomDocument()->documentURI);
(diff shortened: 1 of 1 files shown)


helpers/class.Utils.php
```diff
@@ -180,7 +180,7 @@ static public function buildAssessmentItemRefsTestMap(XmlDocument $test, taoQtiT
         $basePath = helpers_File::truePath($basePath);
         $basePath .= DIRECTORY_SEPARATOR;
         
-        $documentURI = preg_replace('/^file:\/{3}/', '', $test->getDomDocument()->documentURI);
+        $documentURI = preg_replace("|^file:/+|", '', $test->getDomDocument()->documentURI);
```
antoinerobin: Can you replace the + by {1,3} to not match more than 3 / ? 
Can you escape the / to keep the pattern clear ?
Can you keep the / to delimit pattern to keep consistency over the project ?

zShurick:
all done

PR #257, opened 2015-12-18
Added test change event on all navigation

jbout:


models/classes/event/QtiTestChangeEvent.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+    
+    public function getServiceCallId()
+    {
+        return $this->session->getSessionId();
+    }
+    
+    public function getNewStateDescription()
+    {
+        $pos = $this->session->getRoute()->getPosition();
```
antoinerobin: maybe user wants to have item from 1 to ```count``` of 0 to ```count - 1``` ?

PR #261, opened 2015-12-28
Trigger event before saving session

hutnikau:


jbout:
Closed for now, as the timing of the event would be on save instead of on change

PR #263, opened 2016-01-07
Feature/custom session desc

jbout:


boajer:
Ops. What about registering new service ?

jbout:
It's not a new service, only the constant and the option are new. If the option is not set in config the default is used automatically

PR #269, opened 2016-01-14
Feature/implement runner service

antoinerobin:
implementation of getItemdata, get and set state


models/classes/runner/QtiRunnerService.php
```diff
@@ -323,21 +324,20 @@ public function getRubrics(RunnerServiceContext $context)
...
-        $response = [];
-
-        // TODO: Implement getItemData() method.
-
         if ($context instanceof QtiRunnerServiceContext) {
+            $directoryIds = explode('|', $itemRef);
+            $dirPath = \tao_models_classes_service_FileStorage::singleton()->getDirectoryById($directoryIds[2])->getPath();
+            $itemFilePath = $dirPath . QtiJsonItemCompiler::ITEM_FILE_NAME;
+
+            return $itemFilePath;
```
jsconan: I think there is something missing here. This method should return the item data, not the path of the file containing the data.

I would rather do something like:

```php
return file_get_contents($itemFilePath);
```

PR #273, opened 2016-01-15
TestRunner: add alternative item data loading

jsconan:


antoinerobin:
on a 50kb item it takes an average of 450ms instead of 530ms.
To use this alternative data loading we have to remove the json_decode() from QtiRunnerService/getItemData.

krampstudio:
So we can consider this as an optimization. I'll close this PR, but could you redo it on develop ?

antoinerobin:
Yes i'll do it

PR #277, opened 2016-01-15
Tao 1905 feature/newrunner

krampstudio:
This is a huge PR, could you have a quick review. 
Tests are missing and will be implemented after code is consolidated. Some for jsdoc.


views/js/runner/provider/qti.js
```diff
@@ -0,0 +1,231 @@
...
+                //to be sure load start after unload...
+                //we add an intermediate ns event on unload
+                self.on('unloaditem.' + action, function(){
+                    self.off('.'+action);
+
+                    self.getProxy()
+                        .callItemAction(context.itemUri, action, params)
+                        .then(function(results){
+
+                                self.setTestData(results.testData);
```
jsconan: Maybe I am wrong, but the test data is only got from the init action or the getTestData action, so it will be null after all other actions

PR #290, opened 2016-01-21
TestRunner server side get feedback on store variable

krampstudio:
requires https://github.com/oat-sa/extension-tao-itemqti/pull/533


models/classes/runner/QtiRunnerService.php
```diff
@@ -461,6 +469,107 @@ public function storeItemResponse(RunnerServiceContext $context, $itemRef, $resp
...
+            $session = $context->getTestSession();
+
+            $itemDirectory = $this->getItemDataFolder($itemRef);
+            $varEltPath    = $itemDirectory . QtiJsonItemCompiler::VAR_ELT_FILE_NAME;
+
+
+            if (file_exists($varEltPath)) {
+                $variableData = json_decode(file_get_contents($varEltPath), true); 
+                foreach( $variableData as $key => $element){
+                    if($element['qtiClass'] == 'modalFeedback'){
```
antoinerobin: $element doesn't always have qtiClass key. You should add ```isset($element['qtiClass'])```

PR #291, opened 2016-01-21
dialog plugin implementation

krampstudio:
naive implementation of the dialog plugin, tests and improved behavior to come


views/js/controller/runner/runner.js
```diff
@@ -67,15 +68,33 @@ define([
...
+         *
+         * TODO verify required options
+         *
+         * @param {Object} options - the testRunner options
+         * @param {String} options.testDefinition
+         * @param {String} optins.testCompilation
+         * @param {String} optins.serviceCallId
+         * @param {String} optins.serviceController
+         * @param {String} optins.serviceExtension
+         * @param {String} optins.exitUrl - the full URL where to return at the final end of the test
```
jsconan: Misspelling name of parameter in the doc: `optins` should be replaced by `options`

views/js/runner/plugins/content/dialog/dialog.js
```diff
@@ -0,0 +1,52 @@
...
+         * Initialize the plugin (called during runner's init)
+         */
+        init : function init(){
+            var self = this;
+            var testRunner = this.getTestRunner();
+
+            //change plugin state
+            testRunner
+                .on('alert',  dialogAlert)
+                .on('confirm', dialogConfirm);
```
jsconan: I have a concern about this: how can we be aware of the user's response ? Is it by providing the callback through the event triggering ? In that case, the 'alert' and 'confirm' events must always be listened by one plugin, otherwise the callback will never be called and the runner can be locked.

krampstudio:
Feedback becomes also a required plugin.

views/js/runner/provider/qti.js
```diff
@@ -282,9 +283,13 @@ define([
                     assetManager: assetManager
                 })
                 .on('error', reject)
-                .on('render', resolve)
-                .on('responsechange', changeState)
-                .on('statechange', changeState)
+                .on('render', function(){
+
+                    this.on('responsechange', changeState);
```
krampstudio: I need to fix that...

PR #299, opened 2016-01-26
Fix/timeout alert

jsconan:
Fix double alert/warning
Prevent the test to start if timed out

### DIFF
--- a/views/js/runner/provider/qti.js
+++ b/views/js/runner/provider/qti.js
@@ -256,20 +256,21 @@ define([
             .on('timeout', function(){
 
                 var context = self.getTestContext();
-                var computeNextTimeout = _.partial(computeNext, 'timeout');
 
                 context.isTimeout = true;
 
                 self.disableItem(context.itemUri);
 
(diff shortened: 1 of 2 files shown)


views/js/runner/provider/qti.js
```diff
@@ -139,7 +139,11 @@ define([
                 var context = self.getTestContext();
                 var states = self.getTestData().states;
                 if(context.state <= states.interacting){
-                    self.loadItem(context.itemUri);
+                    if (context.isTimeout) {
+                        self.timeout();
```
krampstudio: This leads on infinite loop like workflow : 

`self.timeout() -> alert -> store -> computeNext -> load -> self.timeout() -> etc.`

jsconan: Actually not. The timeout action requests the server, which will send a new context.
This action is called only when the test is still alive (interacting or something approaching), and only when the timeout flag is set. 
When a timeout is triggered, the server will update the context and must not return a new timeout context, it will rather return the next available route or a closed state.

PR #303, opened 2016-01-27
Feature/plugin comment

jsconan:


views/js/runner/plugins/navigation/button.tpl
```diff
@@ -3,4 +3,5 @@
         <span class="icon-{{icon}}"></span>
         <span class="text">{{text}}</span>
     </a>
+    {{{content}}}
```
krampstudio: I think this is confusing as part of the buttonTpl. For this plugin I would even prefer to have the content of `commentTpl` injected with jQuery.

PR #310, opened 2016-01-29
turn off alert() error message with http response =  403

zShurick:
https://oat-sa.atlassian.net/browse/TAO-2101
depends on
oat-sa/tao-core#721
oat-sa/extension-tao-delivery#156


zShurick:
@bugalot Could you check it, until the evening, please

PR #331, opened 2016-02-09
Fix/save meta data server

hutnikau:


models/classes/TestSessionMetaData.php
```diff
@@ -231,32 +230,29 @@ public function getData()
     {
         $request = Context::getInstance()->getRequest();
         $data = $request->hasParameter('metaData') ? $request->getParameter('metaData') : array();
-        $action = Context::getInstance()->getActionName();
         $route = $this->getTestSession()->getRoute();
 
-        if (in_array($action, array('index'))) {
+        if ($route->getPosition() === 0) { //very first item
```
jbout: Navigating back to the first item will add another TAO_VERSION variable

PR #336, opened 2016-02-11
TestRunner: add config entry for plugins related config

jsconan:
Check existence of each config entry before accessing it
Add config entry for plugins related config


antoinerobin:
you should add the config also in the update script

PR #337, opened 2016-02-11
send attempt every time we do next/previous/jump

antoinerobin:


jbout:
@krampstudio please take a look
Still needs to be verified with every interaction if they handle empty responses nicely

PR #341, opened 2016-02-12
Tao 2163 timer countdown latency

hutnikau:


jbout:
required by https://github.com/oat-sa/extension-tao-act/pull/151

PR #344, opened 2016-02-12
TestRunner: filter response/state at proxy level

jsconan:


krampstudio:
This has to be handled by the IR itself. It should always send well formatted responses.

krampstudio:
Fixed by https://github.com/oat-sa/extension-tao-testqti/pull/347

PR #356, opened 2016-02-26
Tao 2254 display items count

jsconan:
Related to:
- https://oat-sa.atlassian.net/browse/TAO-2254
- https://oat-sa.atlassian.net/browse/TAO-2255

The displayed messages may be incorrect without this fix: https://github.com/oat-sa/extension-tao-testqti/pull/353

### DIFF
--- a/views/js/runner/helpers/messages.js
+++ b/views/js/runner/helpers/messages.js
@@ -38,7 +38,7 @@ define([
             var response = state && state.response;
             if (_.isObject(response)) {
                 // base or record defined: the interaction has a response, so the item is responded
-                if (_.isObject(response.base) || _.isObject(response.record)) {
+                if (_.isObject(response.base) || _.isObject(response.record) || _.isArray(response.record)) {
                     answered = true;
                 }
                 else if (_.isObject(response.list)) {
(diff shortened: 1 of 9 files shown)


views/js/runner/helpers/messages.js
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Tells is the current item has been answered or not
+     * The item is considered answered when at least one response has been set to not empty {base : null}
+     *
+     * @returns {Boolean}
+     */
+    function isCurrentItemAnswered(runner) {
+        var answered = false;
+        _.each(runner.itemRunner && runner.itemRunner.getState(), function (state) {
```
krampstudio: please use `_.forEach` (as lodash changes the aliases quite often)

views/js/runner/helpers/messages.js
```diff
@@ -0,0 +1,86 @@
...
+            var response = state && state.response;
+            if (_.isObject(response)) {
+                answered = _.isObject(response.base);
```
krampstudio: This is what we talk it could be also `response.list` or `response.record` according to http://www.imsglobal.org/assessment/pciv1p0cf/imsPCIv1p0cf.html#_Toc353965343

PR #360, opened 2016-03-01
Feature/refactor map stats

jsconan:
Use the new map helpers to compute stats in a factorized manner instead of duplicating the code in each needed place


views/js/test/runner/helpers/map/test.js
```diff
@@ -182,4 +185,123 @@ define([
     });
 
 
+    QUnit.test('helpers/map.updateItemStats', 25, function(assert) {
```
krampstudio: Just FYI but don't require changes here (only for the future). The next version of QUnit doesn't support the expect as 2nd argument anymore. So please use `QUnit.expect` instead (it will be easier to refactor then).

PR #368, opened 2016-03-08
TestRunner: add security token to counter CSRF, gather item queries i…

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2197

Requires: https://github.com/oat-sa/extension-tao-test/pull/78

May require: https://github.com/oat-sa/extension-tao-testqti/pull/367

### DIFF
--- a/actions/class.Runner.php
+++ b/actions/class.Runner.php
@@ -25,6 +25,8 @@ use oat\taoQtiTest\models\runner\QtiRunnerServiceContext;
 use oat\taoQtiTest\models\runner\QtiRunnerClosedException;
 use oat\taoQtiTest\models\runner\QtiRunnerPausedException;
 use oat\taoQtiTest\models\event\TraceVariableStored;
+use \oat\taoTests\models\runner\CsrfToken;
+use \oat\taoTests\models\runner\SessionCsrfToken;
 
 /**
  * Class taoQtiTest_actions_Runner
@@ -46,6 +48,12 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
     protected $serviceContext;
 
     /**
+     * The anti-CSRF manager
+     * @var CsrfToken;
+     */
+    protected $csrf;
+
+    /**
      * taoQtiTest_actions_Runner constructor.
      */
     public function __construct()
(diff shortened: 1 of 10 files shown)


actions/class.Runner.php
```diff
@@ -67,6 +103,11 @@ protected function getServiceContext($check = true)
         if (!$this->serviceContext) {
             $testDefinition = $this->getRequestParameter('testDefinition');
             $testCompilation = $this->getRequestParameter('testCompilation');
+            $csrfToken = $this->getRequestParameter('x-security-token');
+
+            if (!$this->getCsrf()->checkCsrfToken($csrfToken)) {
+                throw new common_exception_InconsistentData('CSRF attempt!');
```
krampstudio: What's the effect of this exception ? Is the exception caught ?
The expected behavior would be to
1. log the XSRF attempt (with an error level)
2. response with a 403 status code but without any explicit reason.

views/js/runner/proxy/qtiServiceProxy.js
```diff
@@ -31,23 +31,36 @@ define([
...
      * @returns {Promise}
      */
-    function request(url, params, contentType) {
+    function request(proxy, url, params, contentType) {
+        var headers = {};
+        var securityToken = proxy.getSecurityToken();
+        var token = securityToken.getToken();
+
+        if (token) {
+            headers['x-security-token'] = token;
```
krampstudio: Usually the header used to cover that situation is `X-Auth-Token` (

krampstudio:
joining the both requests is really a good improvement in terms of performances but also the code is more clean and clear. :+1:

actions/class.Runner.php
```diff
@@ -57,6 +65,34 @@ public function __construct()
...
+    protected function returnJson($data, $httpStatus = 200)
+    {
+        // auto append the CSRF token to the result
+        if (is_array($data)) {
+            $data['token'] = $this->getCsrf()->getCsrfToken();
+        } else if (is_object($data)) {
+            $data->token = $this->getCsrf()->getCsrfToken();
+        }
+
+        parent::returnJson($data, $httpStatus);
```
krampstudio: Could you just add a `return` before `parent::returnJson($data, $httpStatus);` as a good practice because the parent method send headers and write out content so we don't want something to append after.

jsconan:
All changes made

PR #374, opened 2016-03-14
check session state

hutnikau:


actions/class.Runner.php
```diff
@@ -341,7 +341,7 @@ public function submitItemState()
         $serviceCallId = $this->getRequestParameter('testServiceCallId');
 
         try {
-            $serviceContext = $this->getServiceContext(false);
+            $serviceContext = $this->getServiceContext(true);
```
jsconan: You cannot do that, this will prevent the item state to be stored when the test session is suspended. The parameter is to allow the session check.

By example:
- the proctor pauses a test
- the test taker click on the next button
- the current item state must be stored, so when the test is resumed the test taker retrieve its choices.

With this change, the session is checked and as it is suspended, the state is not stored.

hutnikau: Why state should be stored when session is paused?

As far as i understand this is expectable behavior. Data must not be stored anymore if session is paused/terminated.

Without this fix test taker gets an error after submitting item (and GUI no longer works)  if proctor terminates session.

jsconan: It is related to this issue:

https://oat-sa.atlassian.net/browse/TAO-2333

The response must not be stored after a pause, but the state needs to be stored.

After a terminate, I agree this is useless to store the state.

jsconan: Maybe this code can fix your issue:

```php
            $serviceContext = $this->getServiceContext(false);
            $sessionState = $serviceContext->getTestSession()->getState();
            if ($sessionState == AssessmentTestSessionState::CLOSED) {
                throw new QtiRunnerClosedException();
            }
```

hutnikau:
@jsconan branch has been updated. Please review it again.

P.S. Thank you for advice.

PR #375, opened 2016-03-16
Fix/tao 2290 delivery compilation errors report

siwane:


models/classes/class.QtiTestCompiler.php
```diff
@@ -338,9 +338,23 @@ public function compile() {
...
             $report->add($subReport);
 
+            $previous = $e->getPrevious();
+            while (true) {
+
+                if (!$previous instanceof Exception) {
+                    break;
+                }
+
+                $subReport = new common_report_Report(common_report_Report::TYPE_ERROR, __($previous->getMessage()));
```
bugalot: This will be hard to translate as we do not know the nature of the various possible messages.

PR #377, opened 2016-03-16
UI fixs for the the new Test Runner

krampstudio:
- prevent horizontal scrolling on the review panel
 - increase the overlay opacity to


boajer:
@krampstudio  What was the case with horizontal scrolling ?  I can't reproduce original issue

krampstudio:
Try with spanish enabled

krampstudio:
For timing reason, and because of the simplicity of the PR, I'm reviewing myself...

PR #380, opened 2016-03-17
Fix/tao 2381 timer delay

krampstudio:
- accumulate decimals on timer delays
 - sync also the constraints so the section isn't timeout too early


krampstudio:
Sorry @boajer to flood you of PR...

PR #382, opened 2016-03-18
check if item session given

hutnikau:


krampstudio:
could you please elaborate

hutnikau:
When delivery execution paused/terminated from proctoring interface user gets error after click on  `next` button instead of appropriate message because result object has no `itemSession` property.

krampstudio:
In that case, in addition the server's `move` should return 
```js
{ 
    success: false,
   code: 0,
   type: 'TestState',
   message: __('Session paused')
 }
```

hutnikau:
But move request even never be called without this fix.

PR #384, opened 2016-03-18
Removed redundant count correction

boajer:
Original issue:
Incorrect count of unanswered items in exit confirmation message if test taker jumped back to item with selected answer.
Reproduction video: https://goo.gl/xJT2d5


krampstudio:
Could you please update the unit test to reflect your changes.

PR #398, opened 2016-03-25
Feature/csrf token option

krampstudio:


actions/class.Runner.php
```diff
@@ -114,10 +114,14 @@ protected function getServiceContext($check = true, $checkToken = true)
             $testCompilation = $this->getRequestParameter('testCompilation');
 
             if ($checkToken) {
-                $csrfToken = $this->getRequestParameter('X-Auth-Token');
-                if (!$this->getCsrf()->checkCsrfToken($csrfToken)) {
-                    \common_Logger::w("CSRF attempt! The token $csrfToken is no longer valid!");
-                    throw new \common_exception_Unauthorized();
+
+                $config = common_ext_ExtensionsManager::singleton()->getExtensionById('taoQtiTest')->getConfig('testRunner');
```
jsconan: I think there is a better way to get the option:

`$config = $this->runnerService->getTestConfig();`

PR #404, opened 2016-04-01
Add interfaces to describe the timing API

jsconan:


jsconan:
Moving this to the taoTests extension

PR #408, opened 2016-04-05
Feature/tao 2486 timing tests

hutnikau:


krampstudio:
That's sound great, could you also test the compute with more points, like the duration with at least 3 points pairs

hutnikau:
@krampstudio I added 4 more time points to `testCompute()` method.

PR #409, opened 2016-04-05
Refactor/test definition not recursive

siwane:


jbout:
Closed until update script and relative file paths are taken into account

PR #412, opened 2016-04-06
Tao 2486 qti timer test

hutnikau:
Tests coverage - 100%


hutnikau:
I've merged develop branch, because of that there changes in manifest and Updater.php files.

PR #418, opened 2016-04-08
Fix/tao 2487 occurrences

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2487

Add a tag with item ref and num attempts in each TimePoint of the time.

### DIFF
--- a/models/classes/runner/session/TestSession.php
+++ b/models/classes/runner/session/TestSession.php
@@ -66,7 +66,7 @@ class TestSession extends taoQtiTest_helpers_TestSession
     public function getTimer()
     {
         if (!$this->timer) {
-            $this->timer = new QtiTimer();
+            $this->timer = new QtiTimer($this);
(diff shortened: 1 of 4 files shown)


models/classes/runner/session/TestSession.php
```diff
@@ -66,7 +66,7 @@ class TestSession extends taoQtiTest_helpers_TestSession
     public function getTimer()
     {
         if (!$this->timer) {
-            $this->timer = new QtiTimer();
+            $this->timer = new QtiTimer($this);
```
krampstudio: cyclic dependency...

krampstudio:
By linking the session to the Timer you rely on session's state, just to get the attempt. I'm wondering how we could avoid that. Using an itemRef that contains it could be an idea, then it wouldn't be a RouteItem anymore but an object that contains only the data we need.
Or does it make sense to have the attempt number as part of the interface ?

PR #419, opened 2016-04-08
retrieve the duration using an event

krampstudio:
- duration defaults to 0
 - use an event to communicate with the duration plugin instead of a shared store instance

### DIFF
--- a/views/js/runner/plugins/controls/duration/duration.js
+++ b/views/js/runner/plugins/controls/duration/duration.js
@@ -68,7 +68,6 @@ define([
                     //how many time elapsed from the last tick ?
                     var elapsed = self.stopwatch.tick();
                     var context = testRunner.getTestContext();
-                    var storage = testRunner.durationStore;
 
                     //store by attempt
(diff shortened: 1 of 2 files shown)


krampstudio:
requires ttps://github.com/oat-sa/extension-tao-test/pull/102

views/js/runner/plugins/controls/duration/duration.js
```diff
@@ -70,15 +73,13 @@ define([
                     //store by attempt
                     var itemAttemptId = context.itemIdentifier + '#' + context.attempt;
 
-                    if(storage){
-                        storage.getItem(itemAttemptId).then(function(duration){
-                            duration = _.isNumber(duration) ? duration : 0;
-                            elapsed  = _.isNumber(elapsed) && elapsed > 0 ? (elapsed / 1000) : 0;
+                    durationStore.getItem(itemAttemptId).then(function(duration){
```
jsconan: Not a big issue, but the old storage variable is left unused

PR #423, opened 2016-04-11
Fix double displayed message

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2531

Do not display error message if it was already processed


krampstudio:
It works as expected but I don't like the way the errors are handled. There's a slot for improvments

PR #425, opened 2016-04-12
Take care of existing client range when adjusting duration

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2484

If a client range already exist, replace it by the new duration.


krampstudio:
what's the best way/scenario to test it ? pause/resume ?

jsconan:
To test it:

TestTaker: start a test, go to a timed section
Proctor: pause

TestTaker: want to resume
Proctor: resume

TestTaker: stay on the same item without moving
Proctor: pause

TestTaker: want to resume
Proctor: resume

TestTaker: stay on the same item without moving
Proctor: pause

krampstudio:
Looks ok

PR #436, opened 2016-04-18
Fix/tao 2545 timer end

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2545

When submitting the item, the timer cannot be updated if the test session is closed, so update it also when the test is terminated.

The test session manager can now instanciate the right test session class, based on config entry


jsconan:
@antoinerobin please do the code review then assign to @krampstudio for tests

helpers/class.SessionManager.php
```diff
@@ -100,11 +105,19 @@ public function getTest() {
     
     /**
      * Instantiates an AssessmentTestSession with the default implementation provided by QTISM.
-     *
+     * @param AssessmentTest $test
+     * @param Route $route
      * @return AssessmentTestSession
      */
     protected function instantiateAssessmentTestSession(AssessmentTest $test, Route $route) {
-        return new taoQtiTest_helpers_TestSession($test, $this, $route, $this->getResultServer(), $this->getTest());
```
antoinerobin: I discovered an issue in ```oat\taoQtiTest\models\runner\QtiRunnerService```that will try to do an ```initItemTimer```that doesn't exists in the ```taoQtiTest_helpers_TestSession```

PR #441, opened 2016-04-22
Removed non required use statements

jbout:


jbout:
covered by https://github.com/oat-sa/extension-tao-testqti/pull/443

PR #444, opened 2016-04-25
Release 2.25.0

jsconan:
Features:
- Use the core plugin loader instead
- Feature/rest import
- Autoscroll the review panel to show the current selected item
- Better handling of test session for the new test runner implementation

Fix:
- Compiled javascripts with FileReader polyfill removed from the UploadInteraction
- Fix missing dependencies
- Fix timer ranges.
- Do not display error message if it was already processed
- Sets the correct state for the show/hide review button after the review panel is restored
- Display item label instead of item identifier in test runner messages
- Check test timeout on load
- Fix duration clearing at the end of the test
- Fix the wrong use of InvalidArgumentTypeException
- The new test runner can now load an run portable custom interactions


krampstudio:
You can create the tag `v2.25.0` from `master`

PR #447, opened 2016-04-28
Tao 2561 restrict item when disabled

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2561

Requires: https://github.com/oat-sa/extension-tao-test/pull/109


views/js/runner/provider/qti.js
```diff
@@ -89,7 +89,36 @@ define([
                 'serviceController',
                 'serviceExtension'
             ]);
-            return proxyFactory('qtiServiceProxy', proxyConfig);
+            return proxyFactory('qtiServiceProxy', proxyConfig)
+                // middle ware invoked on every requests
+                .use(function (req, res, next) {
```
krampstudio: could you give a name to your middlewre function, this will facilitate debugging (especially if we are going to add new ones)

views/js/runner/provider/qti.js
```diff
@@ -89,7 +89,36 @@ define([
...
+                    var context = self.getTestContext();
+                    var data = res && res.data;
+
+                    // test has been closed/suspended => redirect to the index page after message acknowledge
+                    if (data && data.type && data.type === 'TestState') {
+
+                        if(!self.getState('ready')){
+                            //if we open an inconsistent test (should never happen) we let a few sec to
+                            //read the error message then leave
+                            _.delay(function (){
```
krampstudio: is the delay still required ?

views/scss/inc/_test-layout.scss
```diff
@@ -91,6 +91,11 @@
             bottom: 0;
             width: 100%;
             opacity : .9;
+
+            &-full {
+                background-color: Window;
```
krampstudio: ??

views/js/runner/provider/qti.js
```diff
@@ -273,10 +299,7 @@ define([
 
                 submit()
                  .then(updateStats)
-                 .then(computeNextMove)
-                 .catch(function(err){
-                    self.trigger('error', err);
-                 });
+                 .then(computeNextMove);
```
krampstudio: I think we should keep the `catch` method. I know the promise inside the proxy is catched however, here if the promise inside `updateStats` is rejected, we won't go into the proxy's `cacth` (same for the other locations of the provider).

views/js/runner/provider/qti.js
jsconan: This part was taken from the previous implementation in the controller. According to the comment above, this delay was added to let the user see the error message.
But I agree it is not very useful as the browser will redirect the page anyway.
So I can remove it if you prefer.

views/scss/inc/_test-layout.scss
jsconan: By default, the background is transparent and the opacity is not taken into account unless a solid background color is set. This line just arbitrary force the background color with the window's one.

The main concern is about which color to choose, as the custom theme may overwrite it. Taking the color of the window was a mere solution.

views/js/runner/provider/qti.js
jsconan: Ok, but in this case there will be several `error` triggering, unless i keep the `catch()` on each points and remove the centralized error handling.

views/scss/inc/_test-layout.scss
krampstudio: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#System_Colors those values are deprecated. 
You can use `$uiGeneralContentBg`. Then it's the responsibility of the theme to override them.

views/js/runner/provider/qti.js
krampstudio: Did you try ? if the promise is returned, then only the first catch is called (it should be in the same promise chain), see http://jsbin.com/lusizur/edit?js,console

jsconan:
All changes made

PR #450, opened 2016-05-04
Refactor/change tmp location to state storage

bugalot:


actions/class.TestRunner.php
```diff
@@ -222,6 +224,21 @@ protected function getTestMeta() {
 	}
 
     /**
+     * Return the State storage
+     * @return StateStorageQtiFileManager
+     * @throws common_exception_Error
+     */
+    protected function getStateStorageQtiFileManager()
```
bugalot: We should instantiate the StateStorageQtiFileManager just once in this controller.

PR #454, opened 2016-05-11
Feature/tao 2583 communicator polling

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2583

Requires:
- https://github.com/oat-sa/tao-core/pull/823
- https://github.com/oat-sa/extension-tao-test/pull/111
- https://github.com/oat-sa/extension-tao-test/pull/113
- https://github.com/oat-sa/extension-tao-delivery/pull/172

Implementation of the polling communication for the new test runner

### DIFF
--- a/views/js/controller/runner/runner.js
+++ b/views/js/controller/runner/runner.js
@@ -69,13 +69,6 @@ define([
     function initRunner(config) {
         var plugins = pluginLoader.getPlugins();
 
-        /**
-         *  At the end, we are redirected to the exit URL
-         */
-        var leave = function leave (){
-            window.location = config.exitUrl;
-        };
-
         config = _.defaults(config, {
             renderTo: $('.runner')
         });
@@ -96,13 +89,10 @@ define([
                 //TODO move the loading bar into a plugin
                 loadingBar.stop();
             })
-            .after('finish', function () {
-                this.trigger('leave');
+            .on('destroy', function () {
+                // at the end, we are redirected to the exit URL
+                window.location = config.exitUrl;
             })
-            .on('leave', function () {
-                this.destroy();
-            })
-            .on('destroy', leave)
             .init();
     }
 
--- a/views/js/controller/runner/runner.js
+++ b/views/js/controller/runner/runner.js
@@ -26,6 +26,8 @@ define([
     'i18n',
     'module',
     'core/promise',
+    'core/communicator',
+    'core/communicator/poll',
     'layout/loading-bar',
     'ui/feedback',
     'taoTests/runner/runner',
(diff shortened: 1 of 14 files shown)


views/js/controller/runner/runner.js
```diff
@@ -44,7 +46,8 @@ define([
      */
 
     runner.registerProvider('qti', qtiProvider);
-    proxy.registerProxy('qtiServiceProxy', qtiServiceProxy);
+    proxy.registerProvider('qtiServiceProxy', qtiServiceProxy);
+    communicator.registerProvider('poll', pollProvider);
```
krampstudio: Shouldn't you register the provider in the proxy itself ? This is the responsibility of the proxy to select the way to communicate.

jsconan: In fact, the type of communication is set from a config entry. Here it is just the provider registration.

views/js/runner/provider/qti.js
```diff
@@ -569,14 +586,22 @@ define([
...
-            this.itemRunner = null;
         }
     };
 
+    /**
+     * Displays an exit message, then leaves the runner once the user has acknowledged
+     * @param {runner} runner
+     * @param {Object} data
+     */
+    function leaveRunner(runner, data) {
```
krampstudio: Could you move it next to the `load` and `submit` function to keep consistency ?

views/js/runner/provider/qti.js
```diff
@@ -400,12 +394,29 @@ define([
...
             return this.getProxy().init()
                        .then(function(results){
+                            var communicator = self.getProxy().getCommunicator();
+
                             self.setTestData(results.testData);
                             self.setTestContext(results.testContext);
                             self.setTestMap(results.testMap);
+
+                            // install the communication channel if enabled
+                            if (communicator) {
```
krampstudio: we need to discuss that part. I'm not sure if this should go into the proxy or not...

views/js/runner/provider/qti.js
jsconan: I am afraid it is not possible. This function need to be accessible from everywhere in the component, not only from the init() method. However I can move it on top instead of bottom.

views/js/runner/provider/qti.js
jsconan: This will be difficult to put this part in the proxy, as this code needs to access to the runner instance.

views/js/runner/provider/qti.js
krampstudio: The leave flow is incorrect, and it will be difficult to change. What about the following refactoring : 

Inside the `init` method : 
```js
this.on('leave', function(){
      this.trigger('destroy');
});
```

inside a plugin for example : 
```js
runner.before('leave', function(e, data){
  var done = e.done();
    var context = runner.getTestContext();
       runner.disableItem(context.itemUri);
        runner.trigger('alert', data.message, function () {
                runner.trigger('endsession', 'teststate', data.code);
                done();
        });
});
```

jsconan:
Changes made: 
- to isolate the communicator into the proxy
- to use plugins for exit messages and test state events

PR #455, opened 2016-05-11
Tao 2600 gradual timer warnings

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-2600
https://github.com/pcgroup-us/extension-tao-pcg/pull/4
https://github.com/oat-sa/extension-tao-act/pull/224


krampstudio:
@jsconan please check the impacts for taoAct.

jsconan:
I see two possible issues:

- The old test runner will not work anymore with the new timer config format. Could you also update the old test runner?
- Some customer extensions also push some dedicated config for the timers. So please update those extensions too.

Alroniks:
@jsconan Please, take a look. Fixed your notices. And also https://github.com/oat-sa/extension-tao-act/pull/224

jsconan:
I have an issue with the new warning format, as well as in the new TR and the old one.

With the previous timer version, when I tried to run an assessment with a timer below the warning time, I got a warning and the timer changed its color.

With the new version the timer is still in normal mode and I do not get any warning.

Alroniks:
@jsconan Fixed and updated PR

jsconan:
Sorry, but I still have the issue.

I think the problem is here: https://github.com/oat-sa/extension-tao-testqti/pull/455/files#diff-998efc74cdd9f9a21d59e66f31dc4ebaR123

You define the warning as already shown when the remaining time is below the timer threshold. Thus the warning will never be displayed when you resume a test that has a lower remaining timer than the warning threshold. But this warning must be displayed, as well as the displayed time must have its style changed to reflect the warning.

Lets see a simple example.

warning threshold: 300
remaining timer: 60
the warning must be displayed as soon as the test is resumed/started 

warning threshold: 60
remaining timer: 120
when the test is resumed/started no warning is displayed, but it will when the remaining time will reach the threshold

Note: when several threshold are involved, only the closest one must be displayed instead of all warnings messages, and it must also affect the visual style of the timer

jsconan:
Sorry, please forgot what I said. I have reset my config to the previous format to work on another task, and I did not update it while checking you new work.

It work perfectly nice now.

PR #456, opened 2016-05-13
PHP7 Support

Alroniks:


jbout:
* This change requires a version upgrade, and should reference
* lib-tao-qti is now indirectly required by including taoQtiItem

PR #458, opened 2016-05-18
Feature/tao 2584 disconnection detection

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2584

Requires: https://github.com/oat-sa/tao-core/pull/832


views/js/controller/runner/runner.js
```diff
@@ -89,7 +89,7 @@ define([
                 //TODO move the loading bar into a plugin
                 loadingBar.stop();
             })
-            .on('destroy', function () {
+            .before('destroy', function () {
```
krampstudio: do you want to prevent other `destroy` listeners to execute  ?

jsconan: Oops, my mystake. It should rather be `after`. I fixed this.

models/classes/runner/session/TestCsrfToken.php
```diff
@@ -0,0 +1,151 @@
...
+    
+    /**
+     * Generates and returns the CSRF token
+     * @return string
+     */
+    public function getCsrfToken()
+    {
+        $service = self::getExtendedStateService();
+
+        $token = bin2hex(openssl_random_pseudo_bytes(self::TOKEN_LENGTH / 2));
```
krampstudio: not a blocker, but `openssl` calls are known to be slow on non-unix OS. As the token is generated per request, this can be a bottleneck. (To be verified)

views/js/runner/plugins/controls/connectivity/connectivity.js
```diff
@@ -0,0 +1,107 @@
...
+
+            function reconnect() {
+                if (testRunner.getState('disconnected')) {
+                    testRunner.trigger('reconnect');
+                }
+            }
+
+            // immediate detection of connectivity loss using Offline API
+            $(window).on('offline.connectivity', function() {
+                testRunner.trigger('error', {
```
krampstudio: Shouldn't you trigger a `disconnect` here ?

models/classes/runner/session/TestCsrfToken.php
jsconan: Yes, I know SSL is slower, but this is a recommendation from the PHP doc to generate secured tken: http://php.net/manual/en/function.uniqid.php

views/js/runner/plugins/controls/connectivity/connectivity.js
jsconan: An error is triggered here to simplify the design, in order to use only one way to handle network errors.

The `disconnect` event is just an effect of the network error handling.

models/classes/runner/session/TestCsrfToken.php
krampstudio: [Owasp recommands](https://www.owasp.org/index.php/PHP_CSRF_Guard) to use `random_bytes` or [paragonie/random_compat](https://github.com/paragonie/random_compat) as a polyfill for previous versions. We had multiple performances pb with openssl at larger scale (check with the backend team for the policy regarding PHP polyfill).

I've also found this [bug](https://bugs.php.net/bug.php?id=70014) which is fixed only from PHP5.6.

PR #460, opened 2016-05-19
Feature/tao 2611 calculator plugin

ssipasseuth:
requires https://github.com/oat-sa/tao-core/pull/833

### DIFF
--- a/views/js/runner/plugins/tools/calculator.js
+++ b/views/js/runner/plugins/tools/calculator.js
@@ -45,13 +45,6 @@ define([
          * Initialize the plugin (called during runner's init)
          */
         init : function init(){
-
-        },
-        /**
-         * Called during the runner's render phase
-         */
-        render : function render(){
-
             var self = this;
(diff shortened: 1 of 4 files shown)


views/js/runner/plugins/tools/calculator.js
```diff
@@ -0,0 +1,185 @@
...
+                .on('unloaditem', function (){
+                    self.disable();
+                    if(self.calculator){
+                        //destroy calculator to create a new instance of calculator each time
+                        self.calculator.destroy();
+                        self.calculator = null;
+                    }
+                });
+
+            areaBroker.getToolboxArea().append(this.$button);
```
krampstudio: Could you move everything above into the `init` method. The goal is to make the render step the shortest (only append to the dom), when element creation, event binding, etc. are supposed to be made during the initialization

ssipasseuth:
ok, done; moved the DOM initialization into the method init()

PR #462, opened 2016-05-23
Feature/send variables

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2585

Requires: https://github.com/oat-sa/extension-tao-test/pull/115

Ensure the variables are sent before the test is finished.
Add a `flush` step in the runner's life cycle. This step is placed before the destroy. 

Now the workflow is:
- event `leave` *or* event `finish` -> flush() -> event `flush` -> destroy() -> event `destroy` -> browser redirect

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -33,10 +33,10 @@ return array(
     'label' => 'QTI test model',
 	'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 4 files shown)


krampstudio:
please update the manifest to reflect the dependency to taoTests 2.18.0

PR #467, opened 2016-05-25
Tao 2600 colored timers

Alroniks:
https://oat-sa.atlassian.net/browse/TAO-2600

https://github.com/oat-sa/tao-core/pull/841
https://github.com/oat-sa/extension-tao-itemqti/pull/698
https://github.com/oat-sa/extension-tao-testqti/pull/467
https://github.com/oat-sa/extension-tao-devtools/pull/60


krampstudio:
I think there's no need to display timer warnings in feedback boxes others than the warning style otherwise the user will be confused because he will get a "timer warning" in an error or a success feedback. From a GUI perspective, we keep the `feedback.error` only for errors, and so on.
You should always warn using the warning feedback but you can change the color of the timer text.

For the timer text style, please do not change the element class property using Regexp, you can use instead the style name directly as class name : `warning` instead of `qti-timer__warning`, `error` instead of qti-timer__error`, etc.

views/js/controller/runtime/testRunner.js
```diff
@@ -720,13 +720,15 @@ function (
...
-                                    cst.warnings = self.testContext.timerWarning[cst.qtiClassName];
-                                    _(cst.warnings).forEach(function (value, key) {
-                                        cst.warnings[key] = {
-                                            type: value,
-                                            showed: cst.seconds <= key,
-                                            point: parseInt(key, 10)
-                                        };
+                                    cst.warnings = {};
+                                    _(self.testContext.timerWarning[cst.qtiClassName]).forEach(function (value, key) {
+                                        if (_.indexOf(['info', 'warning', 'danger'], value) != -1) {
```
krampstudio: Use either `(_.indexOf(['info', 'warning', 'danger'], value) !== -1)` (to use consistenl rules)  or `(_.contains(['info', 'warning', 'danger'], value))` if you don't mind of the index

views/js/runner/plugins/controls/timer/timer.js
```diff
@@ -114,19 +114,21 @@ define([
...
                         };
 
-                        _(timer.warnings).forEach(function (value, key) {
-                            timer.warnings[key] =  {
-                                type: value,
-                                showed: timer.remaining / precision <= key,
-                                point: parseInt(key, 10) * precision
-                            };
+                        _(timerWarning[timeConstraint.qtiClassName]).forEach(function (value, key) {
+                            if (_.indexOf(['info', 'warning', 'danger'], value) != -1) {
```
krampstudio: same here

krampstudio:
A last change to be done : I've just seen the color is applied to the timer title when it should only apply to the timer itself (target `.qti-timer_time`)

Alroniks:
@krampstudio fixed

PR #468, opened 2016-05-25
TAO-2625 Allow skipping has not effect

Alroniks:
https://oat-sa.atlassian.net/browse/TAO-2625


Alroniks:
@antoinerobin Please, take a look again

models/classes/runner/QtiRunnerService.php
```diff
@@ -589,6 +588,30 @@ public function storeItemResponse(RunnerServiceContext $context, $itemRef, $resp
...
+                $similar = 0;
+
+                /** @var ResponseVariable $responseVariable */
+                foreach ($responses as $responseVariable) {
+                    $defaultValue = $responseVariable->getDefaultValue();
+                    $value = $responseVariable->getValue();
+
+                    if (!$defaultValue) {
+                        $class = get_class($value);
+                        $defaultValue = new $class($responseVariable->getBaseType(), array());
```
antoinerobin: It doesn't work for QtiScalar response type such as QtiInteger for a slider interaction for example. The QtiScalar have only one parameter in the construct which is the value. maybe you should talk to @bugalot to have more insight about this part

bugalot: Actually @antoinerobin, @Alroniks, it works even with scalar-based items. Don't know why yet, but it works.

PR #483, opened 2016-06-14
Fix/duration reflection issue

krampstudio:
With the new version of qtism, if you try to save a test using the Creator you had an error, because of a wrong type reflection.


krampstudio:
Anyone can take it, but this is a blocking issue.

bugalot:
Excellent, thanks to have taken this into account while I was out of office !

PR #485, opened 2016-06-15
Fix/get from item types

antoinerobin:
depends on : https://github.com/oat-sa/extension-tao-itemqti/pull/714


bugalot:
Tested by importing a QTI package with items (TAO Sample package)

PR #493, opened 2016-06-24
Adapted rest interface and added doc

jbout:


jbout:
wrong destination, missing update script

PR #494, opened 2016-06-24
Adapted rest interface and added doc

jbout:


actions/class.RestQtiTests.php
```diff
@@ -37,57 +43,35 @@ public function __construct()
...
-     * @author Rashid Mumtaz & Absar - PCG Team - {absar.gilani6@gmail.com & rashid.mumtaz372@gmail.com}
-     * @return returnSuccess and returnFailure
-     */
-    protected function post()
-    {
-        $data = $this->importQtiPackage(tao_helpers_Http::getUploadedFile("qtiPackage"));
-        if ($data->getType() === common_report_Report::TYPE_ERROR) {
-            $e = new common_exception_Error($data->getMessage());
-            return $this->returnFailure($e);
         } else {
```
siwane: COuld u delete this else? there is a return before

jbout: I would like to keep the if to make the control structure easier to read

chrisphpgarcia: Mising or invalid QTI test package => Missing or invalid QTI test package

I don't get a 400 error code when i upload an invalid package. juste response code 200 without content.

PR #497, opened 2016-06-29
zoom out | zoom in for new testRunner

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-2689

### DIFF
--- a/views/js/runner/plugins/tools/zoom.js
+++ b/views/js/runner/plugins/tools/zoom.js
@@ -1,0 +1,247 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


no-chris:
I have 2 strange behaviors:
- when zooming out, item remains centered, but when zooming in, item is not centered anymore
- when moving through items in a delivery, the zoom level is reset. but when I click on zoom in or zoom out, it apply the zoom action from the last zoom level in the previous item. To reproduce:
1- zoom out to a significant level (5 or 6 times)
2- go to next item
3- zoom in

views/js/runner/plugins/tools/zoom.js
```diff
@@ -0,0 +1,247 @@
...
+            });
+
+            //attach behavior
+            this.$buttonZoomIn.on('click', function (e){
+                e.preventDefault();
+                zoomAction(1);
+            });
+            
+            //start disabled
+            togglePlugin();
```
no-chris: Please just use this.show() and remove the togglePlugin function as the plugin will be registered from client extension

zagovorichev:
1. when zooming out, item remains centered, but when zooming in, item is not centered anymore
That because we always have left and top borders
2. I don't know what is the correct behavior, for now - zoom set as default for each item

PR #499, opened 2016-07-01
Support for test import using flysystem

jbout:


jbout:
abandoned in favor of refactor-fs

PR #508, opened 2016-07-12
Fix/tao 2734 mark for review overwritten

krampstudio:
since the ExtendedState is accessed across multiple instances, the cache is shared accross them...

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -441,6 +441,6 @@ class Updater extends \common_ext_ExtensionUpdater {
             $this->setVersion('3.1.0');
         }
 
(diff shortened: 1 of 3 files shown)


scripts/update/Updater.php
```diff
@@ -440,5 +440,7 @@ public function update($initialVersion) {
             
             $this->setVersion('3.1.0');
         }
+
+        $this->skip('3.0.0', '3.0.1');
```
antoinerobin: the skip should be from version 3.1.0 to 3.1.1

krampstudio:
up to date

PR #511, opened 2016-07-13
Removed old QTI Test FileSystem. New one with ID added.

bugalot:


hutnikau:
Testers here (in ACT) asked me to check why they can't export tests (they got 500 error) and after investigation I found that problem in filesystem. As far as I understand this PR fixes this issue, right? Is this fix still under development?

PR #519, opened 2016-07-18
Show/hide the review screen when the item is loaded

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2774

The review screen was only shown at render time, so if it is enabled later, it will not be displayed until a refresh.

This patch shows/hides the review screen for each loaded item.

### DIFF
--- a/views/js/runner/plugins/controls/review/review.js
+++ b/views/js/runner/plugins/controls/review/review.js
@@ -192,12 +192,15 @@ define([
                 }
             });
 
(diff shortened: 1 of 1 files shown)


krampstudio:
This conflicts with https://oat-sa.atlassian.net/browse/TAO-2736 that requires to keep the review pane hidden across items.

jsconan:
Ok, this other story is now taken into account.

PR #523, opened 2016-07-20
Fix/tao 2809 concurrency issues

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2809

Requires: https://github.com/oat-sa/extension-tao-test/pull/132

The issue can be reproduced with the blurPause plugin enabled:
1. start/resume a test
2. finish it
3. while it finishing, do not wait and loose the focus (click on another tab, selection of the console panel, etc.)
4. the finish action has been called twice

### DIFF
--- a/views/js/runner/provider/qti.js
+++ b/views/js/runner/provider/qti.js
@@ -587,17 +587,17 @@ define([
                 this.before('flush', function(e) {
                     var done = e.done();
 
-                    this.getProxy().callTestAction('finish').then(function() {
-                        if (self.stateStorage) {
-                            self.stateStorage.clear()
-                                .then(done)
-                                .catch(function(err) {
-                                    self.trigger('error', err);
-                                })
-                        } else {
-                            done();
-                        }
-                    });
+                    this.getProxy()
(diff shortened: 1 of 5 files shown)


views/js/runner/provider/qti.js
```diff
@@ -569,34 +569,37 @@ define([
...
+                this.trigger('disablenav disabletools')
+                    .trigger('endsession', 'finish');
+
+                // will be executed after the runner has been flushed
+                // use the "before" queue to ensure the query will be fully processed before destroying
+                // we do not use the "destroy" event because the proxy is destroyed before this event is triggered
+                this.before('flush', function(e) {
+                    var done = e.done();
+
+                    this.getProxy().callTestAction('finish').then(function() {
```
krampstudio: 1st level promise not catched, what about : 
```js
 this.getProxy()
       .callTestAction('finish')
       .then(function() {
           if (self.stateStorage) {
              return self.stateStorage.clear();
           }
       })
       .then(done)
       .catch(function(err) { 
           self.trigger('error', err); 
       });
```

PR #530, opened 2016-07-26
File should be deleted instead of directory

boajer:
Issue is easily reproducible during delivery compilation based on  test with rubrics


boajer:
closed in favor of https://github.com/oat-sa/tao-core/pull/928

PR #539, opened 2016-08-04
Fix/tao 2820 pause on disconnect

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2820

Enhance the pause handling.

I also fixed some ESLint errors and warnings.

### DIFF
--- a/views/js/runner/provider/qti.js
+++ b/views/js/runner/provider/qti.js
@@ -380,12 +380,21 @@ define([
                             self.trigger('error', err);
                         });
                 })
-                .on('pause', function(){
-                    // will notify the server that the test was auto paused
-                    self.getProxy().callTestAction('pause')
+                .on('pause', function(data){
+                    var pause;
+
+                    if (!self.getState('disconnected')) {
+                        // will notify the server that the test was auto paused
+                        pause = self.getProxy().callTestAction('pause');
+                    } else {
+                        pause = Promise.resolve();
+                    }
+
+                    pause
                         .then(function() {
                             self.trigger('leave', {
-                                code: self.getTestData().states.suspended
+                                code: self.getTestData().states.suspended,
+                                message: data && data.message
                             });
                         })
                         .catch(function(err){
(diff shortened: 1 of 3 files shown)


views/js/runner/provider/qti.js
```diff
@@ -175,9 +175,9 @@ define([
             /**
              * Compute the next item for the given action
              * @param {String} action - item action like move/next, skip, etc.
-             * @param {Object} [params] - the item action additionnal params
+             * @param {Object} [params] - the item action additional params
              */
-            var computeNext = function computeNext(action, params){
+            function computeNext(action, params){
```
krampstudio: Why did you unassign the functions ?

views/js/runner/provider/qti.js
```diff
@@ -292,171 +293,183 @@ define([
...
-                 */
-                var getItemLabel = function getItemLabel(){
-                    var item = mapHelper.getItem(self.getTestMap(), context.itemIdentifier);
-                    return item && item.label ? item.label : context.itemIdentifier;
-                };
+                .on('pause', function(data){
+                    var pause;
 
-                this.trigger('enablenav enabletools');
+                    if (!self.getState('disconnected')) {
```
krampstudio: just thinking, but should'n we check the "disconnected" state directly within the proxy ? for all requests ?

views/js/runner/provider/qti.js
jsconan: This is related to ESLint errors fixing. Using variables functions and calling them before they are declared make ESLint unhappy.
In this code there is no need to use variable functions, as there is no reassignment. So I preferred to go to simple declaration instead of moving them all along to make the errors disappear.

views/js/runner/provider/qti.js
jsconan: Yes we could, but in that case we need to handle the disconnected state within the proxy. For now the connectivity state is handler at the provider level, and the proxy does not have access to.
To do this properly we should redesign a little the proxy.

PR #542, opened 2016-08-05
adding auto completion for long (long) test category

ssipasseuth:
(low priority PR)
Because the special predefined categories are difficult to remember, I propose to add this autocompletion feature at little cost.


jsconan:
Very nice feature! The one I have wanted for a while.
Maybe a way of improvement: provide a list of extra categories to take care of the customer's additions

PR #546, opened 2016-08-09
Fixed modalFeedback for items

zagovorichev:


ssipasseuth:
Hi, 
Is there another PR related to TAO-2665 ?
I indeed expected some change in https://github.com/oat-sa/extension-tao-itemqti/blob/master/views/js/runner/provider/qti.js#L91-L101 to allow adding support for both inline and modal display modes for <modalFeedback> as implemented in the old item runner : https://github.com/oat-sa/extension-tao-itemqti/blob/master/views/js/qtiRunner/core/QtiRunner.js#L298-L308

zagovorichev:
Hi, @ssipasseuth 
first of all I decide to fix old feedback form, because it should work, there are little bugs.
After that I'll going to make changes in the taoItemQti (I have local branch)

PR #552, opened 2016-08-18
Feature/configurable force branch preconditions

bugalot:
This PR aims at implementing TAO-2933. Configuration methods have been added to QTI-SDK legacy (QTI-SDK 0.12.12, required by lib-tao-qti 1.8.0, required by extension-item-qti 4.1.0) to force branchRules/preconditions to be executed even if the current navigation mode of a QTI Test is NonLinear.

In addition, a path tracking system has been integrated to QTI-SDK 0.12.12 in order rely on previously visited items for forward/backward navigation, instead of inspecting the regular item route flow.

To make sure this PR works, please make sure you are up to date with lib-tao-QTI 1.8.0.

A sample QTI Test has been included in this PR, so that you can test the behaviour of this new feature by changing config in testRunner.conf. This is the forceBranchRule.zip sample. By entering this test, you will be shown a flow diagram representing the expected flow when the feature is turned on/off.

Please triple check that update scripts are working, and that currently running test executions do not break after update.

@jsconan Could you please review the code?
@antoinerobin Could you please this PR functionally, using the forceBranchRules.zip test sample?

Warning!!!! @antoinerobin will be the final merger. Make sure that it has been reviewed by JS first. @jsconan just tell us about your code review but please do not merge the PR.

The testRunner.conf params to be turned to true/false in order to test this functionnality are:
- force-branchrules
- path-tracking

Be careful!!!! Changing the configuration parameter while running a test may lead to unexpected behaviour, as this feature is designed to be platform-wide, and to run from the beginning of a specific customer project.


helpers/class.SessionManager.php
```diff
@@ -111,13 +111,22 @@ public function getTest() {
...
         } else {
             $className = $config['test-session'];
         }
-        return new $className($test, $this, $route, $this->getResultServer(), $this->getTest());
+        
+        $assessmentTestSession = new $className($test, $this, $route, $this->getResultServer(), $this->getTest());
+        
+        $assessmentTestSession->setForceBranching($config['force-branchrules']);
+        $assessmentTestSession->setForcePreconditions($config['force-preconditions']);
+        $assessmentTestSession->setPathTracking($config['path-tracking']);
```
jsconan: Can lead to issue if the config does not contains the attributes. It should be better  to check before with some `isset()`

jsconan:
The code is ok for me. You can proceed ahead.

PR #565, opened 2016-08-31
Refactor/tao 2941 test model

antoinerobin:
depends on : https://github.com/oat-sa/extension-tao-test/pull/143

### DIFF
--- a/config/default/TestModel.conf.php
+++ b/config/default/TestModel.conf.php
@@ -6,11 +6,11 @@
  */
 
 return new \oat\taoQtiTest\models\TestModelService(array(
-    'exportHandler' => array(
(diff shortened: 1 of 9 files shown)


config/default/TestModel.conf.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
+ * Default config header
+ *
+ * To replace this add a file taoQtiTest/conf/header/TestModel.conf.php
+ */
+
+return new \oat\taoQtiTest\models\TestModelService(array(
+    'exportHandlers' => array(
+        new taoQtiTest_models_classes_export_TestExport(),
```
jbout: The export handlers need to be properly serializable

PR #566, opened 2016-09-01
Feature/tao 3008 autocomplete timeout

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3008

Requires: https://github.com/oat-sa/tao-core/pull/976

Use namespace for alert/confirm events
Use the new async events syntaxe.

The dialogs can now be closed by the `closedialog` event.  The namespace allows to only target particular kinds of dialogs.

```javascript
// display simple alerts
runner.trigger('alert.hello', 'Hello!');
runner.trigger('alert.welcome', 'Welcome here');

// close only .hello dialog
runner.trigger('closedialog.hello');

// close all dialogs
runner.trigger('closedialog');


// display confirms
runner.trigger('confirm.want', 'Do you want?', function() {
    console.log('sure you want!');
}, function() {
    console.log('oh! really?');
});

runner.trigger('confirm.quit', 'You will leave if you hit Ok'), function() {
    console.log('Bye!');
}, function() {
    console.log('Take a sit...');
});

runner.trigger('confirm.stupid', 'Are you a piece of cake ?'), function() {
    console.log("Oh... that's your life");
}, function() {
    console.log('Are you sure?');
});

// close only .want dialog, with reject
runner.trigger('closedialog.want');

// close only .quit dialog, with accept
runner.trigger('closedialog.quit', true);

// close all dialogs, with reject
runner.trigger('closedialog');
```

### DIFF
--- a/views/js/runner/plugins/content/dialog/dialog.js
+++ b/views/js/runner/plugins/content/dialog/dialog.js
@@ -23,11 +23,12 @@
  */
 define([
     'jquery',
+    'lodash',
     'i18n',
     'taoTests/runner/plugin',
     'ui/dialog/alert',
-    'ui/dialog/confirm',
-], function ($, __, pluginFactory, dialogAlert, dialogConfirm){
+    'ui/dialog/confirm'
+], function ($, _, __, pluginFactory, dialogAlert, dialogConfirm){
     'use strict';
 
     /**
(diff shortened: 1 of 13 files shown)


views/js/runner/plugins/content/dialog/dialog.js
```diff
@@ -40,13 +41,64 @@ define([
...
+
+                dialog.on('closed.modal', function() {
+                    removeHandle(stack, dialog);
+                });
+            }
+
+            function removeHandle(stack, dialog) {
+                _.forEach(stack, function(handle, index) {
+                    if (handle && dialog && dialog === handle.dialog) {
+                        stack[index] = null;
```
krampstudio: Is there a reason why you set the entry to `null` and then doing a `cleanup` ? Do you need to keep the indexes ? Otherwise, you can use `_.remove`

PR #567, opened 2016-09-05
Feature/tao 2665/modal feedback

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-2665


krampstudio:
The workflow isn't correct, especially the communication between the TR and IR. Here is the correct one : 

1. The test runner submit and item and has `result.displayFeedbacks` in the response. 
2. Then give the feedback to the item runner to process it, something like `var feedBackContent = itemRunner.renderFeedback(feedbacks)` that produces the HTML of the feedback.
3. Trigger an event from the test runner with the feedback : `this.trigger('feedback', feedBackContent)`
4. The registered plugins takes care of rendering the feedback. For example inside the modalFeedback plugin, you listen for the `feedback` event to render it into a dialog.

zagovorichev:
@krampstudio Not tested all with qunit test, but check please, maybe something incorrect in implementation

zagovorichev:
@krampstudio Update, check it, please

krampstudio:
Sorry, to give you a late feedback on that but I realize that having only 2 plugins would simplify everything a lot. Except of having a modalFeedback plugin that instantiate sub plugins based on the configuration, why don't we have only 2 plugins : 
 - modalFeedback
 - inlineFeedback 

modalFeedback would be activated by default, and if one wants the inlineFeedback on it's platform he need to uninstall the modalFeedback plugin and install the inlineFeedback plugin.

Again, I know this is a late feedback  but it will prevent us from checking the mode when this is something that doesn't change.

views/js/runner/provider/qti.js
```diff
@@ -256,10 +271,21 @@ define([
                                     context.itemAnswered = result.itemSession.itemAnswered;
                                 }
 
-                                if(result.displayFeedbacks === true && itemRunner){
-                                    return itemRunner.trigger('feedback', result.feedbacks, result.itemSession, resolve);
+                                if(result.displayFeedbacks === true && result.feedbacks && result.itemSession){
+
+                                    itemRunner.renderFeedbacks(result.feedbacks, result.itemSession, function(queue){
+                                        self
```
krampstudio: Try to avoid calling plugin events. 

You can give a callback to the modalFeedback event: 

```js
self.trigger('modalFeedbacks', queue, resolve);
```

and in the plugin 

```js
this.getTestRunner().on('modalFeedback', function(queue, done){

});
```

views/js/runner/provider/qti.js
```diff
@@ -241,6 +255,7 @@ define([
                     return self.getProxy().submitItem(context.itemUri, itemRunner.getState(), itemRunner.getResponses(), params)
                         .then(function(result){
                             return new Promise(function(resolve, reject){
+                                var feedbackLoaded;
```
krampstudio: Where is it used ?

zagovorichev:
I thought that: now it is only QTI modalFeedback implementation (https://www.imsglobal.org/question/qtiv2p1/examples/items/Example01-modalFeedback.xml)
we just make 2 ways to display it

But in the future we could implement inlineFeedbacks (https://www.imsglobal.org/question/qtiv2p1/examples/items/Example02-feedbackInline.xml)

views/js/runner/provider/qti.js
```diff
@@ -35,7 +35,21 @@ define([
     'taoItems/assets/manager',
     'taoItems/assets/strategies',
     'tpl!taoQtiTest/runner/provider/layout'
-], function($, _, __, store, Promise, cachedStore, areaBroker, proxyFactory, probeOverseerFactory, mapHelper, qtiItemRunner, assetManagerFactory, assetStrategies, layoutTpl) {
+], function(
```
krampstudio: Please do not auto format the header...

views/js/test/runner/plugins/content/dialog/itemAlertMessage/test.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Plugin for the output of the messages in the item in test runner as a modal window</title>
+    <link rel="stylesheet" type="text/css" href="/tao/views/js/lib/qunit/qunit.css">
```
krampstudio: Use relative paths instead of `/` except for `/tao/ClientConfig/config'`

views/js/test/runner/plugins/content/dialog/itemAlertMessage/test.html
```diff
@@ -0,0 +1,28 @@
...
+    <script type="text/javascript" src="/tao/views/js/lib/require.js"></script>
+    <script type="text/javascript" src="/tao/views/js/lib/qunit/qunit.js"></script>
+    <script type="text/javascript" src="/tao/views/js/lib/qunit/qunit-parameterize.js"></script>
+    <script type="text/javascript" src="/tao/views/js/lib/blanket/blanket.min.js" data-cover-only="/taoQtiTest/views/js/runner/plugins/content/dialog/itemAlertMessage.js"></script>
```
krampstudio: add branch tracking

views/js/runner/provider/qti.js
zagovorichev: It was not auto... just too long string. Make it in one row?

views/js/runner/plugins/content/dialog/itemInlineMessage.js
```diff
@@ -0,0 +1,138 @@
...
+        }
+    };
+
+    /**
+     * Create the button based on the current context
+     * @param {Object} selfPlugin
+     * @param {Object} context - the test context
+     * @returns {*|jQuery|HTMLElement} the button
+     */
+    var createOkButton = function createElement(selfPlugin, context){
```
krampstudio: Move this function into `init`so you can benefit of the closure's context instead of giving the context as 1st parameter

views/js/runner/plugins/content/modalFeedback/modalFeedback.js
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * All feedback messages
+     */
+    var renderedFeedbacks;
+
+    /**
+     * modalFeedback was resolved and all components were destroyed
+     */
+    var isDestroyed;
```
krampstudio: Please use the [plugin's state](https://github.com/oat-sa/tao-core/blob/master/views/js/core/plugin.js) instead. For example you can check if the state is still in `init` to destroy.

views/js/runner/plugins/content/modalFeedback/modalFeedback.js
```diff
@@ -0,0 +1,185 @@
...
+         */
+        render: function render() {
+            var self = this;
+            var testRunner = this.getTestRunner();
+
+            var createMessages = function createMessages(renderingQueue, inline){
+
+                var bInlineMode = inlineMode;
+
+                if (inline === true || inline === false) {
```
krampstudio: did you mean `_.isBoolean` ?

views/js/runner/plugins/content/modalFeedback/modalFeedback.js
```diff
@@ -0,0 +1,185 @@
...
+                nextStep = done;
+                createMessages(renderingQueue, inline);
+            });
+        },
+
+        /**
+         * Called during the runner's destroy phase
+         * allow to run that function only once
+         */
+        destroy: function destroy() {
```
krampstudio:
```js
if(this.getState('init')){
   _.forEach(renderedFeedbacks, destroyFeedback);
}
```

views/js/runner/plugins/content/modalFeedback/modalFeedback.js
```diff
@@ -0,0 +1,185 @@
...
+        if (removed) {
+            feedback.destroy();
+
+            if (!renderedFeedbacks.length) {
+                nextStep();
+            }
+        }
+    }
+
+    function defineMode(inline) {
```
krampstudio: I'm not sure what you are doing by changing the state of those module wide variables ? 
Shouldn't you do the inverse by checking the current configuration using an `isInline()` function ?

zagovorichev: This is for qunit tests https://github.com/oat-sa/extension-tao-testqti/blob/0a303b3f4cd1b5fbb32e48bc29c5b8cebc55be43/views/js/test/runner/plugins/content/modalFeedback/test.js#L437-L437 
I can determine what exactly I want show, even if in the module config will be default value

views/js/runner/plugins/content/modalFeedback/modalFeedback.js
zagovorichev: It isn't works for me, example: modalFeedback creates 2 inlineMessages with only 1 ok button, when I press it should be runned modalFeedback.destroy(), but only once
step by step:
1. click on the OK button
https://github.com/oat-sa/extension-tao-testqti/blob/63529ce60329271550a69b6b57e8dc9a8b174390/views/js/runner/plugins/content/dialog/itemInlineMessage.js#L87-L87
2. for inline mode we have only one button, so needs to be run modalFeedback.destroy()
https://github.com/oat-sa/extension-tao-testqti/blob/39d77297ccee69eb09041ae9fc47792ccac6617e/views/js/runner/plugins/content/modalFeedback/modalFeedback.js#L154-L154
3. Destroy each of the feedback will halt his destroy (which again run click on the ok button for current feedback)

views/js/runner/plugins/content/modalFeedback/modalFeedback.js
zagovorichev: but in the destroyFeedback will be called modalFeedback.destroy before the previous deletion change plugin status

zagovorichev: will try to explain again: really I create ok button for each message and I can close each of them separately, but for navigation I show only first of them and when I press this button from navigation, I call click on each feedback

krampstudio:
During a test, all feedbacks are displayed, not only the correct one.

PR #568, opened 2016-09-07
Expose more details about errors to logs

boajer:
Not sure if all of the details should be exposed to end user. 
Inspired by solving https://github.com/oat-sa/qti-sdk/pull/57

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -33,7 +33,7 @@ return array(
     'label' => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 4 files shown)


jbout:
Those details should indeed not be exposed to the enduser, but:
* Having them in the logs would be a great start
* We need to improve our error reporting, to at least show the item the error appeared on

boajer:
@jbout could you take a look again, now it just expose details into logs. Without changing end-user experience as for now.

jbout:
We will need to improve the error reporting on item level errors in the future

PR #569, opened 2016-09-08
Feature/tao 2873 informational items

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2873

Feature related to the new test runner and the review panel.

When the review panel is enabled, the items that are intended to be only informational are handled in a particular manner:
- a special icon is dedicated
- those items are not taken into account in the answered/flagged counts
- those items cannot be flagged

An item is considered as informational when it does not contains any responses (Text Block) or when the category `x-tao-itemusage-informational` is attached.

To enable this feature you will need to set the `check-informational` option in the taoQtiTest config file: `config/taoQtiTest/testRunner.conf.php`

### DIFF
--- a/views/js/runner/helpers/map.js
+++ b/views/js/runner/helpers/map.js
@@ -258,9 +258,9 @@ define([
          */
         each: function each(map, callback) {
             if (_.isFunction(callback)) {
-                _.each(map && map.parts, function(part) {
-                    _.each(part && part.sections, function(section) {
-                        _.each(section && section.items, function(item) {
+                _.forEach(map && map.parts, function(part) {
+                    _.forEach(part && part.sections, function(section) {
+                        _.forEach(section && section.items, function(item) {
                             callback(item, section, part, map);
                         });
                     });
--- a/views/js/runner/helpers/map.js
+++ b/views/js/runner/helpers/map.js
@@ -29,6 +29,7 @@ define([
      */
     function getEmptyStats() {
         return {
+            questions: 0,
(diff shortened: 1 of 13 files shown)


views/js/runner/plugins/navigation/review/review.js
```diff
@@ -109,6 +110,18 @@ define([
...
     /**
+     * Checks if the current context allows to mark the item for review
+     * @param {Object} testRunner
+     * @returns {Boolean}
+     */
+    function canFlag(testRunner) {
+        var context = testRunner.getTestContext();
+        var map = testRunner.getTestMap();
+        var item = mapHelper.getItemAt(map, context.itemPosition);
+        return !!(!context.isLinear && context.options.markReview && !(item && item.informational));
```
krampstudio: I think the `!!` casting isn't necessary

views/js/runner/helpers/map.js
```diff
@@ -204,6 +251,25 @@ define([
...
 
         /**
+         * Applies a callback on each item of the provided map
+         * @param {Object} map - The assessment test map
+         * @param {Function} callback(item, section, part, map) - A callback to apply on each item
+         * @returns {Object}
+         */
+        each: function each(map, callback) {
+            if (_.isFunction(callback)) {
+                _.each(map && map.parts, function(part) {
```
krampstudio: When need to always use the source function over the alias in order to ease the migration to the next version (the aliases are about to change)

PR #571, opened 2016-09-12
Fixed deletion of the inline feedback messages

zagovorichev:
and continue process after that


krampstudio:
Could you please remove the compiled files from the PR. We bundle them at release time to prevent conflicts and incorrect dependencies inclusions.

zagovorichev:
@krampstudio Done

PR #573, opened 2016-09-14
Fix/tao 3102 timer with dialogs

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3102

Let the timer continue when exit dialog is displayed

### DIFF
--- a/views/js/runner/plugins/controls/timer/timer.js
+++ b/views/js/runner/plugins/controls/timer/timer.js
@@ -334,20 +334,9 @@ define([
                             //check for new timers
                             updateTimers(true);
                         })
-                        .on('enableitem', function() {
-                            doEnable();
-                        })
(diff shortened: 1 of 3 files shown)


views/js/runner/plugins/controls/timer/timer.js
```diff
@@ -313,33 +331,25 @@ define([
                     //change plugin state
                     testRunner
                         .on('loaditem resumeitem', function(){
-
                             //check for new timers
                             updateTimers(true);
                         })
                         .on('enableitem', function() {
-                            self.enable();
+                            doEnable();
```
krampstudio: You don't need the wrapping function anymore

jsconan: You are right. Change done!

PR #593, opened 2016-09-26
Enable by default the review screen and the informational items

jsconan:
As requested, the "Review Panel" and the "Informational items" options are now enabled by default.


krampstudio:
Could you add a version update (feature), so we can associate this change to a version

jsconan:
Version added. I do not think an update script is needed, unless we want to force an update of every installed instances with the review panel and the informational items activated.

krampstudio:
You're right, this is only for new installs.

PR #597, opened 2016-09-29
Feature/tao 3136 add end test warning

no-chris:
https://oat-sa.atlassian.net/browse/TAO-3136


no-chris:
now working in new TR.
some files in the JIRA ticket might help you validate the PR

PR #600, opened 2016-10-04
Feature/tao 3144 add custom weight item fields

no-chris:
depends on https://github.com/oat-sa/tao-core/pull/1019
https://oat-sa.atlassian.net/browse/TAO-3144


views/js/controller/creator/views/itemref.js
```diff
@@ -20,24 +20,45 @@
...
-    * Set up an item ref: init action beahviors. Called for each one.
-    *
-    * @param {jQueryElement} $itemRef - the itemRef element to set up
-    * @param {Object} model - the data model to bind to the ref
-    */
-   var setUp =  function setUp ($itemRef, model){
+    /**
+     * We need to resize the itemref in case of long labels
+     */
+    var resize = _.throttle(function resize(){
```
krampstudio: This should be done using CSS (flexboxes should help in that situation).

views/js/controller/creator/templates/itemref-props.tpl
```diff
@@ -81,13 +81,39 @@
...
+                <div class="tooltip-content">
+                    {{__ 'Controls the contribution of an individual item score to the overall test score.'}}
+                </div>
+            </div>
+        </div>
+<!-- assessmentTest/testPart/assessmentSection/sectionPart/weights -->
+        <div data-bind-each="weights" data-bind-tmpl="itemrefweight"></div>
+        <div class="grid-row">
+            <div class="col-12">
+                <a class="itemref-weight-add"><span class="icon-add"></span> add a new weight</a>
```
krampstudio: Could you translate this string : `{{__ 'Add a new weight'}}`

views/js/controller/creator/views/itemref.js
no-chris: The function was there and has been left untouched. I just moved it to remove an eslint warning. Thus, I would prefer to address this issue in a separate PR.

PR #615, opened 2016-10-20
Fix/tao 3230 timed section

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3230

By default the TestRunner closes the session of each item contained by the timed section the TestTaler is leaving. This is not the standard wanted behavior.

This patch add an option to keep the timer until its timeout, even if we leave a timed section. So it is now possible to go back on a timed section and restore the timer to the state it had before leaving.

The name of the option: `keep-timer-up-to-timeout`.
This option is disabled by default, in order to prevent regression on existing installs.


jsconan:
@no-chris You are the final reviewer.

@bugalot Please validate the backend side.

bugalot:
@jsconan @no-chris 

I'm OK with the PHP code in use for this feature !

PR #623, opened 2016-10-28
Feature/tao 3248 runner shortcuts

jsconan:
Related to:
- https://oat-sa.atlassian.net/browse/TAO-3248
- https://oat-sa.atlassian.net/browse/TAO-3247

Add shortcuts to the following tools:
- calculator
- zoom
- theme switcher
- review panel
- mark for review
- comment

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -715,5 +715,35 @@ class Updater extends \common_ext_ExtensionUpdater {
 
             $this->setVersion('5.31.0');
         }
+        
+        if ($this->isVersion('5.31.0')) {
+
+            $extension = \common_ext_ExtensionsManager::singleton()->getExtensionById('taoQtiTest');
+
+            $config = $extension->getConfig('testRunner');
+            $config['shortcuts'] = [
+                'calculator' => [
+                    'toggle' => 'C',
+                ],
+                'zoom' => [
+                    'in' => 'I',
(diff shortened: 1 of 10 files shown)


scripts/update/Updater.php
```diff
@@ -715,5 +715,35 @@ public function update($initialVersion) {
...
+                ],
+                'itemThemeSwitcher' => [
+                    'toggle' => 'T',
+                ],
+                'review' => [
+                    'toggle' => 'R',
+                    'flag' => 'M'
+                ]
+            ];
+
```
no-chris: I'm wondering about conflicts with a client extension if the client update happens before, won't we lose its customisations ?

jsconan: I think this is the same problem with the plugins' config.
The script that updates the config has to be clever enough to avoid unwanted overwriting.

For instance:
```php
// keep other existing shortcuts
$testConfig['shortcuts']['calculator']['toggle'] = 'C';
```

Instead of:
```php
// loose existing shortcuts
$testConfig['shortcuts'] = [
    'calculator' => [
        'toggle' => 'C'
    ]
];
```

Or even:
```php
// still loose existing shortcuts
$testConfig = array_merge($testConfig, [
    'shortcuts' => [
        'calculator' => [
            'toggle' => 'C'
        ]
    ]
]);
```

PR #627, opened 2016-11-03
Feature/tao 3250 shortcut navigation

no-chris:


config/default/testRunner.conf.php
```diff
@@ -353,6 +353,12 @@
         'review' => [
             'toggle' => 'R',
             'flag' => 'M'
+        ],
+        'next' => [
+            'toggle' => 'J'
```
jsconan: Just wondering, is it really a toggle ?

scripts/update/Updater.php
```diff
@@ -747,5 +747,26 @@ public function update($initialVersion) {
...
         }
+
+        if ($this->isVersion('5.32.0')) {
+
+            $extension = \common_ext_ExtensionsManager::singleton()->getExtensionById('taoQtiTest');
+
+            $config = $extension->getConfig('testRunner');
+            if (!isset($config['shortcuts'])) {
+                $config['shortcuts'] = [];
+            }
```
jsconan: Not sure this check is relevant here, as the previous update should have set the `shortcuts` array entry. Anyway, this is not so important.

views/js/runner/plugins/navigation/next.js
```diff
@@ -96,41 +101,85 @@ define([
...
             this.$element = createElement(testRunner.getTestContext());
 
             //plugin behavior
-            this.$element.on('click', function(e){
-                var enable = _.bind(self.enable, self);
+            /**
+             * @param {Boolean} nextItemWarning - enable the display of a warning when going to the next item.
+             * Note: the actual display of the warning depends on other conditions (see nextWarningHelper)
+             */
+            function doNext(nextItemWarning) {
```
jsconan: You should keep the function declarations on top, mixing instructions and declaration can lead to bad reading.

views/js/runner/plugins/navigation/next.js
```diff
@@ -96,41 +101,85 @@ define([
...
+
+            this.$element.on('click', function(e){
+                e.preventDefault();
+                testRunner.trigger('nav-next');
             });
 
+            if(testConfig.allowShortcuts && pluginShortcuts.toggle){
+                shortcut.add(namespaceHelper.namespaceAll(pluginShortcuts.toggle, this.getName(), true), function(e) {
+                    if (self.getState('enabled') === true) {
+                        e.preventDefault();
```
jsconan: This instruction is not needed here, the shortcut helper already does this by default.
At least you should use the `prevent` option to be sure the event is prevented.

views/js/runner/plugins/navigation/next.js
```diff
@@ -144,6 +193,10 @@ define([
                 })
                 .on('disablenav', function(){
                     self.disable();
+                })
+                .on('nav-next', function(data) {
+                    var nextItemWarning = data && typeof(data[0]) !== 'undefined' ? data[0] : false;
```
jsconan: This is weird, why an array instead of a simple boolean parameter ?

views/js/runner/plugins/navigation/nextWarningHelper.js
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * @param {Object} options
+     * @param {Boolean} options.endTestWarning - enables the end test warning, when applicable
+     * @param {Boolean} options.isLast - if the item is the last of the test
+     * @param {Boolean} options.isLinear - if the current part is linear
+     * @param {Boolean} options.nextItemWarning - enables the next item warning, when applicable
+     * @param {Object} options.nextPart - description of the next part of the test
+     * @param {Number} options.remainingAttempts - remaining attempts for the current item
+     * @param {String} options.testPartId - current test part identifier
+     * @returns {Object}
```
jsconan: The API seems a little complex, why don't you just provide a reference to the runner and let the helper gathering the relevant data from the context ?

Also, if it is a helper, its place should be in the helper folder, not in plugins

views/js/runner/plugins/navigation/nextWarningHelper.js
```diff
@@ -0,0 +1,120 @@
...
+            warnBeforeNext = shouldWarnBeforeNext(),
+            warnBeforeEnd = shouldWarnBeforeEnd();
+
+        /**
+         * Convert a value to a boolean
+         * @param {*} value
+         * @param {Boolean} defaultValue
+         * @returns {Boolean}
+         */
+        function toBoolean(value, defaultValue) {
```
jsconan: This function could be external to its current scope, as there is nothing related to a local variable. Moreover this would ease the read as this function is used in var declarations

views/js/runner/plugins/navigation/previous.js
```diff
@@ -66,15 +79,46 @@ define([
...
             //attach behavior
-            this.$element.on('click', function(e){
-                e.preventDefault();
-                if(self.getState('enabled') !== false){
-                    self.disable();
+            function doPrevious(previousItemWarning) {
+                var context = testRunner.getTestContext();
 
-                    testRunner.previous();
+                previousItemWarning = previousItemWarning || false;
```
jsconan: Is it useful ? If the `previousItemWarning` variable is already a falsey, what is the purpose here ? type enforcement optimisation ?

views/js/runner/plugins/navigation/next.js
```diff
@@ -96,41 +101,85 @@ define([
...
                                 'test', testRunner),
-                            _.partial(doNext, context), // if the test taker accept
+                            _.partial(triggerNextAction, context), // if the test taker accept
+                            enable  // if the test taker refuse
+                        );
+
+                    } else if (warningHelper.shouldWarnBeforeNext()) {
+                        testRunner.trigger(
+                            'confirm.next',
+                            __('You are about to go to the next item. Click OK to continue and go to the next item.'),
```
jsconan: The message is redundant...
What about something like:
"You are about to go to the next item. You may not be able to go back to the current item. Continue anyway ?"

views/js/runner/plugins/navigation/previous.js
```diff
@@ -66,15 +79,46 @@ define([
...
+                    testRunner.trigger('enablenav enabletools');
                 }
+
+                if(self.getState('enabled') !== false){
+                    testRunner.trigger('disablenav disabletools');
+
+                    if (previousItemWarning && context.remainingAttempts !== -1) {
+                        testRunner.trigger(
+                            'confirm.previous',
+                            __('You are about to go to the previous item. Click OK to continue and go to the previous item.'),
```
jsconan: The message is redundant...
What about something like:
"You are about to go to the previous item. You may not be able to go back to the current item. Continue anyway ?"

views/js/runner/plugins/navigation/previous.js
```diff
@@ -66,15 +79,46 @@ define([
...
+
+            this.$element.on('click', function(e){
+                e.preventDefault();
+                testRunner.trigger('nav-previous');
             });
 
+            if(testConfig && testConfig.allowShortcuts){
+                shortcut.add(namespaceHelper.namespaceAll(pluginShortcuts.toggle, this.getName(), true), function(e) {
+                    if (canDoPrevious() && self.getState('enabled') === true) {
+                        e.preventDefault();
```
jsconan: Same as above. This instruction is already invoked by default by the shortcut helper. At least you should use the `prevent` option to be sure the event is prevented.

views/js/runner/plugins/navigation/previous.js
```diff
@@ -87,9 +131,14 @@ define([
                 })
                 .on('disablenav', function(){
                     self.disable();
+                })
+                .on('nav-previous', function(data){
+                    var previousItemWarning = data && typeof(data[0]) !== 'undefined' ? data[0] : false;
```
jsconan: This is weird, why use an array instead of a simple boolean argument ?

views/js/test/runner/plugins/navigation/next/test.html
```diff
@@ -0,0 +1,24 @@
...
+    <title>Test Runner Plugin - next button - warning helper</title>
+    <link rel="stylesheet" type="text/css" href="/tao/views/js/lib/qunit/qunit.css">
+    <script type="text/javascript" src="/tao/views/js/lib/require.js"></script>
+    <script type="text/javascript" src="/tao/views/js/lib/qunit/qunit.js"></script>
+    <script type="text/javascript" src="/tao/views/js/lib/qunit/qunit-parameterize.js"></script>
+    <script type="text/javascript" src="/tao/views/js/lib/blanket/blanket.min.js" data-cover-only="/taoQtiTest/views/js/runner/plugins/navigation/nextWarningHelper.js"></script>
+    <script  type="text/javascript">
+        QUnit.config.autostart = false;
+        require(['/tao/ClientConfig/config'], function(){
+            require(['taoQtiTest/test/runner/plugins/navigation/next/test'], function(){
```
jsconan: If it is related to the helper, should not it be named the same way ? The name of the test refer to the next plugin itself... (`plugins/navigation/next/test`)

It should rather be names: `helper/nextWarningHelper/test`

PR #631, opened 2016-11-14
Feature/tao 3271 time handling

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3271

Requires: https://github.com/oat-sa/extension-tao-test/pull/152

Add implementation for the Extra Time handling. By default no extra time is added. This is the role of the proctor (when the proper option is set).

The Timing API has also been updated to get rid of silly exceptions.


models/classes/runner/QtiRunnerService.php
```diff
@@ -1200,6 +1201,60 @@ protected function onTimeout(RunnerServiceContext $context, AssessmentTestSessio
...
+    protected function buildTimeConstraints(RunnerServiceContext $context) {
+        $constraints = array();
+        
+        /* @var TestSession $session */
+        $session = $context->getTestSession();
+
+        foreach ($session->getRegularTimeConstraints() as $tc) {
+            $timeRemaining = $tc->getMaximumRemainingTime();
+            
+            // Only consider time constraints in force.
```
jsconan: Oops, copy/paste of existing code

PR #633, opened 2016-11-16
Fetch timer values from storage during initialization of timer

hutnikau:


krampstudio:
So if I understand well you're main problem is the event `plugin-addtimer.timer` is triggered too late in some situations. 
But you cannot initialize the timer before your are sure to read the value, if the db is busy you can have the wrong value taken.

Instead you can  either : 
 -  expose the values through an event (checkout how we retrieve the duration https://github.com/oat-sa/extension-tao-testqti/blob/develop/views/js/runner/plugins/controls/duration/duration.js#L120 ) 
 - change the event sequence (depends on the event sequence but you can use before with promises to ensure the before handler completes), for example it could make sense to add the timers `before('itemload'` and ensure the value is read using a promise.
 - read directly in the store (less recommended because it introduce hard coupling)

hutnikau:
@krampstudio the branch has been updated. Please review it again.

views/js/runner/plugins/controls/timer/timer.js
```diff
@@ -208,37 +208,42 @@ define([
              * Update the timers.
              * It will remove, let, add or update the current timers based on the current context.
              */
-            var updateTimers = function updateTimers(checkStorage){
+            var updateTimers = function updateTimers(checkStorage) {
+                return new Promise(function (resolve) {
+                    _.forEach(timerTypes, function(type){
```
krampstudio: A Promise is resolved only once. Here everything will be considered as resolved at the 1st round of the loop. 
Instead do something like  this : 

```js
 var updateTimers = function updateTimers(checkStorage) {
     //add all promises in an array
     var updates = _.map(timerTypes, function(type) {
         return new Promise(function(resolve) {
             var timerConfig = getTimerConfig(type);
             if (currentTimers[type]) {
                 if (!timerConfig) {
                     removeTimer(type);
                 } else if (currentTimers[type].id() !== timerConfig.id) {
                     removeTimer(type);
                     addTimer(type, timerConfig);
                 }
                 return resolve();
             } else if (timerConfig) {
                 if (checkStorage) {
                     //check for the last value in the storage
                     self.storage.getItem(timerConfig.id).then(function(savedTime) {
                         if (_.isNumber(savedTime) && savedTime >= 0) {
                             timerConfig.remaining = savedTime;
                         }
                         addTimer(type, timerConfig);
                         resolve();
                     }).catch(function() {
                         //add the timer even if the storage doesn't work
                         addTimer(type, timerConfig);
                         resolve();
                     });
                 } else {
                     addTimer(type, timerConfig);
                     return resolve();
                 }
             }
             //haven't check all the cases but resolve MUST be called
            return resolve();
         });
     });

     //this create a promise that resolve once all promises are resolved (in parrallel)
     return Promise.all(updates);
 };
```
(it's untested but you have the big picture)

views/js/runner/plugins/controls/timer/timer.js
```diff
@@ -337,9 +342,12 @@ define([
 
                     //change plugin state
                     testRunner
-                        .on('loaditem resumeitem', function(){
-                            //check for new timers
-                            updateTimers(true);
+                        .before('renderitem resumeitem', function() {
+                            return new Promise(function(resolve) {
```
krampstudio: this structure is not necessary, returning the promise is sufficient : 
```js
.before('renderitem resumeitem', function() {
    return updateTimers(true);
})
```

PR #634, opened 2016-11-16
Feature/tao 3310 answer masking

krampstudio:
https://oat-sa.atlassian.net/browse/TAO-3310

The plugin let's you hide areas of the item. 
You can play with the mask component in the mask's test or inside a test.
I haven't added CSS transitions for the opacity because it was causing too much reflows and freezing the UI.


views/js/runner/plugins/tools/areaMasking/mask.js
```diff
@@ -0,0 +1,212 @@
...
+             */
+            resize : function resize(width, height) {
+                if( this.is('rendered') && !this.is('disabled')) {
+                    this.setSize(
+                        width  > this.config.minWidth  ? width  : this.config.minWidth,
+                        height > this.config.minHeight ? height : this.config.minHeight
+                    );
+
+                    /**
+                     * @event maskComponent#move the component has been resized
```
jsconan: isn't it event `resize` instead of move ?

views/js/runner/plugins/tools/areaMasking/areaMasking.js
```diff
@@ -0,0 +1,157 @@
...
+         */
+        destroy : function destroy(){
+            this.$button.remove();
+        },
+
+        /**
+         * Enable the button
+         */
+        enable : function enable(){
+            this.setState('disabled', false);
```
krampstudio: Yes, I've fixed it. But we have a consistency issue with test plugins and component... one use the `disabled` state when the other uses the `enabled` state

krampstudio:
Everything's fixed

PR #635, opened 2016-11-17
Feature/tao 3313 text highlighter

no-chris:
depends on https://github.com/oat-sa/tao-core/pull/1067


views/js/runner/provider/qti.js
```diff
@@ -656,7 +656,7 @@ define([
         unloadItem : function unloadItem(){
             var self = this;
 
-            self.trigger('disablenav disabletools');
+            self.trigger('beforeunloaditem disablenav disabletools');
```
ssipasseuth: hmmm, should you not trigger the event unloaditem then use eventifier.before('unloaditem') instead ?
https://github.com/oat-sa/tao-core/blob/master/views/js/core/eventifier.js#L450-L457

no-chris: That was my first plan but actually, due to the test runner architecture, the event is triggered once the action has been delegated and performed by the provider. So, using before would still launch the callback once the action is performed...

ssipasseuth: Ok

PR #651, opened 2016-11-24
Feature/movable component

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3377

Requires: https://github.com/oat-sa/tao-core/pull/1076

Refactor the `mask` component in order to use a common component, as the basis will be the same for the magnifier tool.

### DIFF
--- a/views/js/runner/plugins/tools/areaMasking/areaMasking.js
+++ b/views/js/runner/plugins/tools/areaMasking/areaMasking.js
@@ -97,7 +97,6 @@ define([
                 .on('unloaditem', function (){
                     //remove all masks
                     _.invoke(self.masks, 'destroy');
-                    self.masks = [];
                 })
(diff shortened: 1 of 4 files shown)


views/js/runner/plugins/tools/areaMasking/areaMasking.js
```diff
@@ -97,6 +97,7 @@ define([
                 .on('unloaditem', function (){
                     //remove all masks
                     _.invoke(self.masks, 'destroy');
+                    self.masks = [];
```
krampstudio: The reference is removed from the array inside the destroy handler : https://github.com/oat-sa/extension-tao-testqti/pull/651/commits/38bcf43de60a1542e85ae61b94773d258b14ccf8#diff-baf76f6a005822fdf326f8434dffb6d1R78

PR #659, opened 2016-12-01
Feature/tao 3377 magnifier

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3377

Requires:
- https://github.com/oat-sa/tao-core/pull/1080
- https://github.com/oat-sa/tao-core/pull/1087
- https://github.com/oat-sa/tao-core/pull/1089

Implementation of the magnification glass tool.

It works by cloning the content and applying a scale factor on it. A MutationObserver is used to feedback the changes made inside the runner UI.


krampstudio:
There's small glitches when you play a lot with a lot of elements, but I think you're aware of them and they will be addressed if needed ( click doesn't work everywhere, select2 does not open, my CPU got crazy after some amount of time).

jsconan:
Yes there is some performances issues and select2 is not working. There is also an issue with the scroll if you open the magnifier after a scroll.

I have some clues to improve it, but this needs more time

PR #669, opened 2016-12-06
Fix/magnifier unit test

jsconan:
Fix the "move" unit test of the magnifier plugin.

Seems related to a WebKit issue: the transform CSS rule is not read correctly when provided as an inline style.

The patch is a little bit dirty, but I did not find a better way, other than just disable the test.

### DIFF
--- a/views/js/test/runner/plugins/tools/magnifier/magnifier/test.js
+++ b/views/js/test/runner/plugins/tools/magnifier/magnifier/test.js
@@ -18,8 +18,9 @@
 
 define([
     'jquery',
+    'lodash',
     'taoQtiTest/runner/plugins/tools/magnifier/magnifierPanel'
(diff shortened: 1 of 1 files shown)


krampstudio:
Just asking (don't change anything) did you try to defer the `getComputedStyle` or may be you know the style is already applied because calling  `getComputedStyle` before the reflows occurs will give you the previous values, or may be you're in the middle of a transition ?

jsconan:
Yes, I already tried to defer the call to `getComputedStyle`, with no success, event with large delay.
Maybe is it due to the hidden state of the element: as the method should return computed value Chrome doesn't want to compute hidden fragment. Or something like that.

I don't remember which post, but also one of the stackoverflow results I found talked about strange Chrome behaviour with inline style.

PR #670, opened 2016-12-07
Add test id to the report

hutnikau:


actions/class.RestQtiTests.php
```diff
@@ -90,13 +98,31 @@ public function getStatus()
...
+     */
+    protected function getTaskData($taskId)
+    {
+        $data = $this->traitGetTaskData($taskId);
+        $task = $this->getTask($taskId);
+        $report = \common_report_Report::jsonUnserialize($task->getReport());
+        $plainReport = $this->getPlainReport($report);
+
+        //the third report is report of import test
+        if (isset($plainReport[2]) && isset($plainReport[2]->getData()['rdfsResource'])) {
```
hutnikau: I know that this is not the most elegant solution and not really reliable, but for now i did not found better solution to work with nested reports.

PR #673, opened 2016-12-09
Fix/tao 3350 tab key issue

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3350

Requires: https://github.com/oat-sa/tao-core/pull/1098

Fix the shortcuts issue happening with the modal dialogs:
- the TAB key should not focus outside of the dialog anymore
- when cancelling a dialog, the remaining dialogs should stay
- the dialogs now are controlled using these shortcuts (by default): `X` to cancel, `Enter` to validate

### DIFF
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -391,7 +391,7 @@ return array(
         ],
         'dialog' => [
             'accept' => 'Enter',
-            'reject' => 'X'
+            'reject' => 'Esc'
         ]
     ],
 );
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -388,6 +388,10 @@ return array(
         ],
         'previous' => [
             'trigger' => 'K'
+        ],
+        'dialog' => [
+            'accept' => 'Enter',
+            'reject' => 'X'
         ]
     ],
 );
--- a/views/js/runner/plugins/content/dialog/dialog.js
+++ b/views/js/runner/plugins/content/dialog/dialog.js
@@ -27,35 +27,117 @@ define([
     'i18n',
     'taoTests/runner/plugin',
     'ui/dialog/alert',
-    'ui/dialog/confirm'
-], function ($, _, __, pluginFactory, dialogAlert, dialogConfirm){
+    'ui/dialog/confirm',
+    'util/shortcut/registry',
+    'util/shortcut',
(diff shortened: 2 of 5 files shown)


config/default/testRunner.conf.php
```diff
@@ -388,6 +388,10 @@
         ],
         'previous' => [
             'trigger' => 'K'
+        ],
+        'dialog' => [
+            'accept' => 'Enter',
+            'reject' => 'X'
```
krampstudio: Why don't we use the `esc` key ?

views/js/runner/plugins/content/dialog/dialog.js
```diff
@@ -27,35 +27,117 @@ define([
...
-                stack.push({
+                var handle = {
                     context: namespace,
                     dialog: dialog
-                });
+                };
+
+                // prevents all registered shortcuts to be triggered
+                // and brings back the dialog shortcuts
+                shortcut.disable();
```
no-chris: maybe this could be named something like globalShortcuts or, even better, windowShortcuts if those are bound to window ?

config/default/testRunner.conf.php
jsconan: The reason why is the ESC key has an inconsistent behaviour, from the user's point of view.

For instance, when the page is in fullscreen mode, the ESC key is bound to the browser at a low level, to always allow the user to be able to leave this mode. And while the browser is in fullscreen mode, this key is unreachable from JavaScript. So if the user want to cancel the dialog, he/she will first escape the fullscreen, then will have to press ESC again to actually cancel the dialog.

views/js/runner/plugins/content/dialog/dialog.js
jsconan: Yes, you are right, I will change this, as it will improve code readability.

config/default/testRunner.conf.php
krampstudio: May be, in order to be consistent with other modals in tao, and especially since `<esc>` is only unavailable when the fullscreen is requested using the API (not when the user asks it) which is only for one customer.

jsconan: Ok, so let keep the `<esc>` key as a default. I will provide a dedicated config for the customer who need to deal with fullscreen.

PR #674, opened 2016-12-09
Sesstion Stickyness FileUpload support

boajer:
depends on https://github.com/oat-sa/tao-core/pull/1104


boajer:
this PR is duplicated

PR #675, opened 2016-12-12
Features/review panel

dietertaotesting:
Related: https://github.com/oat-sa/extension-tao-mext/pull/1
- Base install was Sprint 39
- Install extension-tao-mext

- see criteria and pdf in https://oat-sa.atlassian.net/browse/TAO-3471
Note that the PDF does not render completely in a browser, download it instead


views/js/runner/plugins/navigation/review/review.js
```diff
@@ -200,17 +200,31 @@ define([
...
+                    if(forcedState === true) {
+                        self.navigator.show();
+                    }
+                    else {
+                        self.navigator.hide();
+                    }
+                    updateButton(self.$toggleButton, getToggleButtonData(self.navigator));
+                }
+
+                else if (self.getState('enabled') !== false) {
```
jsconan: Not sure this is the right way: here you are allowing to bypass the `enable` state of the plugin.

It should be better to take care of the forced state only if the plugin is enable.

What I'd do would be:
```javascript
if (self.getState('enabled') !== false) {
    isHidden = _.isUndefined(forcedState) ? self.navigator.is('hidden') : !forcedState;
    if (isHidden) {
        self.explicitlyHidden = false;
        self.navigator.show();
    } else {
        self.explicitlyHidden = true;
        self.navigator.hide();
    }
}
```

views/js/runner/plugins/navigation/review/review.js
```diff
@@ -268,10 +282,13 @@ define([
             //disabled by default
             this.disable();
 
+
             if (!isEnabled()) {
                 this.hide();
             }
 
+            togglePanel(testConfig.review.defaultOpen);
+
```
jsconan: To enforce my previous comment, here you force the display of the component, even if it has been previously disabled...

scripts/update/Updater.php
```diff
@@ -843,6 +843,6 @@ public function update($initialVersion) {
             $this->setVersion('5.38.2');
         }
 
-        $this->skip('5.38.2', '5.38.4');
+        $this->skip('5.38.2', '5.39.0');
```
jsconan: You should also provide an update script to ensure the default values are there.

PR #677, opened 2016-12-13
Fix/tao 3329 theme switcher keyboard

no-chris:


views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -92,6 +92,17 @@ define([
...
 
             /**
+             * Move the hovered index to the next previous available index
+             */
+            function moveUp() {
+                state.hoveredIndex--;
+                if (state.hoveredIndex < 0) {
+                    state.hoveredIndex = state.availableThemes.length - 1;
+                }
+                highlightMenuEntry();
```
jsconan: Not a change to do, but just a hint.
To make a circular increment/decrement you could also rely on the modulo operator.
The advantage is that prevent to get into weird issue when the iterator is altered for some reason and is out of bound.

For instance:
```javascript
// circular increment
cursor = (cursor + 1) % count;

// circular decrement
cursor = (cursor + count - 1) % count;
```

views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -100,6 +111,31 @@ define([
...
+                }
+            }
+
+            /**
+             * unregister plugin's own shortcuts
+             */
+            function unregisterInnerShortcuts() {
+                shortcut.remove('up.' + self.getName());
+                shortcut.remove('down.' + self.getName());
+                shortcut.remove('select.' + self.getName());
```
jsconan: Not a change to do, but you could also use the same loop you did in  `registerInnerShortcuts()`.

PR #679, opened 2016-12-14
Feature/tao 3492 compiler index

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3492

Requires:
- https://github.com/oat-sa/tao-core/pull/1111
- https://github.com/oat-sa/extension-tao-item/pull/253
- https://github.com/oat-sa/extension-tao-itemqti/pull/863

By default the review panel display the label of each items. This feature add the ability to display the  QTI title instead. To do so an index is built at compile time to gather the attributes of each items of the test. The TestMap builder relies on this index to get the data.

An option set the use of the title (enabled by default): `test-taker-review-use-title`, in `config/taoQtiTest/testRunner.conf.php`.

A sample test is attached to the JIRA ticket.


jsconan:
Changes have been made

PR #687, opened 2016-12-20
Fixed button text

dietertaotesting:
@see https://oat-sa.atlassian.net/browse/TAO-3579


no-chris:
This might introduce a regression on an italian instance, could you please check if this is the case ?

PR #693, opened 2016-12-22
Feature/tao 3535 max masking areas

krampstudio:


views/js/runner/plugins/tools/areaMasking/areaMasking.js
```diff
@@ -48,6 +52,8 @@ define([
 
             var testRunner = this.getTestRunner();
             var $container = testRunner.getAreaBroker().getContentArea().parent();
+            var testConfig = testRunner.getTestData().config || { plugins : {} };
+            var config     = _.defaults(testConfig.plugins[this.getName()] || {}, defaultConfig);
```
hutnikau: Basic plugin config (first argument) will be overwritten here. Maybe it would be better not to change it?
```js
var config = _.defaults({}, testConfig.plugins[this.getName()] || {}, defaultConfig);
```

krampstudio:
updated.

PR #695, opened 2016-12-29
Fix statechange events on session retrieval

jbout:


jbout:
Workaround for rollback https://github.com/oat-sa/qti-sdk/releases/tag/0.10.24

jbout:
Not proud of this solution, but only way to make it work without changing SDK

PR #699, opened 2017-01-05
Feature/tao 3545 score authoring

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3545

Add authoring for the score processing.

Several modes are currently supported:
- `none`: no outcome processing.
- `custom`: a custom outcome processing is defined, it cannot be authored and the existing rules are just kept intact.
- `total`: an outcome and the related processing rules are added to compute a global score, set into the variable `SCORE_TOTAL`.
- `category`: an outcome is created for each categories (the special ones are discarded: `x-tao-*`), and the related processing rules are added to compute score per category basis. The result is stored in each `SCORE_xxx` outcome, where `xxx` is the category name.
- `cut`: a cut score must be provided, and an outcome is created for each categories (the special ones are discarded: `x-tao-*`). For each categories processing rules are added to compute score per category basis and compare it to the cut score. The boolean result is stored in each `PASS_xxx` outcome, where `xxx` is the category name.

To allow to manage the processing rules at client side, some helpers have been added, with very basic support. Not all processing rules are supported yet.

### DIFF
--- a/views/js/controller/creator/helpers/baseType.js
+++ b/views/js/controller/creator/helpers/baseType.js
@@ -1,0 +1,333 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA ;
+ */
+/**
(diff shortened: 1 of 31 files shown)


jsconan:
@krampstudio is the final PR assignee.
@bugalot please do the functional review, as this feature is heavily related to QTI

views/js/controller/creator/helpers/baseType.js
```diff
@@ -0,0 +1,333 @@
...
+ * in the specification itself. The use of an enumeration to define the set of baseTypes
+ * used in the runtime model, as opposed to the use of classes with similar names, is
+ * designed to help distinguish between these two distinct levels of modelling.
+ *
+ * @author Jérôme Bogaerts <jerome@taotesting.com>
+ * @author Jean-Sébastien Conan <jean-sebastien@taotesting.com>
+ */
+define([
+    'lodash'
+], function (_) {
```
jsconan: The reason why is because I only found numerical values in the JSON provided by the server.
For the purpose of this feature, a list of simple constants could do the job, but I integrated the full helper to provide more support. If you think this is overkill I can replace it.

For the cardinalities, I did not see it, just wondering the numbers was the allowed occurrences numbers. So I will replace the magic numbers by constants.

krampstudio:
Ok for me. The validation could be improved but it's ok for now.

jsconan:
The "Cut Score" mode is only related to categories. So if no categories has been set, the score mode is not defined.

bugalot:
I will merge after some extra checks.

PR #704, opened 2017-01-10
Fix/tao 3545 total cut score

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3545

Change the way to author the score processing. Instead of having several exclusive modes to compute category score or total score, the UI now offers to choose between Total or Cut score, and then allows to add the categories score.


bugalot:
Excellent Job !

PR #722, opened 2017-02-03
Check whether tool is enabled or not

hutnikau:


views/js/runner/plugins/tools/answerElimination/eliminator.js
```diff
@@ -113,6 +124,7 @@ define([
 
             //update plugin state based on changes
             testRunner
+                .on('loaditem', togglePlugin)
```
krampstudio: You should also check `isEnabled` inside the `renderitem` handler (it does a `show` if a choice interaction is there)

PR #723, opened 2017-02-03
Feature/tao 3725 item precaching

krampstudio:
- Implement a proxy provider loader to select dynamically the provider based on the configuration
 - Create a cache proxy provider that loads the next item just after the current item rendering
 - Preload assets (only images and css right now) using a DOM approach (css are cached using `<link relf="prefetch" />`, this feature is difficult to feature detect, so I've disabled the link to ensure it has no effect on the current item)
 - moved the assetManager into a proper factory so we can share it
 - Item's data are stored into a proper cache
 - The ability to browser the next item needs to be activated by a flag (see companion PR)

A few tests are missing (`cache/proxy` and `cache/assetLoader`) but I'll add them in a separate PR (after the sprint)


views/js/test/runner/proxy/cache/itemStore/test.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8">
+        <title>Test Runner - qtiServiceProxy</title>
```
jsconan: Wrong name here

PR #728, opened 2017-02-06
Fix/tao 3257 fix svg scaling for select point interaction

eacraft13:
This prevents the zoom plugin from setting margins on an item thus removing an item's responsiveness.


krampstudio:
Requires https://github.com/oat-sa/extension-tao-itemqti/pull/883  (@eacraft13 we always describe the relationship between PRs)

krampstudio:
@jsconan what the impact of removing the zoom _centering_

jsconan:
@krampstudio Well, the impact is about the user experience. When zooming, the point of view will stick to the top left corner instead of staying focused on the center. The correct behaviour should be to keep the point of view on a natural position, aka the center.

However, when this plugin has been ported from the old version of the test runner, the behaviour was not well implemented: the centering only works when zooming out, as the item becomes smaller, but stick to the top left when zooming in. On the previous implementation, the item stayed centered in both zoom directions.

So, for the current implementation, removing the item centering brings more consistency, unless we restore the correct behaviour.

krampstudio:
@ssipasseuth please take over, I'll be in holidays :palm_tree:

manifest.php
```diff
@@ -34,7 +34,7 @@
     'label' => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
-    'version' => '6.3.0',
+    'version' => '6.4.0',
```
ssipasseuth: this is supposed to be a bug fix, so would it not be 6.3.1 instead ?

PR #738, opened 2017-02-08
Feature/remove finish test call

hutnikau:


krampstudio:
Why do you remove the call to finish ?

hutnikau:
Because sometimes we have problems with this call.
One of cases is:
1. test taker presses next on the last item to finish test
2. lose focus (or close tab/connection lose) and leave the test 
3. delivery execution is in pause state but test session is closed.

There are a few another cases. I don't see the reason to have this call if we can finish delivery execution inside move/skip/exit.

Finally it just faster.

jbout:
The reason these calls are currently separate, is that in the past a delivery could contain multiple tests. As we no longer support this use case, we could rethink these interfaces

hutnikau:
The problem is that we close test session and finish delivery execution in two different requests. If one of requests is failed or not executed then delivery execution will be in inconsistent state.

krampstudio:
Ok on the principle, but in that situation you should remove completely the finish action from the controller and unimplemenet it from the provider (method and even).
@jsconan could you check the impacts please.

krampstudio:
@jsconan You'll take over, I'll be in holidays :palm_tree:

jsconan:
What is the status of this PR?
Have you done the changes requested by @krampstudio, but did not push them?

krampstudio:
I'm closing this PR, but I'll use your branch as a basis for the changes.

krampstudio:
_Please do not delete the branch_

PR #743, opened 2017-02-10
Feature/tao 3749 per tool collapse order

no-chris:
https://oat-sa.atlassian.net/browse/TAO-3749


views/js/runner/plugins/content/responsiveness/collapser.js
```diff
@@ -37,12 +37,29 @@ define([
...
+     * ex:
+     * var collapseOrder = [
+     *      '[data-control="highlight-clear"],[data-control="highlight-trigger"]',  // those will collapse first...
+     *      '[data-control="hide-review"]',                                         // this one second...
+     *      '[data-control="set-item-flag"]',                                       // third...
+     *      ...                                                                     // ...
+     * ];
+     * @type {String[]}
+     */
+    var collapseOrder = [];
```
jsconan: Why are you using an event instead of a config entry? Is there any chance the list may change at runtime? In the companion PR the list is set roughly at init. So that seems to be artificial, isn't it? Or is there something I didn't see/understand?

PR #748, opened 2017-02-15
Fix/tao 3732 tools name and order

no-chris:


scripts/update/Updater.php
```diff
@@ -1051,5 +1049,7 @@ public function update($initialVersion) {
 
             $this->setVersion('6.3.1');
         }
+
+        $this->skip('6.1.1', '6.3.2');
```
ssipasseuth: becareful : 6.**1**.1

PR #756, opened 2017-02-22
added option to remove eliminations

dietertaotesting:
Depencies between:
https://github.com/oat-sa/extension-tao-testqti/pull/756
https://github.com/oat-sa/extension-tao-itemqti/pull/894
https://github.com/oat-sa/tao-core/pull/1199
https://github.com/oat-sa/extension-tao-act/pull/493, instructions are provided there

### DIFF
--- a/views/js/runner/plugins/tools/answerElimination/eliminator.js
+++ b/views/js/runner/plugins/tools/answerElimination/eliminator.js
@@ -167,7 +167,7 @@ define([
                                 $choices.each(function() {
                                     var input = this.querySelector('.real-label input');
                                     if(!!this.dataset.wasEliminated) {
-                                        delete this.dataset.wasEliminated;
+                                        this.dataset.wasEliminated = null;
                                         this.classList.add('eliminated');
                                         input.setAttribute('disabled', 'disabled');
                                         input.checked = false;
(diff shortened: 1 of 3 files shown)


views/js/runner/plugins/tools/answerElimination/eliminator.js
```diff
@@ -144,12 +158,31 @@ define([
...
                 .on(actionPrefix + 'toggle', function () {
+                    var $choices = self.$choiceInteractions.find('.qti-choice');
                     if (isEnabled()) {
                         self.$choiceInteractions.toggleClass('eliminable');
                         if (self.$choiceInteractions.hasClass('eliminable')) {
                             self.button.turnOn();
+                            if(config.restoreEliminationsOnOpen) {
+                                $choices.each(function() {
+                                    var input = this.querySelector('.real-label input');
+                                    if(!!this.dataset.wasEliminated) {
```
jsconan: The double negation is not needed here, as it is already in condition.

views/js/runner/plugins/tools/answerElimination/eliminator.js
```diff
@@ -144,12 +158,31 @@ define([
...
+                                        delete this.dataset.wasEliminated;
```
jsconan: Is the `delete` really needed? This can lead to memory leak. Prefer to use `null`.

PR #757, opened 2017-02-22
Test Runner Timer target

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3788

Make the class QtiRunnerConfig a service.

Load the timer target from the config when needed.


hutnikau:
Good job, thanks!

PR #759, opened 2017-02-23
Fix/tao 3835 redirect session end

krampstudio:
Refactor the controller to have a better error handling. If any 403 we redirect the user to the `exitUrl`. This will help in the situation of the original bug so users will be redirected to the correct url.

Note [on exit](https://github.com/oat-sa/extension-tao-testqti/compare/develop...fix/TAO-3835_redirect-session-end?expand=1#diff-11034f9ca39a14fd1bc50ce0ddd29a83R120)  my first intention was to provide an exit reason, but due to the redirection mechanism, the message is kept even after login again...

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -34,7 +34,7 @@ return array(
     'label' => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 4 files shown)


krampstudio:
No, I forgot it... It's done now

PR #768, opened 2017-03-09
Feature/tao 3963 item responses

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3963

Improve the `currentItem` helper by exposing internal functions in order to be able to manage item responses:
- `getDeclarations`: gets the responses declarations of the current item
- `getResponseDeclaration`: gets a particular response declaration from the current item
- `toResponse`: build a response descriptor based on the provided value (need baseType and cardinality from the declaration)
- `isQtiValueNull`: checks if the value is considered as NULL
- `isQuestionAnswered`: tells whether a particular question has been answered
- `isAnswered`: tells whether the current item has been answered

Also add unit tests.

### DIFF
--- a/views/js/runner/helpers/currentItem.js
+++ b/views/js/runner/helpers/currentItem.js
@@ -21,9 +21,8 @@
  * @author Jean-Sébastien Conan <jean-sebastien.conan@vesperiagroup.com>
  */
 define([
-    'jquery',
     'lodash'
-], function ($, _) {
+], function (_) {
     'use strict';
 
     /**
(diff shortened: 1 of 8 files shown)


views/js/runner/helpers/currentItem.js
```diff
@@ -21,104 +21,153 @@
...
-        var response = {};
+        /**
+         * Gets a response declaration by the identifier of the response
+         * @param {Object} runner - testRunner instance
+         * @param {String} identifier - The identifier of the response
+         * @returns {Object|null}
+         */
+        getResponseDeclaration: function getResponseDeclaration(runner, identifier) {
+            var found = null;
+            _.forEach(currentItemHelper.getDeclarations(runner), function (declaration) {
```
no-chris: why not use 
`this.getDeclarations(runner)` ?
(in other methods as well)

PR #771, opened 2017-03-13
Use AbstractTaskAction class

hutnikau:


manifest.php
```diff
@@ -34,12 +34,13 @@
     'label' => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
-    'version' => '6.12.0',
+    'version' => '6.12.1',
```
chrisphpgarcia: If you add dependecy, Is that better to upgrade minor version instead of patch.

PR #779, opened 2017-03-21
Feature/tao 3942 pause message

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3942

In the new TestRunner displays a more accurate pause/terminate message.

A service is implemented to provide messages related to the state of the session. This service should be overridden to provide custom messages.

When a state change event occurs, a message is built and stored into the extended state storage of the test session. To do so an event queue is handled to maintain a list of pending messages. Each time an action is requested, the response will also contain the pending messages, as the pause/terminate messages to be displayed. The client side will process these messages, and display them.

These messages are the same as those managed by the communicator channel.

The service `TestSessionService` has been moved, for a big part, from the taoProctoring extension, as this service is heavily related to QTI test: it provides a `TestSession` from a session ID. This service is needed to get access to the session from a delivery state event.

The `ExtendedStateService` was not registered, so this PR also fix that. It also replace obsolete CONFIG_ID by SERVICE_ID constants. Hence the number of changed files.


models/classes/runner/communicator/QtiCommunicationService.php
```diff
@@ -145,10 +151,23 @@ protected function getChannel($channelName, $channelType)
     {
         $channels = $this->getOption(self::OPTION_CHANNELS);
         $channel = new $channels[$channelType][$channelName];
+        $this->propagateDependencies($channel);
```
antoinerobin: please use 
```php
$this->getServiceManager()->propagate($instance);
```
instead

models/classes/runner/QtiRunnerService.php
```diff
@@ -1317,7 +1325,7 @@ public function switchClientStoreId(RunnerServiceContext $context, $receivedStor
             $session = $context->getTestSession();
             $sessionId = $session->getSessionId();
 
-            $stateService = new ExtendedStateService();
+            $stateService = $this->getServiceLocator()->get(ExtendedStateService::SERVICE_ID);
```
antoinerobin: can you use getServiceManager() instead of getServiceLocator() to keep consistency

PR #795, opened 2017-04-03
Changes templates to provide button roles, labels and labelled-by att…

zorkow:
…ributes. Adds new ids to item descriptions.
Replicates titles as labels to fix a problem with NVDA/Firefox.


jsconan (commented):
Looks good. But you should add spaces between HTML attributes.

krampstudio:
That's a great addition. We will review the PR.

PR #807, opened 2017-04-14
Fix/tao 4134 menu container is undefined error

eacraft13:
This is for https://oat-sa.atlassian.net/browse/TAO-4134.

I could not reproduce bug, so I checked for possible places for reference errors. The solution adds default properties to avoid possible reference errors.


views/js/runner/ui/toolbox/menu.js
```diff
@@ -43,6 +43,19 @@ define([
...
+        $menuButton: $(),
+        $menuContainer: $(),
+        $menuContent: $(),
+        $menuItems: $(),
+        $menuStateIcon: $(),
+        config: {},
+        hoverIndex: null,
+        id: null,
+        menuItems: []
+    };
```
eacraft13: Good catch on `_defaults` placement. I'll move that to the factory. And remove `config` as well.

views/js/runner/ui/toolbox/menu.js
```diff
@@ -432,6 +445,9 @@ define([
                 this.$menuButton.off('.menuNavigation');
             });
 
+        // Apply default properties to the menuComponent
+        _.defaults(menuComponent, _defaults);
```
eacraft13: I'm going to opt to use default properties. I prefer this to adding conditionals around all menu properties.

PR #811, opened 2017-04-21
Fix/tao 4264 fix inconsitent messages

krampstudio:
Only send consistent events


models/classes/runner/communicator/TestStateChannel.php
```diff
@@ -88,18 +88,21 @@ public function process(QtiRunnerServiceContext $context, array $data = [])
 
                     if ($state == AssessmentTestSessionState::CLOSED) {
                         $type = 'close';
+
                     } else if ($state == AssessmentTestSessionState::SUSPENDED) {
                         $type = 'pause';
-                    } else {
+
+                    }
+                    if(is_null($type)){
```
jsconan: You should rather use `isset()`, as it will check for `null` as well as the existence of the variable. Otherwise that may trigger a PHP notice.

PR #817, opened 2017-04-27
Feature/tao 4220 answer masking tool

no-chris:
https://oat-sa.atlassian.net/browse/TAO-4220

companion  PR: https://github.com/oat-sa/extension-tao-act/pull/533


views/js/runner/plugins/tools/answerMasking/answerMasking.js
```diff
@@ -0,0 +1,179 @@
...
+                    this.setState('masked', false);
+
+                    return this;
+                },
+
+                /**
+                 * Cover the whole choice with the mask
+                 * @returns {component}
+                 */
+                mask: function hide() {
```
jsconan: the function name is not the same

views/js/runner/plugins/tools/answerMasking/plugin.js
```diff
@@ -0,0 +1,229 @@
...
+        /**
+         * Initialize the plugin (called during runner's init)
+         */
+        init: function init() {
+            var self = this;
+
+            var testRunner = this.getTestRunner(),
+                testData = testRunner.getTestData() || {},
+                testConfig = testData.config || {},
+                pluginConfig = _.defaults((testConfig.plugins || {})['answer-masking'] || {}, defaultConfig),
```
jsconan: Even if this naming is correct, it should be better to use the exact plugin name for more consistency.

views/js/runner/plugins/tools/answerMasking/plugin.js
```diff
@@ -0,0 +1,229 @@
...
+                var testContext = testRunner.getTestContext(),
+                    itemId = testContext.itemIdentifier;
+
+                if (answerMasking.getState('enabled')) {
+                    itemStates[itemId] = answerMasking.getMasksState();
+                }
+
+                answerMasking.disable();
+                self.button.turnOff();
+                testRunner.trigger('plugin-end.answer-masking');
```
jsconan: Same remark as for the config, you should use the plugin name for more consistency.
You could also use the `plugin.trigger()` method that will automatically prefix the event.

views/js/runner/plugins/tools/answerMasking/plugin.js
```diff
@@ -0,0 +1,229 @@
...
+            function enableMasking() {
+                var testContext = testRunner.getTestContext(),
+                    itemId = testContext.itemIdentifier;
+
+                answerMasking.enable();
+                if (pluginConfig.restoreStateOnToggle) {
+                    answerMasking.setMasksState(itemStates[itemId]);
+                }
+                self.button.turnOn();
+                testRunner.trigger('plugin-start.answer-masking');
```
jsconan: Same remark as for the config, you should use the plugin name for more consistency.
You could also use the `plugin.trigger()` method that will automatically prefix the event.

PR #820, opened 2017-05-03
Fix/tao 4224 qti sm dependency move

krampstudio:
Filter categories which aren't QTI/XML compliant (the validation was previously on taoItems)

Related to https://github.com/oat-sa/extension-tao-item/pull/266


krampstudio:
The dependency is transitive, (taoQtiItem -> taoItems) and adding the dependency will require to have always a match between both extensions, so it's not a good idea if not required

PR #829, opened 2017-05-11
Update TestCategoryPresetProvider.php

markmolenaar:
Here are my proposed textual changes; there's still some inconsistency left in the overall platform (e.g. use of user/candidate/test-taker), but we'll need to address that at a later stage.

### DIFF
--- a/models/classes/TestCategoryPresetProvider.php
+++ b/models/classes/TestCategoryPresetProvider.php
@@ -46,30 +46,30 @@ class TestCategoryPresetProvider extends ConfigurableService
                 'presets'    => [
                     TestCategoryPreset::fromArray([
                         'id'            => 'reviewScreen',
-                        'label'         => __('Review Screen'),
+                        'label'         => __('Enable Review Screen'),
                         'qtiCategory'   => 'x-tao-option-reviewScreen',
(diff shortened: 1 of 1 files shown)


models/classes/TestCategoryPresetProvider.php
```diff
@@ -46,130 +46,130 @@ protected function getDefaultPresets() {
...
+                        'description'   => __('Display a warning before the test-taker skips the section. Requires the Section Skipping option.'),
                         'order'         => 300
                     ]),
                     TestCategoryPreset::fromArray([
                         'id'            => 'noExitTimedSectionWarning',
-                        'label'         => __('noExitTimedSectionWarning'),
+                        'label'         => __('Hide Section Timeout Warning'),
                         'qtiCategory'   => 'x-tao-option-noExitTimedSectionWarning',
-                        'description'   => __('Disable the warning automatically displayed upon exiting a timed section'),
+                        'description'   => __('Hide the warning automatically displayed when a section times out.'),
```
no-chris: actually this is not when the section timeout. It's a weird "three-letter-client" feature where exiting a timed section, finished or not; would automatically display a warning. This category allows to bypass this behavior.

PR #831, opened 2017-05-11
set positions and dimensions to even integers

dietertaotesting:
https://oat-sa.atlassian.net/browse/TAO-4305

As far as my OS/browsers are concerned this is much better then before but it is not perfect. When I zoom the window it mostly looks good but on certain sizes some white lines appear.
I tried a huge number of tricks to improve this situation and would now say this is a limitation in the browser rendering. Eventually fell back to original solution and discarded everything else.


views/js/runner/plugins/tools/lineReader/compoundMask.js
```diff
@@ -115,20 +126,22 @@ define([
         function createMask(maskConfig) {
             var mask,
                 maskAPI = {
-                    place: maskConfig.place,
-                    placeOverlay: maskConfig.placeOverlay,
+                place: maskConfig.place,
```
no-chris: why the indent change ?

views/js/runner/plugins/tools/lineReader/compoundMask.js
```diff
@@ -435,13 +449,24 @@ define([
...
+
+            // round dimensions and positions up to the next even integer
+            _.forOwn(dimensions, function(value, key) {
+                dimensions[key] = ceilToEvenInt(value);
+            });
+
+            _.forOwn(position, function(value, key) {
+                position[key] = ceilToEvenInt(value);
+            });
+
```
no-chris: I'm not sure about this. While the fix does seems to improve the zoom, I have the feeling it has some side effects like this one, that I can reproduce fairly easily now, while I cannot on develop (see the inner top right corner)

screen shot 2017-05-12 at 10 31 02

PR #840, opened 2017-05-19
Tao 4363 score report

hutnikau:


models/classes/runner/rubric/QtiRunnerRubric.php
```diff
@@ -51,6 +51,9 @@ public function getRubrics(RunnerServiceContext $context, $itemRef = null)
         if(!is_null($itemRef)){
             try {
                 $routeItem = $session->getRoute()->getRouteItemsByAssessmentItemRef($itemRef);
+                if ($routeItem) {
```
krampstudio: May be you should check if `$routeItem` is an intanceof `qtism\runtime\tests\RouteItemCollection`  (or a parent class)

actions/class.Runner.php
```diff
@@ -453,7 +453,7 @@ protected function getItemDataResponse(RunnerServiceContext $context, $itemRef,
     {
         $itemData = $this->runnerService->getItemData($context, $itemRef);
         $baseUrl  = $this->runnerService->getItemPublicUrl($context, $itemRef);
-        $rubrics  = $this->runnerService->getRubrics($context);
+        $rubrics  = $this->runnerService->getRubrics($context, $itemData['data']['identifier']);
```
krampstudio: The identifier might not be the itemRef.

models/classes/runner/rubric/QtiRunnerRubric.php
hutnikau: `$routeItem` is always instance of `RouteItemCollection` 
https://github.com/oat-sa/qti-sdk/blob/legacy/qtism/runtime/tests/Route.php#L1007

PR #841, opened 2017-05-19
Feature/tao 4398 progressbar

dietertaotesting:
dep: https://github.com/oat-sa/extension-tao-nbce/pull/59

Instructions are also in this PR


views/js/runner/plugins/controls/progressbar/progressbar.js
```diff
@@ -32,6 +32,11 @@ define([
     'use strict';
 
     /**
+     * Display 'item x of y' when true, 'item x' otherwise
+     */
+    var progressShowTotal;
```
jsconan: Why is this variable global to the module? You should rather make it local to the plugin instance, like `progressIndicator` and `progressScope`.

views/js/runner/plugins/controls/progressbar/progressbar.js
```diff
@@ -162,8 +170,10 @@ define([
             var progressIndicator = config.type || 'percentage';
             var progressScope = config.scope || 'test';
 
+            progressShowTotal = !!config.showTotal;
```
jsconan: As the config becomes more complex, you should probably transform the two last parameters of the `progressUpdater()` function into a config set:
```javascript
...
function progressUpdater(testContext, testMap, progressConfig) {
    ...
}
...
var progressConfig = {
    indicator: config.type || 'percentage',
    scope: config.scope || 'test',
    showTotal: !!config.showTotal
};
...
var progressData = progressUpdater(testRunner.getTestContext(), testRunner.getTestMap(), progressConfig);
...
```

scripts/update/Updater.php
```diff
@@ -1234,6 +1234,6 @@ public function update($initialVersion) {
             $this->setVersion('9.3.3');
         }
 
-        $this->skip('9.3.3', '9.4.1');
+        $this->skip('9.3.3', '9.5.0');
```
jsconan: You must also provide an update script to set the default value of the new option. Otherwise the existing behavior of the progress bar will be lost.

PR #849, opened 2017-05-30
Fix/tao 4393 hide disabled plugins presets

no-chris:
https://oat-sa.atlassian.net/browse/TAO-4393

The new architecture allow each extension to register a CategoryPresetProvider, instead of having a single overridable service. This will allow for more flexibility as more extensions might need to register presets, for example TaoTestRunnerPlugins (this is not the case yet, however).

Also, the main CategoryPresetProvider service now checks the availability of plugins before returning any preset.

Companion PR: https://github.com/oat-sa/extension-tao-act/pull/566


krampstudio:
I'd like a backend dev to perform a review.

models/classes/TestCategoryPresetRegistry.php
```diff
@@ -0,0 +1,55 @@
...
+ * @author Christophe Noël <christophe@taotesting.com>
+ */
+
+namespace oat\taoQtiTest\models;
+
+use common_ext_Extension;
+use common_ext_ExtensionsManager;
+use oat\oatbox\AbstractRegistry;
+
+class TestCategoryPresetRegistry extends abstractRegistry
```
krampstudio: should be `AbstractRegistry`

test/TestCategoryPresetProviderTest.php
```diff
@@ -71,42 +74,163 @@ public function testSort()
...
 
-
-        $presetProvider = new TestCategoryPresetProvider([], $allCategories);
-        $sortedCategories = $presetProvider->getPresets();
-
-        $this->assertCount(4, $sortedCategories, 'sortedCategories have the right size');
+        $this->assertCount(4, $sortedPresetGroups, 'sortedPresetGroups have the right number of preset groups');
         $previousOrder = 0;
-        forEach($sortedCategories as $group) {
+        forEach($sortedPresetGroups as $group) {
```
krampstudio: `forEach` should be by convention `foreach`

manifest.php
```diff
@@ -28,7 +28,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '9.8.0',
+    'version'     => '9.9.0',
```
no-chris: Yes, only the underlying implementation has changed.

no-chris:
@bugalot is going to review it

bugalot:
@krampstudio @no-chris 

OK for me in terms of backend code !

PR #858, opened 2017-06-07
remove tests subdirectory

chrisphpgarcia:


bugalot (changes requested):
This Pull Request does work with a wide range of tests.

The following process cannot be achieved with this PR.

1. Download test https://github.com/oat-sa/extension-tao-testqti/blob/master/test/samples/archives/QTI%202.1/rubrics/RubricBlocksStylesheetsImages.zip.

2. Import the test as QTI 2.1.

3. Run the test in old/new test runner, everything fine.

4. Export the test as QTI 2.1

5. Re-import the test as QTI 2.1 -> It fails to import.

It worked perfectly in previous versions.

Thanks for making the appropriate changes.

bugalot (changes requested):
I still have the same behaviour after updating your changes. No way to export/import the package I reported in my last review.

bugalot (changes requested):
Unfortunately, by reproducing the scenario described in the first PR review, the test cannot be compiled as a delivery. It breaks because paths from test.xml to assets (stylesheets and images) are broken after export.

PR #860, opened 2017-06-08
Feature/tao 4440 toggle timers

krampstudio:
Add a clock icon left to the timers to hide the timers. It hides only the stopwtach to prevent blinking effects. 
The preference is also saved in the local db to survive to page reload, resumes, etc.


views/js/runner/plugins/controls/timer/timer.js
```diff
@@ -323,6 +328,38 @@ define([
...
+                    }
+                }
+            }
+
+            /**
+             * Hide the toggler without timers,
+             * display it otherwise
+             */
+            function toggleToggler() {
+                var $toggler = self.$element.find('.timer-toggler');
```
jsconan: Why don't you use `this.$toggler` instead ?

krampstudio:
updated

PR #868, opened 2017-06-12
Fix version from previous fix so as not to get stuck on 9.12.0

eacraft13:
@chilly86, I thought you should see the fix to versioning from [your hotfix](https://github.com/oat-sa/extension-tao-testqti/pull/867). Basically, just incrementing the setVersion() function could place someone's version permanently into 9.12.0. So, you must add a skip() for these types of version increments.


chilly86:
Fix under the pull request: https://github.com/oat-sa/extension-tao-testqti/pull/869

PR #875, opened 2017-06-20
To the testRunner configuration added new parameter 'next-section-pau…

zagovorichev:
…sed' which pause delivery execution when section has changed


zagovorichev:
backend implementation (I don't no if it is needed)

PR #876, opened 2017-06-20
Fix/tao 4501 collapse on resize

no-chris:


views/js/runner/plugins/content/responsiveness/collapser.js
```diff
@@ -75,28 +91,115 @@ define([
...
+                                $elements.each(function() {
+                                    extraWidth += getExtraWidth($(this));
+                                });
+                                return {
+                                    $elements: $elements,
+                                    extraWidth: extraWidth
+                                };
+                            }
+                            return false;
+                        }).filter(function (collapsible) {
```
jsconan: Instead of using such a filter, you could also rely on the `_.compact()` helper, that does the same: https://lodash.com/docs/2.4.2#compact

views/js/runner/plugins/content/responsiveness/collapser.js
```diff
@@ -108,29 +211,78 @@ define([
...
+                    shouldStop = false;
+
+                collapsiblesCopy.reverse();
+
+                collapsiblesCopy.forEach(function (toExpand) {
+                    if (toExpand.$elements.hasClass(collapseCls) && ! shouldStop) {
+                        if (expandPossible(toExpand.extraWidth)) {
+                            toExpand.$elements.removeClass(collapseCls);
+                        } else {
+                            shouldStop = true;
```
jsconan: Rather using such complex structure, you could also use `_.forEachRight()`, that loops from the end to the beginning, and then return `false` to break the loop when needed.  That way no need to clone the array and to manage a stop flag...

https://lodash.com/docs/2.4.2#forEachRight

views/js/test/runner/mocks/areaBrokerMock.js
```diff
@@ -54,31 +54,42 @@ define([
...
-        var mapping = {};
         var $areaBrokerDom = $('<div />').attr('id', 'area-broker-mock-' + (mockId++)).addClass('test-runner');
 
         config = config || {};
 
-        if (!config.areas) {
-            config.areas = defaultAreas;
+        if (!config.mapping) {
+            config.mapping = {};
+            if (!config.areas) {
```
jsconan: So, you must provide a mapping to be able to get areas from config. Is it wanted?

PR #880, opened 2017-06-23
Feature/tao 4465 runner optimizations

krampstudio:
### Structural changes
 - Removal of the _submitItem_ step in most of the cases.
 - We check if the item will have _modalFeedbacks_ upfront and use the _submitItem_ step only if the item has feedback.
 - unified HTTP parameters for move, skip, timeout, exit and _submitItem_.
 - the _allowSkipping_ is now implemented client side (using the `preventSkipping` plugin) with fallback validation server side (the _submitItem_ request was used also for that).
 - rubric blocks where previously in the _getItem_ response, they are now part of the `testContext` (so we don't open the test session in the _getItem_ query).
 - move the function `isLeavingSection` to a new helper `taoQtiTest/runner/helpers/navigation`.

### Modifications in the HTTP requests

- `/taoQtiTest/Runner/submitItem` isn't used any more expect in the specific case of modal feedback.
- `/taoQtiTest/Runner/move`, `/taoQtiTest/Runner/timeout`, `/taoQtiTest/Runner/submitItem` and `/taoQtiTest/Runner/exitTest` have now all the following additional parameters :
    -  `itemState`  : A JSON string that contains the item state to save
    -  `itemResponses` : A JSON string that contains the item responses
    -  `itemIdentifier` : the identifier of the item
    -  `itemDuration` : the duration of the item is seconds (a float)
    -  `consumedExtraTime` :   extra time is seconds (a float)
- `/taoQtiTest/Runner/getItem` has now an additional parameter
    -  `itemIdentifier` : the identifier of the item
- `/taoQtiTest/Runner/getItem` has lost  the `rubricBlock` from the response
- All requests that produce a `testContext` (`init`, `move`, etc.), have now the following additional entries in the response :
    - `testContext.preventEmptyResponses=<true|false>`
    - `testContext.hasFeedbacks=<true|false>`
    - `testContext.rubrics=<undefined|string|string[]>`


views/js/runner/plugins/navigation/preventSkipping.js
```diff
@@ -0,0 +1,94 @@
...
+                .after('move skip', function(){
+                    toggle();
+                });
+        },
+
+        /**
+         * Called during the runner's destroy phase
+         */
+        destroy : function destroy (){
+        },
```
jsconan: Is this really needed ?

views/js/runner/provider/qti.js
```diff
@@ -190,118 +208,103 @@ define([
...
+
+                                    if(results.displayFeedbacks === true && results.feedbacks) {
+                                        self.itemRunner.renderFeedbacks(results.feedbacks, results.itemSession, function(queue){
+                                            self.trigger('modalFeedbacks', queue, resolve);
+                                        });
+                                        return;
+                                    }
+                                }
+                                return resolve();
+                            });
```
jsconan: There is no `catch` on this promise. What happen if the server refuse the `submit` ?

views/js/runner/provider/qti.js
```diff
@@ -190,118 +208,103 @@ define([
...
-                                        self.trigger('modalFeedbacks', queue, resolve);
-                                    });
+                                if (results.itemSession) {
+                                    context.itemAnswered = results.itemSession.itemAnswered;
 
+                                    if(results.displayFeedbacks === true && results.feedbacks) {
+                                        self.itemRunner.renderFeedbacks(results.feedbacks, results.itemSession, function(queue){
+                                            self.trigger('modalFeedbacks', queue, function(){
+                                                computeNext(action, _.omit(params, 'itemState', 'itemResponse'));
+                                            });
```
jsconan: Maybe I am wrong, but is there chance the feedbacks be displayed twice? There is two places where the feedbacks are handled. I suppose they are sent by the server only after a response submit?

views/js/runner/provider/qti.js
```diff
@@ -430,20 +390,12 @@ define([
...
-                            });
-                        })
-                        .catch(function(err){
-                            self.trigger('error', err);
-                        });
+                    self.trigger('alert.timeout', __('Time limit reached, this part of the test has ended.'), function() {
+                        computeNext('timeout', _.merge(getItemResults(), {
+                            scope: scope,
+                            ref: ref
+                        }));
```
jsconan: I see a potential issue here, as you display the timeout alert before requesting the server.
What happen if the test taker takes a while before acknowledging the message? Or just close the browser, and open the test on another one?
Maybe should it be better to request the server, and wait for the test taker before displaying the result of the request, should it not?

actions/class.Runner.php
```diff
@@ -662,7 +688,18 @@ public function timeout()
...
+
+            if (!$this->runnerService->isTerminated($serviceContext)) {
+                $this->endItemTimer($serviceContext);
+                $this->saveItemState($serviceContext);
+            }
+
+            $this->runnerService->check($serviceContext);
+            $this->serviceContext->init();
+
+            $this->saveItemResponses($serviceContext);
```
krampstudio: `saveItemResponses` is called in different manner, but I'll factorize the context init/check

views/js/runner/provider/qti.js
krampstudio: I just forgot to remove one of them

views/js/runner/provider/qti.js
krampstudio: just pushed a fixed were the timeout request is pushed before displaying the message

krampstudio:
I'm wondering if such HTTP changes aren't considered as breaking changes in semver

jsconan:
Yes, I had the same wondering.

PR #881, opened 2017-06-27
Update README.md

thomasgarrard:
Updated documentation


zagovorichev (commented):
Generally it will be paused on entering in the paused item. 
What do I mean: this property for items and from the backend side only item would be flagged as paused by proctor. But when used tool from the taoTestRunnerPlugin (frontend) it become pause for the section.

PR #884, opened 2017-06-30
Fix spelling mistake of initialization

eacraft13:
A simple spelling mistake fix. I also need to update the translation files before merging.

So first, are po files populated programmatically? Or do I need to update the msgid for each file containing the misspelling manually?

Second, once po files are updated, how do I generate the .lang.php and .js file?

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -28,7 +28,7 @@ return array(
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '9.18.1',
+    'version'     => '9.18.2',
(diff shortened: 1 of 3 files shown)


krampstudio:
This process is automated : a script parse the source code and look for the pattern `__('STRING')`. Then the PO is submitted to our [translation platform](https://translate.taotesting.com/projects/tao32/). Another script build the `.lang.php` and `.js` based on the translated files. 
_(this process might have change, you can check this with @llecaque  or @chazotte)_

krampstudio:
https://hub.taocloud.org/articles/documentation-for-core-components/internationalization-i18n

PR #885, opened 2017-06-30
Fix/show eliminator on fullscreen

eacraft13:


dietertaotesting:
I assume tyou will still need to increase the testqti dependency in the manifest of the customer's extension

PR #886, opened 2017-07-01
enable all missing tools

krampstudio:
Enable by default in the core product : 
 - the highlighter
 - the area masking
 - the answer eliminator
 - the magnifier


krampstudio:
Please verify if this can have an impact on customer deployments

PR #887, opened 2017-07-03
Feature/tao 4571 precaching x items

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4571

Requires: https://github.com/oat-sa/extension-tao-testqti/pull/880

Update the item pre-caching feature to be able to load the next X items.
By default the amount of items to cache is fixed to 3.
According to the acceptance criteria of the story, the previous and next X items are preloaded.

The cache proxy is relying on the test map to list the neighbor items. However the current implementation of the Test Runner, to access items, is using item keys that contains both the URI of the item and the URI of its storage. The test map is not providing item keys in this format. So it is impossible to request items this way.

We need a consistent way to identify items in any components of the runner. To do so we could use the item identifier, which is unique inside a test definition. But we also need to map those identifiers to the actual item definition. This the job of a backend developer. Meanwhile, in order to not be stuck by this missing behavior, I added a temporary map on which the server is relying to access the items. The Test Runner should now only use the item identifier to manipulate items.
All places where this workaround has been applied were tagged with the ID of the JIRA ticket: TAO-4605

Note: the customer specific plugins may still use the old identifier format. We will need to address that case by case.


views/js/runner/proxy/cache/proxy.js
```diff
@@ -63,27 +63,37 @@ define([
...
+            this.getItemFromStore = false;
 
             //keep a ref to the promise of the loadNextItem in case the call is not done when moving
             this.loadNextPromise = Promise.resolve();
 
+            //preload at least this number of items
+            this.cacheAmount = 1;
+
+            //maintain a map of the items chain: for each item, a link to its neighbors
+            this.itemChain = {};
```
krampstudio: This is the point that should be reviewed. The next step will be to work with a dynamic testMap (adaptive session) and the item chain can be changed. I see that's more convenient to maintain such reference, but it will be easier for the future test cases to just read the testMap. So we will have a unique source of data. Something like : 

```js
/**
 * @param {Object} testMap
 * @param {Number} itemPosition - (could be also the item id)
 * @param {String} direction - previous/next/both
 * @param {Number} [size=3] - will be 2xsize if direction is both
 * @returns {Object[]} the collections of items
 */ 
function navigationHelper.getSiblings(testMap, itemPosition, direction, size)
```
No ?

jsconan: You are right! I could also make that from the map builder, server side, aside the list of jumps. That way there should be only one place where we compute the map.

jsconan:
I just made the changes: 
- new helpers: `navigationHelper.getSiblingItems`, `navigationHelper.getNextItem`, `navigationHelper.getPreviousItem` 
- the cache proxy is now using the new helpers

actions/class.Runner.php
```diff
@@ -380,62 +393,34 @@ public function getItem()
...
+                //load item data
+                $response = $this->getItemData($itemIdentifier);
+            }
+
+            if (is_array($response)) {
+                $response['success'] = true;
             }
 
+            //start the timer
+            $this->runnerService->startTimer($serviceContext);
```
krampstudio: You shouldn't start the timer if you retrieve items for caching. In the other actions there's a `start` parameter.

views/js/runner/provider/qti.js
```diff
@@ -549,10 +548,7 @@ define([
          * @returns {Promise} that calls in parallel the state and the item data
          */
         loadItem : function loadItem(itemRef){
-            var self = this;
-            var context = self.getTestContext();
-
-            return self.getProxy().getItem(itemRef, { itemIdentifier : context.itemIdentifier })
+            return this.getProxy().getItem(itemRef)
```
krampstudio: Could you rename it `itemIdentifier`

views/js/runner/proxy/cache/proxy.js
```diff
@@ -63,27 +64,37 @@ define([
...
+
             /**
              * Update the item state in the store
              * @param {String} uri - the item identifier
-             * @param {Objetc} state - the state of the item
+             * @param {Object} state - the state of the item
              * @returns {Boolean}
              */
-            this.updateState = function updateState(uri, state){
+            this.updateState = function updateState(uri, state) {
```
krampstudio: Could you please rename `uri` by `id` or `itemIdentifier` everywhere to avoid confusion.

views/js/runner/proxy/cache/proxy.js
```diff
@@ -145,27 +163,38 @@ define([
...
+                            list.push(sibling.id);
+                        }
+                        return list;
+                    }, []);
 
-                                        if(response.baseUrl && response.itemData && response.itemData.assets){
-                                            assetLoader(response.baseUrl, response.itemData.assets);
-                                        }
+                    //don't run a request if not needed
+                    if (missing.length) {
```
krampstudio: If you don't check if it's the last item anymore, please remove the occurrences of `isLastItem`

jsconan:
I made the requested changes and I also fixed a caching issue that prevented to load item from cache under particular circumstances.

PR #891, opened 2017-07-11
Fix/timeline range

jsconan:
This patch intends to fix common `TimeLine` issues, especially when there is missing `TimePoint`.
I encountered some of them when working on the items pre-caching story.

The principle of this patch is to try to create the missing `TimePoint` when we detect unopened or unclosed time ranges.

This fix should also correct the mixed/nested ranges:
- `Start1`, `Start2`, `End2`, `End1` => `Start1`, **`End1a`**, `Start2`, `End2`, **`Start1b`**, `End1`
- `Start1`, `Start2`, `End1`, `End2` => `Start1`, **`End1a`**, `Start2`, **`End2b`**, **`Start1c`**, `End1`, **`Start2d`**, `End2`


krampstudio:
How does this affects the duration computation ? 
If I take your first example : 
- `Start1`, `Start2`, `End2`, `End1` => `Start1`, **`End1a`**, `Start2`, `End2`, **`Start1b`**, `End1`
Do we have  this equality ?
- `End1 - Start1 == (End1a - Start1) + (End1 - Start1b)`

jsconan:
For my first example, it should be equal to `End1` - `Start1`. The added `TimePoint` are put with the same timestamp as the wrong `TimePoint`.

So taking the first example terms, in which there are 2 new `TimePoint`:
`Start1`, `Start2`, `End2`, `End1` => `Start1`, **`End1a`**, `Start2`, `End2`, **`Start1b`**, `End1`

we have:
`End1a` = `Start2`
`Start1b` = `End2`

and that produces:
`End1` - `Start1` = `(End1a - Start1) + (End2 - Start2) + (End1 - Start1b)`

krampstudio:
Ok, great so there's no way it affects the duration

krampstudio:
Do you suggest a way to produce wrong TimePoint for testing purposes ?

jsconan:
The issue appeared when I was trying to modify the TestRunner requests to replace the `itemRef` by `itemIdentifier`. And in that case I encountered a lot a failed requests, that lead to broken test execution. So I think the issue could now appear only in case of strong server issue, or maybe with network issues.
I also encountered the issue before I patch the `TimePoint::compare()`: as sometimes the `TimePoint` were created in a very small time frame (within the same millisecond), so the `TimePoint` were badly sorted and that introduced a situation similar to the example 2.
What I can do may be to add dedicated unit tests

krampstudio:
Yes a unit test could be a good idea

PR #892, opened 2017-07-11
Feature/compiled assessment item ref hrefs

bugalot:
This PR provides a way to store in a reliable and compiled way the assessmentItemRef->href values of a test delivery. They are retrievable by assessmentItemRef->identifier without loading the test definition.

@krampstudio I just put a "@bertrand" notice in the code where to put your fallback for backward compatibility.


krampstudio:
related to TAO-4350

bugalot:
Understood @krampstudio I agree! I'll change that !

bugalot:
@krampstudio Done ! You can go ! Thanks for your review.

krampstudio:
@jsconan will be in charge of this part

jsconan:
@krampstudio the fallback is now in place. It is still relying on the test state storage, but is only activated if the delivery has not been compiled using the updated compiler (i.e. the item href index has not been built in the package).
The reason why we didn't choose to force an update of deliveries, to get rid of this fallback, is because the script, that inject the index in the existing deliveries, is consuming to many resources. And also this kind of updater is dangerous to be automated.

models/classes/runner/map/QtiRunnerMap.php
```diff
@@ -40,28 +40,63 @@ class QtiRunnerMap extends ConfigurableService implements RunnerMap
...
+    }
+
+    /**
+     * Checks if the AssessmentItemRef Identifier is in the index.
+     *
+     * @param QtiRunnerServiceContext $context
+     * @param string $itemIdentifier
+     * @return boolean
+     */
+    public function hasItemHrefIndexFile(QtiRunnerServiceContext $context, $itemIdentifier)
```
krampstudio: should be protected

scripts/tools/RecompileHrefIndexes.php
```diff
@@ -0,0 +1,79 @@
...
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
+ */
+ 
+namespace oat\taoQtiTest\scripts\tools;
+
+use oat\oatbox\extension\AbstractAction;
+use \common_report_Report as Report;
+use qtism\data\storage\php\PhpDocument;
+
+class RecompileHrefIndexes extends AbstractAction
```
krampstudio: Could you please provide a description of what this script is used to

models/classes/class.QtiTestCompiler.php
```diff
@@ -777,12 +782,44 @@ private function getServiceLocator()
...
+     * Build Href Index Path
+     * 
+     * Builds the Href Index Path from given $identifier.
+     * 
+     * @param string $identifier
+     * @return string
+     */
+    static public function buildHrefIndexPath($identifier)
+    {
+        return TAOQTITEST_COMPILED_HREF_INDEX_FILE_PREFIX . md5($identifier) . TAOQTITEST_COMPILED_HREF_INDEX_FILE_EXTENSION;
```
krampstudio: I just realize it, but dies it means we will have one file per itemRef ?

krampstudio: Because if we have a lot of items in the test, when building the test map, we will loop over all the items and open all the files and read all the files ?

krampstudio: I'm wondering if building the map won't be faster with the fallback than with the file reads, when the number of items is higher than a threshold.

jsconan: Yes, that was also my though. I did not ask @bugalot about that. Maybe should we gather all items into one file instead, as it has been done for the `test-index.json`.

jsconan: I just asked @bugalot about the index. This solution is the faster, as we request only one item when needed, instead of having a big index to load each time. When building the map we are not accessing every single item index, but only one at the beginning to check if we need to build the fallback.

krampstudio: ok, I see. So the rule is if we have already the TestSession opened, we should get the itemhref from it, and we get it from the index file only if we don't have a session or an AssessmentItem opened. We should be careful, this index is used only accurately. Maybe we can hide the implementation details.

krampstudio:
I'm also wondering if the TestMap is the correct class to manage the retrieval if the `itemHref`,

krampstudio:
Last but not least, we should test it on a prod like deployment. Since the index files will be stored on S3 (with a local cache) we are going to create a lot of s3 requests. I'll approach the Devops team. 

I prefer to be very careful, because it will be difficult to roll-back or use another approacj since it involves the compilation.

PR #897, opened 2017-07-16
Feature/tao 4733 lti extended time

tikhanovichA:


models/classes/runner/QtiRunnerService.php
```diff
@@ -1281,13 +1283,19 @@ protected function buildTimeConstraints(RunnerServiceContext $context) {
...
-    protected function buildExtraTime(RunnerServiceContext $context) {
+    protected function buildExtraTime(RunnerServiceContext $context)
+    {
         /* @var TestSession $session */
         $session = $context->getTestSession();
         $timer = $session->getTimer();
+        $sessionMaxTime = null;
+        $sessionTimeLimits = $session->getCurrentTestPart()->getTimeLimits();
+        if ($sessionTimeLimits) {
+            $sessionMaxTime = $sessionTimeLimits->getMaxTime()->getSeconds(true);
```
jsconan: Caution, `$sessionTimeLimits->getMaxTime()` could be null.

jsconan:
The `getMaxTime()` issue is fixed, but please avoid to mix "merge" and "feature" commits into one, that leads to messy history. 
You should have one commit dedicated to merge, if needed, and one another for the feature itself.
Moreover the merge commit should be the place to fix the version if conflict occurs. Having several commits to manage merge conflicts is a bad habit that could introduce issues. Of course it can happen that sometimes we do a mess, but this should be avoided.

PR #900, opened 2017-07-17
Feature/tao 4567 client state cache

krampstudio:
> You can start reviewing the PR. The integration with the server is missing (through the `sync` channel`). Some unit tests are also missing and will come soon.

 - new methods to the navigation helper : `isFirst`, `isLast`, `isLeavingTestPart`
 - `navigator/navigator` let's you navigate offline (ie. build a context offline)
 - the `connectivity` plugin takes now it's source from the proxy (I was thinking to add the `disconnect` and `reconnect` event directly in the test runner, but the behaviour can be different in other implementations)
 - `actionStore` stores actions performed offline
 - `proxy/cache/proxy` uses a "request then offline" strategy, it tries to perform the action offline, register it and sync it next time online.
 
Requires : 
 - https://github.com/oat-sa/tao-core/pull/1414
 - https://github.com/oat-sa/extension-tao-test/pull/184


views/js/runner/proxy/cache/proxy.js
```diff
@@ -133,6 +128,178 @@ define([
...
+                //perform the request, but fallback on offline if the request itself fails
+                var runRequetThenOffline = function runRequetThenOffline(){
+                    return self.request(url, actionParams).then(function(result){
+                        //if the request fails, we should be offline
+                        if(self.isOffline()){
+                            return self.offlineAction(action, actionParams);
+                        }
+                        return result;
+                    });
+                };
```
jsconan: The promise rejection is not catched, so the connectivity error should only appear on resolve?

views/js/runner/proxy/cache/proxy.js
```diff
@@ -167,62 +333,56 @@ define([
...
-                                        });
-                                    }
-
-                                    resolve();
-                                })
-                                .catch(resolve);
-                        }, loadNextDelay);
-                    } else {
-                        resolve();
+            var loadNextItem = function loadNextItem(){
```
jsconan: So now we never wait for the `getNextItemData` to be achieved. Won't it produce any issue, especially with security token?

views/js/runner/proxy/cache/proxy.js
```diff
@@ -133,6 +128,178 @@ define([
...
+                return this.queue.serie(function(){
+                    return self.actiontStore.flush().then(function(data){
+                        if(data && data.length){
+                            return self.send('sync', data);
+                        }
+                    })
+                    .catch(function(err){
+                        self.trigger('error', err);
+                    });
+                }, 'sync');
```
jsconan: What is the last parameter for? It seems to be unused in the `promiseQueue`.

views/js/runner/proxy/cache/proxy.js
krampstudio: no, since it uses the `qtiServiceProxy.request` which push the request in the `promiseQueue`, so having a second wrapping promise was unecessary

krampstudio:
I saw I've pushed the commit from the backend PR

PR #905, opened 2017-07-19
Feature/tao 4587 offline test runner

siwane:
Requires https://github.com/oat-sa/extension-tao-act/pull/627


models/classes/runner/RunnerParamParserTrait.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    protected $serviceContext;
+
+    /**
+     * Check if the parameter associated to $name exists
+     *
+     * @param $name
+     * @return mixed
+     */
+    abstract public function hasRequestParameter($name);
```
jsconan: Is it the right place for that? I mean isn't it better to implement an interface instead of using a trait to set abstracts?

models/classes/runner/QtiRunnerService.php
```diff
@@ -1343,18 +1343,19 @@ public function storeTraceVariable(RunnerServiceContext $context, $itemUri, $var
 
     /**
      * Starts the timer for the current item in the TestSession
+     *
      * @param RunnerServiceContext $context
+     * @param $timestamp
```
krampstudio: Could you please explain what means the timestamp is used for, and when it's null

models/classes/runner/QtiRunnerService.php
```diff
@@ -1370,19 +1371,20 @@ public function startTimer(RunnerServiceContext $context)
 
     /**
      * Ends the timer for the current item in the TestSession
+     *
      * @param RunnerServiceContext $context
      * @param float $duration The client side duration to adjust the timer
      * @param float $consumedExtraTime The extra time consumed by the client
+     * @param $timestamp
```
krampstudio: same here why do we have a timestamp when we have the duration

scripts/update/Updater.php
```diff
@@ -1322,5 +1329,46 @@ public function update($initialVersion) {
         }
 
         $this->skip('10.1.0', '10.3.0');
+
+        if ($this->isVersion('10.3.0')) {
+
+            // Install the synchronisation service
+            if ($this->getServiceManager()->has(SynchronisationService::SERVICE_ID)) {
```
jsconan: Why do you duplicate the behavior of the install scripts instead of directly call them? It should be the same, and should also simplify the code

PR #906, opened 2017-07-19
Feature/tao 4728 ibn metadata guardian

bugalot:
**DEPENDS ON https://github.com/oat-sa/extension-tao-itemqti/pull/984**


bugalot:
Depends on https://github.com/oat-sa/extension-tao-itemqti/pull/984

PR #908, opened 2017-07-20
Refactor the API that manage the service context

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4605

Refactor the API that manage the service context by removing the contraption-like code.


krampstudio:
Merged into develop see 8b2259b276b06d5458d76bc69d05d2506d84f1f2

PR #913, opened 2017-07-25
Added zoom tool to Test taker tools

dietertaotesting:
Don't move the case in Jira since there is still a backport to 3.2 to do


jsconan:
Code is OK. However despite the tool is available after an update, it is not after an install. You should probably activate the plugin in the installer

PR #915, opened 2017-07-26
Feature/tao 4793 booklet multiple deliveries

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4793

Make the rubric block helper a service in order to centralize the feature.

This helper is now used in both version of the test runner, and will also be used in the booklet (from taoQtiPrint).


helpers/class.TestRunnerUtils.php
```diff
@@ -514,36 +515,11 @@ static public function buildAssessmentTestContext(AssessmentTestSession $session
...
-                include($tmpFile);
-                $rubrics[] = ob_get_clean();
-                unlink($tmpFile);
-                rmdir($tmpDir);
-            }
-             
-            $context['rubrics'] = $rubrics;
+            /** @var QtiRunnerRubric $rubricBlockHelper */
+            $rubricBlockHelper = ServiceManager::getServiceManager()->get(QtiRunnerRubric::SERVICE_ID);
+            $context['rubrics'] = $rubricBlockHelper->getRubricBlock($session->getRoute()->current(), $session, $compilationDirs);
```
antoinerobin: `$compilationDirs` here should be an array, the documentation says string and the description "An array containing..."
Which one is true ?
`TestContextBuilder::extendAssessmentTestContext` has also this ambiguous description

helpers/class.TestRunnerUtils.php
```diff
@@ -514,36 +515,11 @@ static public function buildAssessmentTestContext(AssessmentTestSession $session
...
-                ob_start();
-                include($tmpFile);
-                $rubrics[] = ob_get_clean();
-                unlink($tmpFile);
-                rmdir($tmpDir);
-            }
-             
-            $context['rubrics'] = $rubrics;
+            /** @var QtiRunnerRubric $rubricBlockHelper */
+            $rubricBlockHelper = ServiceManager::getServiceManager()->get(QtiRunnerRubric::SERVICE_ID);
```
antoinerobin: could you please factorise this retrieval of ServiceManager so when we will change it we will have only one place to change

jsconan:
Changes made and conflict resolved.

PR #916, opened 2017-07-26
added indicator for connectivity

dietertaotesting:
https://oat-sa.atlassian.net/browse/TAO-4833
Depends on https://github.com/oat-sa/tao-core/pull/1423 for the icons
Related https://github.com/oat-sa/extension-tao-devtools/pull/97
Don't move the case in Jira yet, this needs to be backported


manifest.php
```diff
@@ -29,7 +29,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '10.11.1',
+    'version'     => '10.11.2',
```
krampstudio: You should increase the feature number since you've added a new feature. 
You should also increase the dependency to the tao extension.

views/scss/inc/_connectivity.scss
```diff
@@ -0,0 +1,23 @@
+.connectivity-box {
+    display: none;
+    width: 40px;
+    margin-left: 40px;
+    .icon-connect, .icon-disconnect {
+        display: none;
+        line-height: 1.6;
+        font-size: 2.3rem;
```
krampstudio: you should use the font mixin `@include font-size(23)` to be consistent with the application.

PR #923, opened 2017-08-02
Feature/tao 4840 cat answer required

eacraft13:
Feature - ACT

Supplemental pull requests: 
- https://github.com/oat-sa/extension-tao-act/pull/655

Acceptance Criteria:
- On an adaptive test, such as the PreACT, a response is required on a displayed item within the test delivery platform before the user can navigate (by any means) elsewhere.
- On an adaptive test, a response is allowed to be de-selected (as it is today), but a response is required before the user can navigate (by an means) elsewhere.
- On an adaptive test, a response is allowed to be changed by clicking on a different response than was previously chosen.
- User is notified of the disallowed behavior mentioned above (#1 & #2) when attempted.
- On non-adaptive test, the ability to not have to choose a response behaves as was previously designed.

Notes: 
The solution essentially fixes the qti 'allowSkipping' option to solve this issue. By deselection 'allowSkipping', any test (including a cat test), will require a response before moving to the next item. Further, the qti option 'validateResponses' was added, as it is similar to 'allowSkipping'. You'll see the difference is that 'validateResponses' requires an item to be answered fully, while allowSkipping only requires a partial answer (which seems to correspond with the qti definitions: http://www.imsglobal.org/question/qtiv2p2p1/QTIv2p2p1-ASI-InformationModelv1p0/imsqtiv2p2p1_asi_v1p0_InfoModelv1p0.html#DerivedCharacteristic_ItemSessionControl.Attr_allowSkipping).


views/js/controller/creator/templates/testpart-props.tpl
```diff
@@ -202,11 +202,11 @@
...
-            </div>
-            <div class="col-1 help">
-                <span class="icon-help" data-tooltip="~ .tooltip-content" data-tooltip-theme="info"></span>
-                <div class="tooltip-content">
-                {{__ "The candidate is not allowed to submit wrong responses."}}
+                <div class="col-1 help">
+                    <span class="icon-help" data-tooltip="~ .tooltip-content" data-tooltip-theme="info"></span>
+                    <div class="tooltip-content">
+                    {{__ "The candidate is not allowed to submit invalid responses."}}
+                    </div>
```
krampstudio: Could you please uncomment this part too

views/js/controller/creator/templates/section-props.tpl
```diff
@@ -366,7 +366,7 @@
             <div class="col-1 help">
                 <span class="icon-help" data-tooltip="~ .tooltip-content" data-tooltip-theme="info"></span>
                 <div class="tooltip-content">
-                {{__ "The candidate is not allowed to submit wrong responses."}}
+                {{__ "The candidate is not allowed to submit invalid responses."}}
```
krampstudio: Could you please uncomment this part too

views/js/controller/creator/templates/itemref-props.tpl
```diff
@@ -241,7 +241,7 @@
             </div>
         </div>
 
-{{!-- Property not yet available in delivery
+{{!-- Property not yet available in delivery --}}
```
krampstudio: This comment needs to be removed

views/js/controller/creator/templates/itemref-props.tpl
```diff
@@ -256,11 +256,11 @@
             <div class="col-1 help">
                 <span class="icon-help" data-tooltip="~ .tooltip-content" data-tooltip-theme="info"></span>
                 <div class="tooltip-content">
-                {{__ "The candidate is not allowed to submit wrong responses."}}
+                {{__ "The candidate is not allowed to submit invalid responses."}}
                 </div>
             </div>
         </div>
---}}
+{{!-- --}}
```
krampstudio: This comment needs to be removed

scripts/update/Updater.php
```diff
@@ -1461,5 +1461,14 @@ public function update($initialVersion) {
         }
         
         $this->skip('10.17.0', '10.18.0');
+
+        if ($this->isVersion('10.18.0')) {
+            $extension = \common_ext_ExtensionsManager::singleton()->getExtensionById('taoQtiTest');
+            $config = $extension->getConfig('testRunner');
+            $config['enable-validate-responses'] = false;
+            $extension->setConfig('testRunner', $config);
```
krampstudio: You should also register the 'validateResponses' plugin

eacraft13:
@krampstudio, it should be good to go now.

krampstudio:
FYI the Jira issue number in the branch name, should be in upper case to be linked automatically.

PR #944, opened 2017-08-08
Fix/tao 4891/missing spanish translation

chilly86:


chilly86:
Yes I know also, but the mentioned change with the configs would be a conceptual change and it's a bigger story, not a small bugfix. I create a TT ticket about it to not forget.

PR #954, opened 2017-08-10
compute test map for adaptive section

hutnikau:

PR #960, opened 2017-08-15
Fix/prevent progressbar error for adaptive pkgs

Alroniks:


hutnikau:
already fixed in https://github.com/oat-sa/extension-tao-testqti/pull/954/files#diff-063c7d99fbd5417288e0f1d84339dd2bR127

PR #975, opened 2017-08-22
Fix/tao 4963 end of test connectivity error

eacraft13:
Bugfix - ACT
Jira: https://oat-sa.atlassian.net/browse/TAO-4963

To test, you only need to click 'end test' on an item while disconnected.


views/js/runner/plugins/controls/connectivity/connectivity.js
```diff
@@ -84,7 +84,10 @@ define([
                 }
 
                 //offline navigation error, we pause the test
-                if(err.source === 'navigator' && err.purpose === 'proxy' && err.code === 404){
+                if (
+                    proxy.isOffline() ||
+                    (err.source === 'navigator' && err.purpose === 'proxy' && err.code === 404)
```
dietertaotesting: I don't really see how this could be improved, I'd leave it like that

eacraft13: Ok, thanks for looking @dietertaotesting. I'm going to try to give it one last review today and I'll probably assign it to you tomorrow.

eacraft13:
@dietertaotesting thanks again for looking. I agree, checking if the proxy is offline is a more robust way of handling the error.

eacraft13:
@dietertaotesting, I'm unable to get the reference error ('content' of null) on my branch. Perhaps the js was not loaded (maybe via caching of js from the develop branch)? If you have time, could you walk me through how to produce that error again?

PR #976, opened 2017-08-24
updated connect/disconnect icon

dietertaotesting:
Related:
https://github.com/oat-sa/tao-core/pull/1447 (also dependency)
https://github.com/oat-sa/extension-tao-devtools/pull/99

https://oat-sa.atlassian.net/browse/TAO-4833
This is an update of the icons used to indicate connection/disconnection. So far they have only be used in the context of a delivery so there won't be any surprises. 

https://github.com/search?p=1&q=org%3Aoat-sa+disconnect&type=Code

The original task refers to a customer platform but this is working in all deliveries:
These icons old-icons 
have been replaced with these new-icons


views/scss/inc/_connectivity.scss
```diff
@@ -4,8 +4,9 @@
     margin-left: 40px;
     .icon-connect, .icon-disconnect {
         display: none;
-        line-height: 1.6;
-        @include font-size(23);
+        line-height: 2;
```
eacraft13: I agree that this line-height is the best placement for the icon, but, 1.6 is more in line with the other things such as the title and progress bar.
screen shot 2017-08-27 at 10 55 13 pm

dietertaotesting: I understand your concern but I would not want to change this. Actually the progress bar has been diplaced at some point and now hangs too low. I suggest you take the PR as it is and I file a bug concerning the progress bar. How about it?

PR #977, opened 2017-08-24
move item timer start.

bugalot:
What kind of impact could have this change?

This is really important, because if we find there is no impact, we can significantly improve the performances (CPU consumption) of TAO at delivery time (current gain ratio is currently +- 1.7).

Would you like to help by testing this in timing scenarios, especially with our beloved secret blue customer? Is this "start" GET request parameter really in use? Could you help to crawl the frontend source code to see it is really triggered somewhere?

Thanks so much!

PS: we can discuss the "why" internally.
PS2: As soon as we had an exhaustive tour of this, we might envision to enlarge the people assigned as reviewers.
**PS3: Please find great QTI Test Samples to test this in this PR!**


bugalot:
For the pain, additional samples have been introduced.

PR #980, opened 2017-08-28
Fix/tao 4840 cat answer required

eacraft13:
Bugfix - ACT

Requires
- https://github.com/oat-sa/extension-tao-testqti/pull/991

This fix is to show/hide allowSkipping/validateResponses options from test authors based on whether the feature is enabled.


ssipasseuth:
Hi, I am just wondering whether the product own has approved of this change ? I mean that sounds like a legitimate improvement but there is also the case in which people would like to author the test in TAO and export it to run it somewhere else. In this case I can imagine the complaint that one can no longer turn this attribute to true/false because their intention is to edit the test metadata+attribute not necessarily run it. I hope I am making any sense...

eacraft13:
I had created a think tank issue a while back and commented again not to long ago but haven't heard anything yet. I slacked Mark directly today and hopefully he'll review it soon.

PR #985, opened 2017-08-30
Fix/allow skipping blocks default or null responses

eacraft13:
Via IMS, allowSkipping should reject responses that are 'not null' or default.


antoinerobin:
works fine I'll merge once I've got @bugalot approval

PR #991, opened 2017-08-31
Fix/tao 4840 cat answer required install config

eacraft13:
Bugfix - ACT

This is to fix configuration for previously created tao-4840 feature. The decision was to make all new installs of tao (for any product) enable allowSkipping and validateResponses features. However, updates will not enable this unless specified by that customer. That is, tao update will not set these variables to true, but will instead be updated in a customer's updater if requested.


bugalot:
**If we merge this PR, all our customers will be impacted with their QTI Test Driver behaviour changed. THIS CANNOT HAPPEN!**

PR #1002, opened 2017-09-05
Feature/tao 4942 tao secure delivery

tikhanovichA:


views/js/runner/provider/qti.js
```diff
@@ -405,10 +405,8 @@ define([
                         self.trigger('endsession');
                     }
                 })
-                .on('log', function(data){
-                    self.getProxy().callTestAction('log', {
-                        logData: data
-                    });
+                .on('traceLog', function(traceData){
```
krampstudio: I wouldn't expose this event for the reasons described in https://github.com/oat-sa/extension-tao-test-runner-plugins/pull/21#pullrequestreview-60821957

PR #1011, opened 2017-09-07
pass context but not position to get scope map

hutnikau:


views/js/runner/helpers/map.js
```diff
@@ -189,32 +189,37 @@ define([
         /**
          * Gets the map of a particular scope from a particular position
          * @param {Object} map - The assessment test map
-         * @param {Number} position - The current position
+         * @param {Щиоусе} context - The current session context
          * @param {String} [scope] - The name of the scope. Can be: test, part, section (default: test)
          * @returns {object} The scoped map
          */
-        getScopeMap: function getScopeMap(map, position, scope) {
+        getScopeMap: function getScopeMap(map, context, scope) {
```
krampstudio: I'm not sure to get why you change the API. 
Is the jump table inaccurate ? Or the `itemPosition` in the map ?

krampstudio: I'll suggest you to add a new method `getScopeMapFromContext` instead

PR #1015, opened 2017-09-09
Fix/tao 5078/keep timer notallowed

tikhanovichA:
Added keeping timer when answer is not allowed.

### DIFF
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -503,10 +503,5 @@ return array(
      * @type integer
      */
     'item-cache-size' => 3,
-
-    /**
-     * Keep timer for not allowed answer
-     */
-    'keep-timer-notallowed' => false
-
+    
 );
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -502,6 +502,11 @@ return array(
      * This is required for caching scenarios
      * @type integer
      */
-    'item-cache-size' => 3
+    'item-cache-size' => 3,
+
+    /**
+     * Keep timer for not allowed answer
+     */
+    'keep-timer-notallowed' => false
 
 );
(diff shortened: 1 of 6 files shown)


config/default/testRunner.conf.php
```diff
@@ -502,6 +502,11 @@
      * This is required for caching scenarios
      * @type integer
      */
-    'item-cache-size' => 3
+    'item-cache-size' => 3,
+
+    /**
+     * Keep timer for not allowed answer
+     */
+    'keep-timer-notallowed' => false
```
krampstudio: I don't think you need a flag for that. The behavior should be the same for the whole platform. 
Since the item is displayed in the behind, the timer should continue to countdown

views/js/runner/plugins/controls/timer/timer.js
```diff
@@ -506,21 +506,21 @@ define([
...
-                                    removeTimer(timerTypes.item);
-                                })
-                                .catch(function cancelMove() {
-                                    // Use `defer` to be sure the timer resume will occur after the move event is
-                                    // finished to be handled. Otherwise, the duration plugin will be frozen and
-                                    // the additional time will not be taken into account!
-                                    _.defer(function() {
-                                        testRunner.trigger('enableitem enablenav');
-                                    });
+                            return movePromise;
```
krampstudio: This will broke the actual plugin behvior, please try with only ; 
```
.before('move', function(){
 //no change but
   movePromise.catch(unction cancelMove() {
      ///...
   })
   return movePromise;
})
.on('move', function(){
 doDisable();
removeTimer(timerTypes.item);
})
```

remove

tikhanovichA:
fixed

krampstudio:
Looks better but this need more testing (multiple timers, leaving timed section, etc.)

PR #1017, opened 2017-09-10
Fix/tao 5063 answered items in cat

krampstudio:
The item doesn't come as "asnwered". 
I'm also wondering why the seen items are considered as answered in the test map https://github.com/oat-sa/extension-tao-testqti/commit/e84990a7beb92cf002e7e9b77bb610b4cfed5e96#diff-da05797a06045d7cf1ebd0095656ade4R225

### DIFF
--- a/models/classes/runner/QtiRunnerService.php
+++ b/models/classes/runner/QtiRunnerService.php
@@ -475,7 +475,7 @@ class QtiRunnerService extends ConfigurableService implements RunnerService
                 $response['enableValidateResponses'] = $config->getConfigValue('enableValidateResponses');
 
                 //contextual value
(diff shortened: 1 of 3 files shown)


models/classes/runner/QtiRunnerService.php
```diff
@@ -475,7 +475,7 @@ public function getTestContext(RunnerServiceContext $context)
                 $response['enableValidateResponses'] = $config->getConfigValue('enableValidateResponses');
 
                 //contextual value
-                $response['validateResponses'] = $testOptions['validateResponses'];
+                $response['validateResponses'] = $tesOptions['validateResponses'];
```
bugalot: Hello there :smile: ! Maybe we have a typo? Should'nt it be "$testOptions"?

PR #1022, opened 2017-09-11
Fix/tao 5027 answer masking always mask text

no-chris:


ssipasseuth:
no problem with this PR. tested only in unit test though.
I tried to activate it in an actual tli delivery context but cannot get it working. 
I have ACT extension installed.
In test authoring I checked "answer masking".
And in the LTI consumer, I set the lti options:
```
proctored=false
x_tao_tools={"answerMasking":true}
```
Am I missing something else ?

PR #1028, opened 2017-09-13
Fix/tao 5138 allow skipping closed item

krampstudio:
If the item session is closed or the item in timeout : 
 - do not send state and result on move
 - do not check allowSkipping and valid responses
 - prevent to quickly show the item before the warning 

In order to test the PR, let a section timeout then go back on the section. Items should have the "allowSkipping" false feature to ask for response.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -29,7 +29,7 @@ return array(
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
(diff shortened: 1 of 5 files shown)


bugalot:
@no-chris Any news on this?

krampstudio:
@no-chirs is suffereing from a strange bug that prevent him to test the PR : when a section is timeout and he click "previous" he jumps over the section... I'm not able to reproduce the behavior too.

PR #1030, opened 2017-09-13
Feature/remove timer tags

siwane:
The purpose of this PR is to reduce the amount of data stored in QtiTimeLine by deleting couple of (unnecessary) tags


models/classes/runner/session/TestSession.php
```diff
@@ -150,28 +150,14 @@ public function setTimerTarget($timerTarget)
...
-        $sections = $routeItem->getAssessmentSections();
-        $sections->rewind();
-        $sectionId = key(current($sections));
         $itemRef = $routeItem->getAssessmentItemRef();
         $itemId = $itemRef->getIdentifier();
         $occurrence = $routeItem->getOccurence();
 
-        $tags = [
-            $itemId,
+        return [
```
krampstudio: Dead code after

models/classes/runner/session/TestSession.php
```diff
@@ -150,28 +150,14 @@ public function setTimerTarget($timerTarget)
...
             $itemId . '#' . $occurrence,
-            $sectionId,
-            $testPart->getIdentifier(),
-            $test->getIdentifier(),
-            $itemRef->getHref(),
         ];
 
-        if ($this->isRunning() === true) {
-            $tags[] = $this->getItemAttemptTag($routeItem);
```
krampstudio: I'd keep this one too, otherwise you won't be able to differentiate different attempts o

siwane: I agree and I restore it, however this tag is the most heavy.

jsconan:
Ok, the timers are working well now!

jsconan:
It is also possible to reduce the size of the tags by removing the item id from the occurence and the attempt identifiers. The test id could also be removed. However this will need some small changes in the TimeLine usages.

Here is a sample of what the tags are currently looking: `[item-1, item-1#0, item-1#0-1, assessmentSection-4, testPart-4, Test-1]`.

Once the redundant ID have been removed, this sample could be: `[item-1, #0, #0-1, assessmentSection-4, testPart-4]`

jsconan:
I created a TT story for the tags system improvement: https://oat-sa.atlassian.net/browse/TT-221

PR #1036, opened 2017-09-14
Feature/tao 4941 activate text to speech

eacraft13:
Feature - TAO
Jira: https://oat-sa.atlassian.net/browse/TAO-4941

Requires:
- https://github.com/oat-sa/extension-tao-testqti/pull/1032

Acceptance Criteria:
- TTS is off by default
- TTS needs to be activated on specific test content by use of a specific category (e.g. x-tao-option-tts), set by GUI in the test editor
- TTS can be activated on the LTI launch of a delivery by using a custom parameter (like other ACT accommodations, e.g. tts=true)

Notes:
This is part of two pull requests combined. The other (https://github.com/oat-sa/extension-tao-testqti/pull/1032) is the plugin, and this pr is it's installation. This branch is cut from the other pr branch, not develop.


scripts/install/RegisterTestRunnerPlugins.php
```diff
@@ -350,6 +350,15 @@ class RegisterTestRunnerPlugins extends InstallAction
...
                 'active'      => true,
                 'tags'        => [  ]
+            ], [
+                'id' => 'textToSpeech',
+                'name' => 'Text to Speech',
+                'module' => 'taoQtiTest/runner/plugins/tools/textToSpeech/plugin',
+                'bundle' => 'taoQtiTest/loader/testPlugins.min',
+                'description' => 'Enables text-to-speech via Texthelp',
+                'category' => 'tools',
+                'active' => true,
```
ssipasseuth: should it enabled only on demand and not activated for all client by default ?

scripts/update/Updater.php
```diff
@@ -1590,5 +1590,20 @@ public function update($initialVersion) {
...
+
+        if ($this->isVersion('15.7.6')) {
+            $registry = PluginRegistry::getRegistry();
+            $registry->register(TestPlugin::fromArray([
+                'id' => 'textToSpeech',
+                'name' => 'Text to Speech',
+                'module' => 'taoQtiTest/runner/plugins/tools/textToSpeech/plugin',
+                'description' => 'Enables text-to-speech via Texthelp',
+                'category' => 'tools',
+                'active' => true,
```
ssipasseuth: should it enabled only on demand and not activated for all client by default ?

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -0,0 +1,374 @@
...
+                stopSpeech:              '$rw_stopSpeech',
+                tagSentences:            '$rw_tagSentences',
+                userParameters:          '$rw_userParameters'
+            }
+        };
+        var speed;
+        var volume;
+
+        _.assign(options || {}, {
+            contentArea: $('body'),
```
ssipasseuth: maybe we should throw an error if the contentArea is not defined instead of default to body ?

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -0,0 +1,374 @@
...
+            enableClickToSpeak: false,
+            ignoreEls: ['header', '.left-bar', '.right-bar', '.modal', 'footer', '.action-bar', '.sts-scope', '.media-container'],
+            toolbarUrl: '//taotoolbar.speechstream.net/tao/configQA.js'
```
ssipasseuth: not familiar with this API but why does the config need to be loaded remotely ? Could we not host our own version of it ?

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -0,0 +1,374 @@
...
+                if ($(this).closest('.settings-menu').length) {
+                    e.stopPropagation();
+                }
+            })
+            .hide(); // Hide settings menu to begin
+
+            // Hide/show volume and speed sliders
+            $this.find('.settings > .settings-menu > .option')
+            .on('hover', function () {
+                $(this).find('.slider-container').show();
```
ssipasseuth: For this line and all below: you could use `ui/hider`for show/hide operation. It allow a centralized control on visibility of elements in TAO.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -0,0 +1,374 @@
...
+
+            // prevents clicks from removing highlighted text
+            $this.on('mousedown', function (e) {
+                e.preventDefault();
+                return false;
+            });
+
+            // Action clicks
+            $this.find('.action').on('click', function (e) {
+                // prevents disabled actions from being triggered
```
ssipasseuth: double instanciation of jquery container, it is preferable to have it build only once: 
```
var $action = $(this);
$action.doStuff();
$action.doOtherStuff();
```

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -0,0 +1,374 @@
...
+            // Show/hide settings menu
+            $this.find('.settings').on('click', function () {
+                $(this).toggleClass('active');
+                $this.find('.settings-menu').toggle();
+            });
+
+            // Settings menu
+            $this.find('.settings > .settings-menu')
+            .on('click', function (e) { // prevent child elements triggering a click on settings menu
+                if ($(this).closest('.settings-menu').length) {
```
ssipasseuth: strange assertion... this condition looks useless as closest() will test the element itself and traversing up through its ancestors. And since the selector '.settings > .settings-menu' is already selecting .settings-menu, it is always true. Is there any specific reason for it being here ?

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -0,0 +1,374 @@
...
+            var $el = this.getElement();
+
+            $('.play', $el).show();
+            $('.pause', $el).hide();
+        })
+        .on('stop', function () {
+            var $el = this.getElement();
+
+            $('.play', $el).show();
+            $('.pause', $el).hide();
```
ssipasseuth: Too many repetition of the same selector. Would be better to get them all in the sam variables like: 
```
$element = {
    play : this.getElement().find('.play'),
    pause : this.getElement().find('.pause'),
    clickToSpeak : this.getElement().find('.click-to-speak'),
    ...
}

File: views/js/runner/plugins/tools/textToSpeech/textToSpeech.js [hunk 9]
```diff
@@ -0,0 +1,374 @@
...
+        .setTemplate(tpl)
+        .on('init', function () {
+            var self = this;
+
+            speed = 40; // default speed
+            volume = 40; // default volume
+
+            // we have to mark some blocks as ignored to prevent TTS accessing it
+            $(options.ignoreEls.join(',')).each(function () {
+                $(this).attr('ignore', true);
```
ssipasseuth: Would it not be simpler to write `$(options.ignoreEls.join(',')).attr('ignore', true);`?

scripts/install/RegisterTestRunnerPlugins.php
ssipasseuth: according to the product owner, this plugin is supposed to be off by default.

scripts/update/Updater.php
ssipasseuth: same as previous comment

scripts/install/RegisterTestRunnerPlugins.php
eacraft13: Yes. After speaking with @markmolenaar, a better approach would be to use a tao-premium extension.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
eacraft13: There are multiple versions of this file and thus needs to be a configurable option. 

I will say there is potential to host this particular file ourselves, as the different versions I believe differ only on configuration options set within that file. Meaning, we could update the config options ourselves and host this file.

An argument against would be that we remove Texthelp's access to this file. In which case, we would need to reupload any updates they might wish to add to these config files.

My personal thought is that Texthelp is a service we consume and that we should push as much work to their end (including config file hosting) as much as possible. That is, keep as much of Texthelp off of tao as we can.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
eacraft13: I think in this case 'body' is a reasonable default. If a dev were initiating this plugin outside of the 'test' context (just a hypothetical, not that it is possible), then it would be safe to assume they want the content of the entire 'body' to be used for text to speech.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
eacraft13: Yes, good call.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
eacraft13: Hah, yes. Good catch. I recall adding this as on an ancestor element, where `closest()` was necessary. Later changing the element of the click event while forgetting to remove `closest()`.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
eacraft13: 1) You inadvertently pointed at that it is better to use `find()` over a selector when you have the jquery element you want to search with already.

2) To your main point, you're correct that these jquery elements are used widely enough to justify an abstraction. I believe, however, that this is not an abstraction I want to create.
    - This would require placing these variables on the `ui/component` object. However, I don't want consumers of this object to ever consider these variables. 
    - Over abstraction can cause slowed development (this would be a good place to cite the article I read on this, but... sorry, it should be a quick google search away if interested). So, when such an instance occurs that abstraction provides little benefit (e.g. `$play` vs. `$el.find('.play')`), I tend not to. For instance, when bug finding, you wouldn't have to trust that the `$play` variable was created correctly.

This is obviously a debatable point, so you're welcome to disagree. But, for this case, I'd need you to support your request for making this change.

eacraft13:
@ssipasseuth, I will be closing this pr in preference for creating a new tao-premium extension from which to install this plugin.

Therefore, tao-4941 will move to creating the new tao-premium extension (i.e., installing tts). And the plugin will remain on https://github.com/oat-sa/extension-tao-testqti/pull/1032.

PR #1038, opened 2017-09-15
Correct timer start and add logger

siwane:


jsconan:
Fix improved in https://github.com/oat-sa/extension-tao-testqti/pull/1060

PR #1055, opened 2017-09-21
Fix/modify item identifer export

antoinerobin:
requires : https://github.com/oat-sa/extension-tao-itemqti/pull/1024


models/classes/export/preprocessor/PreProcessor.php
```diff
@@ -0,0 +1,45 @@
...
+
+
+use qtism\data\storage\xml\XmlDocument;
+
+/**
+ * Interface PreProcessor
+ * Allow to preprocess the test exporter
+ * @package oat\taoQtiTest\models\export\preprocessor
+ */
+interface PreProcessor
```
bugalot: It seems to me it's more an "assessmentItemRefPreProcessor" than a "testPreProcessor". Maybe a more "insightful" name would help.

models/classes/export/preprocessor/IdentifierReplacementPreProcessor.php
```diff
@@ -0,0 +1,58 @@
...
+        foreach ($assessmentItemRefs as $assessmentItemRef){
+            $item = new \core_kernel_classes_Resource($assessmentItemRef->getHref());
+
+            if($item->exists()){
+                $itemRdf  = Service::singleton()
+                    ->getDataItemByRdfItem($item);
+                $identifier = $itemRdf->getIdentifier();
+                $identifier = Format::sanitizeIdentifier($identifier);
+                $assessmentItemRef->setIdentifier($identifier);
+                $items[$identifier] = $item;
```
bugalot: What if multiple items to be injected as assessmentItemRefs share a similar ID? :hankey:

bugalot:
Will now review.

PR #1059, opened 2017-09-22
Feature/tao 5206 remove item href from map

krampstudio:
A new step in the replacement of the `itemHref` in favor of the `itemId`. 

This PR removes the properties `itemDefinition` (that contains the itemHref) and `itemUri` (or `uri`) from the `testContext` and the `testMap`.   The goal is to reduce the size of the data payload, especially for large `testMap` and unify the identification of items withing the TR.

Please note there are still parameters called `itemDefintiion` or `itemHref` which could be misleading, but this will be handled by a separated PR since it's a breaking change - agreed on that way to proceed).


no-chris:
Looks good. Could you please assist in merging `views/js/test/runner/navigator/testContexts.json` with develop?

krampstudio:
I've merged develop.

no-chris:
@krampstudio unit test were still failing I tried to fix them could you please check my fix ?

PR #1061, opened 2017-09-24
Fix/tao 4912 responsive tools

dietertaotesting:


no-chris:
This PR was on `master`, I rebased it.

Code looks good, but there is still a slight issue:
screen shot 2017-09-26 at 09 11 08
This occurs both on resizing and page reloading.

no-chris:
Unit tests ok, but I still get a very light flickering sometimes :-(
It is weird because it happens once the button has been expanded and the window resize continues. With the "right" size, it also happens on reloading. I'm attaching the video to the JIRA ticket, that cannot be done here (https://oat-sa.atlassian.net/browse/TAO-4912)

PR #1062, opened 2017-09-25
Fix/tao 4857 toolbar width

ssipasseuth:
Fixed the width of the toolbar for win10


ssipasseuth:
Opps bad branch sorry

PR #1068, opened 2017-09-26
Fix/tao 5177/incorrect message at the end of the test

ionutpad:


models/classes/runner/map/QtiRunnerMap.php
```diff
@@ -324,12 +324,12 @@ protected function updateStats(&$target, $itemInfos)
 
         if (empty($itemInfos['informational'])) {
             $target['stats']['questions'] ++;
+
+			if (!empty($itemInfos['answered'])) {
+				$target['stats']['answered'] ++;
+			}
```
jsconan: Please be careful with indent while moving blocks of code. Messy code can lead to broken [window effect](https://en.wikipedia.org/wiki/Broken_windows_theory).

jsconan:
By the way, I don't think the issue is a backend one. Seems probably related to the helper that compute/update the stats on the frontend side. Maybe one of those: https://github.com/oat-sa/extension-tao-testqti/tree/develop/views/js/runner/helpers

ionutpad:
As we discuss on slack, I tried to reproduce your issue but seems to be not reproducible anymore, Can you please confirm the existence or not existence of your issue?

If not I think we can go further with this PR.

PR #1071, opened 2017-09-27
Tao 5198 message channel for item navigation

hutnikau:


config/default/testRunner.conf.php
```diff
@@ -369,6 +369,33 @@
...
+                'skip' => array(
+                    'deferred' => false,
+                ),
+                'storeTraceData' => array(
+                    'deferred' => true,
+                ),
+                'timeout' => array(
+                    'deferred' => false,
+                ),
+            ),
```
krampstudio: I keep thinking the grouping of actions should not be made by configuration but using a dedicated API. For example, we should be able to group also the `getItem` call only when the next item is not in the cache. 
Even if this solution work for the current use case, I'm afraid it doesn't offer us all the flexibility required for the long term goal.

hutnikau: to group `getItem` call we just need to implement class for this action and add line to the test runner config.
Could you please describe how you this solution.

PR #1084, opened 2017-10-05
Fix/tao 5053 allow skipping checkbox in tao not correct 3

eacraft13:
Bugfix - ACT
Jira: https://oat-sa.atlassian.net/browse/TAO-5053

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -29,7 +29,7 @@ return array(
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '16.2.1',
+    'version'     => '16.2.2',
     'author'      => 'Open Assessment Technologies',
     'requires'    => array(
         'taoTests'   => '>=6.5.0',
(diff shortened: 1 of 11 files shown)


eacraft13:
Issues to address:

- Placing 'allowSkipping' and 'validateResponses' into one of the category presets was to take advantage of the presets' updating between section and itemref. 
  - Having itemSessionControl update between section, testpart, and itemref was overly difficult.
  - Creating itemSessionControl as a new category for presets is a great option, but overly difficult as well.

- It seems plausible, but I can't figure out how to make one of the presets default to `true`. This would be ideal for 'allowSkipping', as the name matches the qti name and the qti default is `true`. Another option would be to change the name on test creator preset only to 'Prevent Skipping'.

manifest.php
```diff
@@ -29,7 +29,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '16.2.1',
+    'version'     => '16.2.2',
```
jsconan: Since the change is heavier that a simple fix, could you increase the minor version instead?

jsconan:
After having discussed the changes with @bugalot, it appears the solution is not acceptable in the state. Especially since this affects every customers, and breaks the QTI compliance.

This feature should rather be implemented using a customer plugin in the authoring.

krampstudio:
After reading the ticket in JIRA it appears that what is needed is the authoring check-boxes to reflect the values correclty : 
If a test is imported with `allowSkipping=false` at the section level, all items of that sections will have the checkbox unchecked, that's it. No ?

PR #1086, opened 2017-10-10
Fix/tao 5330 n tests n items

bugalot:


bugalot:
Resolves bug TAO-5330 (see description in JIRA).

Please use newly created sample QTI Package to reproduce/test the fix -> test/samples/archives/QTI 2.1/basic/BasicMultipleTestsSameItems.zip

PR #1089, opened 2017-10-10
Fix/tao 5053 allow skipping checkbox in tao not correct 6

eacraft13:
Bugfix - ACT
Jira: https://oat-sa.atlassian.net/browse/TAO-5053


eacraft13:
This is a minimal solution to the problem. It simply updates item session control options at a lower level (e.g. itemref) to take on the parent's (e.g. section) options, if options don't exist for itself. @krampstudio, the test creator already sets the item session control variables correctly for the specific level (i.e. testpart, section, or itemref). My assumption is that setting these options should be reflected in lower levels (aka, inherited).

There are still a couple of negatives to this solution, particularly from a ux perspective.
- Using 'allowSkipping' (as opposed to 'preventSkipping') with a default of true (matching IMS' definition for allowSkipping) creates different behavior than other test creator options (with a default of false). For instance, a user expects, when setting an option to true at the section level, all itemrefs in that section will take on true for that option. This is opposite for allowSkipping and having a value set to true may lead users to believe all itemrefs will take on true as well.
- It is not likely intuitive how setting item session control options among the different levels is reflected in a test (especially in regards to allowSkipping, as addressed in the first point). The previous solution tried to address this through 'presets', which utilize the 'third' state for the options.

Perhaps this simple solution is satisfactory to address the bug, and we can later address the UX issues as a feature?

krampstudio:
Yes, that was the idea.
This issue with scoped option is there for most of the options. It's an annoying issue but we need to fix it globally and privilege consistency (isn't it more weird if some options have a behavior and some have another ?).

krampstudio:
When trying to import a test with `allowSkipping="false"` at different levels (section, itemRef, etc.) I always have the checkbox checked

eacraft13:
I'll create a TT issue regarding the UX once we decide on how to proceed with this.

If I understood you correctly, you think it's weird to use 'allowSkipping' over 'preventSkipping' given the current UX behavior? I think so as well. There are pros and cons to each and I am torn on choosing between the two. I'll keep the current 'allowSkipping' but will change it if you think it best otherwise?

Just to give you quickly my reasoning for keeping 'allowSkipping'...
- It matches qti verbiage and is thus more familiar to users. 
- And, regardless of using 'allowSkipping' or 'preventSkipping', understanding the ux is not likely intuitive for either case (although 'preventSkipping' is perhaps a bit more intuitive).

PR #1095, opened 2017-10-13
Fix/tao 5312 answered item

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-5312

Sometimes, the last item of an intermediate section is not taken into account in the test map.
This is especially visible on a test that contains severals test parts and sections, with the option `x-tao-option-endTestWarning` set on the last item of the test.


krampstudio:
Waiting for @bugalot approval before merging

bugalot:
@krampstudio approved and merged

PR #1113, opened 2017-10-20
Tao 5205 mark offline actions

hutnikau:


views/js/runner/proxy/cache/proxy.js
```diff
@@ -369,6 +371,27 @@ define([
...
+                            _.forEach(data, function (action) {
+                                if (action.parameters.actionId === actionId) {
+                                    action.parameters.offline = true;
+                                }
+                                self.actiontStore.push(action.action, action.parameters);
+                            });
+                        }
+                    });
+                });
+            };
```
krampstudio: Instead of flushing and pushing the actions from store, why don't we just add the information at the 1st push ? For example in https://github.com/oat-sa/extension-tao-testqti/blob/develop/views/js/runner/proxy/cache/proxy.js#L260 

```js
           this.scheduleAction = function scheduleAction(action, actionParams) {
                actionParams.actionId = action + '_' + (new Date()).getTime();
                if(this.isOffline()){
                    actionParams.offline = true;
                }
                return self.actiontStore.push(
                    action,
                    self.prepareParams(_.defaults(actionParams || {}, self.requestConfig))
                ).then(function () {
                    return {
                        action : action,
                        params : actionParams
                    };
                });
           };
```

hutnikau: At time of first push test runner may be in online mode but at the time of execution this action test runner will be offline. I guess we have to set this property at time of execution but not push of action.

krampstudio: Ok, I see what you mean, in that situation, you should add a method to update an action inside the actionStore itself https://github.com/oat-sa/extension-tao-testqti/blob/037305b57d79255dfa7ddd31da51c188c37fa2ae/views/js/runner/proxy/cache/actionStore.js

views/js/runner/proxy/cache/actionStore.js
```diff
@@ -81,6 +81,22 @@ define([
                         });
                     });
                 });
+            },
+
+            /**
+             * update action parameters
+             * @param params
```
krampstudio: Please indicate the type of params, `{Object}` it may looks obvious but it's the rule

views/js/runner/proxy/cache/proxy.js
```diff
@@ -369,6 +371,19 @@ define([
                     });
                 });
             };
+
+            /**
+             * Mark action as performed in offline mode
+             *
+             * @param actionId
```
krampstudio: Please set also the parameter type in the doc

PR #1122, opened 2017-10-27
Tao 5429/tts doesnt work in combination with formulae

IlyaYarkavetsOnePoint:
Added missing define in textToSpeech.js to make sure that mathJax lib is loaded before textToSpeech component;


ssipasseuth:
It seems to be working however could you tell me why loading mathjax before fixes this issue ? Mathjax is normally loaded at the same time as the item rendering so I don't understand why the TTS need mathjax loaded as the same time as him. Any clarification will be great, thanks !

IlyaYarkavetsOnePoint:
No problem. This is hidden deeper in tts modules. At first take a look at textToSpeech.js - "on('init'..." handler: it requires configQA.js, after that in callback it requires texthelpMain.js. Inside it, somewhere near line 12340 they have internal check for mathJax lib
```    
if (Cha || typeof (MathJax) != "undefined") {
        var cIb = "sre_browser.js";
        if (cfa || Aga) {
            cIb = "sre_mathjax.js";
        }
        if (Dha != "") {
            cIb = Dha;
        }
        if (Bha != "") {
            SpeechStream.MathSpeak.bXc("//" + Bha + "/MathSpeak/" + Aha + "/" + cIb, SpeechStream.MathSpeak.callBack);
        } else {
            SpeechStream.MathSpeak.bXc("//" + yca + "/MathSpeak/" + Aha + "/" + cIb, SpeechStream.MathSpeak.callBack);
        }
    }
```
If mathjax lib is not loaded at this moment lib sre_browser.js (which is, as I understand, is needed for tts to "pronounce" formulae) will not be loaded - so we won't get working tts for formulas. I spent some time debugging it. And it seems to be working with a fix like that (adding mathJax to define block). Possibly it can be moved somewhere lower to initialize mathJax later.

ssipasseuth:
Thanks for the explanation, it is enlightening. So by change the TTS also support MathJax natively. That is quite a happy coincidence. Nice catch !

PR #1129, opened 2017-11-02
Feature/tao 5470 color contrast

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-5749

Ensure the needed events are properly triggered when using the themes switcher tool.

### DIFF
--- a/views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
+++ b/views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
@@ -35,11 +35,17 @@ define([
     'use strict';
 
     /**
+     * The public name of the plugin
+     * @type {String}
+     */
+    var pluginName = 'itemThemeSwitcher';
+
+    /**
      * Returns the configured plugin
      */
     return pluginFactory({
 
-        name: 'itemThemeSwitcher',
+        name: pluginName,
 
         /**
          * Installation of the plugin (called before init)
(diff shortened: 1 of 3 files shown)


views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -111,14 +122,15 @@ define([
...
                 icon: 'contrast',
                 text: __('Contrast')
-            });
-
-            this.menuButton.on('click', function (e) {
+            }).on('click', function (e) {
                 e.preventDefault();
                 testRunner.trigger('tool-themeswitcher-toggle');
+            }).on('openmenu', function() {
+                testRunner.trigger('plugin-open.' + pluginName, state.selectedTheme);
```
krampstudio: Why don't you use the method trigger on the plugin instance itself  ?
```
self.trigger('open', state.selectedTheme);
```

https://github.com/oat-sa/tao-core/blob/master/views/js/core/plugin.js#L199

views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -111,14 +122,15 @@ define([
...
+            }).on('closemenu', function() {
+                testRunner.trigger('plugin-close.' + pluginName, state.selectedTheme);
```
krampstudio: same here

views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
jsconan: I firstly used the plugin trigger method, however it adds extra parameter that pollutes the arguments, and breaks the compatibility with regular events in the context of probes.

plugin trigger:
```
event(plugin, ...)
```

testRunner trigger:
```
event(...)
```

PR #1136, opened 2017-11-09
Act 94 fix missed time point

hutnikau:


krampstudio:
This should be reworked. It works with the ACT configuration but won't work with another. 
This parameter `start` was used to tell the server to start the timer. 
Previously we had 2 calls to the server, for example

PR #1142, opened 2017-11-16
Feature/tao 5552/echo initialization behavior update

tikhanovichA:


views/js/runner/provider/qti.js
```diff
@@ -226,7 +233,9 @@ define([
              * @param {Promise} [loadPromise] - wait this Promise to resolve before loading the item.
              */
             function computeNext(action, params, loadPromise){
-
+                var dialog = false;
```
zagovorichev: All that concerns this functionality can be moved to a new component. Also all constants should be configurable and came from the backend.
I mean these:
```
var echoDelayUpdate = 15;
var echoPauseLimit = 120;
var echoExceptionName = 'CatEngine';
```

views/js/runner/helpers/testGeneration.js
```diff
@@ -0,0 +1,105 @@
...
+        'Please notify your proctor now, who will authorize you to relaunch your test. If the system is able to load your next items, you will continue testing where you left off.');
+
+    var dialog = false;
+    var time = 0;
+    var timeout = 0;
+
+    /**
+     * @typedef {Object} testGenerationHelper
+     */
+    var testGenerationHelper = {
```
zagovorichev: I think you should use `'ui/component'` to avoid usage of the testRunner and send triggers on each of the user actions

PR #1143, opened 2017-11-16
Fix/tao 5352 informational answered

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-5352

The stats updater is taking care of the `answered` flag of each items to count the number of answered items. Sometimes informational items are considered as answered by the server, and that leads to wrong number in the map stats. The issue was previously fixed in the server side map builder, not on client side. Since we also build/update this map mostly on the client side the issue reappeared.

It is not easy to reproduce the issue as it seems to be based on the test content. Most of the time I spent on this issue was to reproduce it. You need a test that contains a CAT section, followed by a linear test part starting by an informational item and containing questions. You can find a test package attached to the JIRA ticket. Answer all questions till the end of the test, and you should see a message with negative number of answered items.


no-chris:
Works well but it breaks unit tests... a good occasion to assert new test cases ? ;-)

jsconan:
My bad! fixed now, and also updated to take care of the fix

PR #1145, opened 2017-11-20
Act 87 fix nonconnectivity error processing

hutnikau:


krampstudio:
I don't have access to support tickets. Please describe the bug and steps to reproduces.

hutnikau:
1. We need to pass message to the `exit()` function to display it on the return page specified during lti launch. For example you can emulate XSRF attack and you will see that with this fix after redirecting user to the exit page he will see appropriate message.
2. Also In `proxy.js` we need to check if it was connectivity error. For example in case of 403 response code, which is not connectivity error (you may test it with emulating of XSRF attack as well) user should be redirected to the return url and no offline processing should be done.

PR #1146, opened 2017-11-21
Use result storage but not result serer instance

hutnikau:


views/js/controller/runner/runner.js
```diff
@@ -104,7 +104,7 @@ define([
                         warning: reason
                     });
                 }
-                window.location = url;
+                //window.location = url;
```
ionutpad: I think it's a debug line, Please remove the commented line

PR #1148, opened 2017-11-21
Tao 5505/tts cannot click to speak choice content

IlyaYarkavetsOnePoint:
Changes needed for [this pr](https://github.com/oat-sa/extension-tao-itemqti/pull/1068) to work


manifest.php
```diff
@@ -29,7 +29,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '17.17.0',
+    'version'     => '17.17.1',
```
jsconan: As there is a dependency on other extension, pease also update the version in the `requires` clause below.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -239,12 +239,28 @@ define([
...
+                var self = this;
+
                 options.enableClickToSpeak = !options.enableClickToSpeak;
 
                 this._exec('enableClickToSpeak', options.enableClickToSpeak);
                 this._exec('enableContinuousReading', !options.enableClickToSpeak); // continuous reading is off when click to speak is on
                 this.trigger('clickToSpeak');
 
+                //adding each item a special class by presence of which normal click handling could be prevented and passed to click-to-speak handling
+                $('.qti-item').each(function() {
```
jsconan: The TestRunner provider exposes an areaBroker object that gives access to the DOM. Please use it.

```javascript
testRunner.getAreaBroker().getContentArea()
```

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -239,12 +239,28 @@ define([
...
+                    var $_self = $(this);
```
jsconan: Is this var actually needed ? you should inline it, as there is only one use.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -239,12 +239,28 @@ define([
...
+                    $_self.toggleClass('prevent-click-handler');
+                });
+
+                //we should disable click-to-speak while navigating through test, if it was enabled on some item page
+                //just to ensure that after other item load click-to-speak will function normally, and tao click handlers won't work on item-part click
+                $('.navi-box').on('click', 'a', function() {
```
jsconan: The TestRunner provider exposes an areaBroker object that gives access to the DOM. Please use it.

```javascript
testRunner.getAreaBroker().getNavigationArea()
```

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -239,12 +239,28 @@ define([
...
+                    if ($('.qti-item.prevent-click-handler').length > 0) {
```
jsconan: The TestRunner provider exposes an areaBroker object that gives access to the DOM. Please use it.

```javascript
testRunner.getAreaBroker().getContentArea()
```

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
```diff
@@ -349,7 +365,7 @@ define([
                     max: 100
                 },
                 start: speed,
-                step: 10
+                step: 15
```
jsconan: What does this value mean? Could you explain it with a comment?

IlyaYarkavetsOnePoint: Yep, this is the part of the changes requested by textHelp guys. Actually, this value is the step for speed slider. Now the step will be 15. So, values that can be chosen with help of slider are: 0, 15, 30, etc., etc., 100. Some time later, also "border values" of the slider will be fixed, as textHelp requested from us that only some speeds are available to choose to end-user - it's for caching purposes.

manifest.php
IlyaYarkavetsOnePoint: Ok, thx, I'll changes this in taoQtiItem repo.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
IlyaYarkavetsOnePoint: Thanks! Didn't know that we can access it through brokers. Will fix this.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
IlyaYarkavetsOnePoint: Thanks! Didn't know that we can access it through brokers. Will fix this.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
IlyaYarkavetsOnePoint: So, actually, as I see this would be changes to smth like:
```
options['$navigationArea'].on('click', 'a', function() {
                    if (options['$contentArea'].find('.qti-item.prevent-click-handler').length > 0) {
```
Or am I wrong?

jsconan: Well, you should take at look at the build of the areaBroker to know more about the available areas and what are the actual targets:
https://github.com/oat-sa/extension-tao-testqti/blob/master/views/js/runner/provider/qti.js#L69

jsconan: Yes, as the TTS compoent does not have direct access to the TR, you must rely on delegated properties like that.

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
jsconan: I meant: please put a comment in the code to explain the meaning of the value, for other developers this will be helpful :)

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
IlyaYarkavetsOnePoint: Tank you for making it clearer for me :)

views/js/runner/plugins/tools/textToSpeech/textToSpeech.js
IlyaYarkavetsOnePoint: Awwww, I understand, will comment it )

IlyaYarkavetsOnePoint:
Made the requested changes.

manifest.php
jsconan: Well, the change is not meant to be made in the other extension, but in this one. In the `requires` clause there is the required version of `taoQtiItem`. You should increase this number in this file to ensure the requirements are met.

IlyaYarkavetsOnePoint: Hmmm, as I understand it will look like this:
We are changing this repo version. We are saying that this repo requires at least __some__ version of the other repo. But it is not this repo relying on other repo changes but these repo changes are need for other repo to work. Maybe it's better to change `taoQtiItem` repo `requires` block? As it needs at least __some__ version of this repo? Am I wrong? Or it doesn't matter?
p.s.
Sorry for having so much questions :)

jsconan: No problem for the questions. We are always learning, and the better way is to exchange info.

So, you should see the requirements as a hierarchical tree. `taoQtiItem` only takes care of items. `taoQtiTest`, takes care of tests, that contain items. So `taoQtiTest` is using `taoQtiItem`, and it requires it. In the other hand, `taoQtiItem` can be used outside of `taoQtiTest`, and don't require it. See that as boxes containing boxes. `taoQtiItem` cannot even know about `taoQtiTest`, otherwise you will go into impossible situations, where each one is the father of the other.
For some features, `taoQtiTest` must ensure the other extensions it is relying on are in the right version that bring the wanted behavior/api. So it must declare the minimum version of the needed dependencies.

Is that clear enough? :)

IlyaYarkavetsOnePoint: Yeah, thanks, now, I think, I understood what you wanted me to changed and, the main part, - why :)

IlyaYarkavetsOnePoint:
Updated requirements

jsconan:
Works as expected. GG!

PR #1156, opened 2017-11-23
Tao 5356 store cat session after last item

hutnikau:
In case if test taker on the last item presses `previous` button or jumps to any previous item the last item will be submitted and cat session should be stored. Otherwise shadow cat will always contain the last item as if it was not submitted.
Depends on https://github.com/oat-sa/lib-test-cat/pull/21


jbout:
Reverted due to issues on update script

PR #1158, opened 2017-11-23
Revert "Tao 5356 store cat session after last item"

jbout:
Reverts oat-sa/extension-tao-testqti#1156


jbout:
Revert failed, needed to reset the history --hard

PR #1167, opened 2017-11-28
Fix/tao 5552/fix cat engine warning message

tikhanovichA:


views/js/runner/plugins/navigation/previous.js
```diff
@@ -62,6 +62,11 @@ define([
                 var previousSection;
                 var previousPart;
 
+                // check TestMap if empty
+                if (Object.keys(testMap).length === 0 && testMap.constructor === Object) {
```
krampstudio: Could you just check instead (for code consistency)
```
if(  _.isPlainObject(testMap) && _.size(testMap) === 0){
```

PR #1170, opened 2017-11-30
Tao 5577/improve tts caching limit speeds specify ids

IlyaYarkavetsOnePoint:
Some fixes required by textHelp;
From now on if we are in multi-tenant context and `tenantName` is available - `tenantName` will be used as primary cache identifier (`bookId`) for caching on textHelp side, if not - `deliveryUri` will be used as `bookId` as it was before;

Changed extension requirements:
From now on minimal required version of tao core is 14.4.1, as only in that version `containerService` was introduced;


views/js/runner/plugins/tools/textToSpeech/plugin.js
```diff
@@ -109,13 +109,28 @@ define([
                     self.enable();
                     self.ttsButton.show();
 
+                    var bookId;
+                    //If we are in multi-tenant context then we choose tenantName as bookId for caching
+                    //in other cases - deliveryUri will be used as bookId for caching on textHelp side
+                    if (typeof config.tenantName !== typeof undefined && config.tenantName && config.tenantName !== null) {
```
ssipasseuth: this 3 checks are equivalent to a simple: `if(config.tenantName){...}`

manifest.php
```diff
@@ -29,12 +29,12 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '17.21.7',
+    'version'     => '17.21.8',
```
ssipasseuth: it is rather a new feature than a fix, so 17.22.0 is more appropirate

models/classes/container/QtiTestDeliveryContainer.php
```diff
@@ -65,6 +66,15 @@ public function getExecutionContainer(DeliveryExecution $execution)
...
         $container->setData('deliveryExecution', $execution->getIdentifier());
         $container->setData('deliveryServerConfig', []);
+
+        //setting tenantId/Name if it exists (and we are in multi-tenant context)
+        //as for now it is needed for textHelp caching as a bookId
+        $tenantService = $this->getServiceLocator()->get(ContainerService::SERVICE_ID);
+
+        $tenantName = method_exists($tenantService, 'getTenantName') ? $tenantService->getTenantName() : false;
+
+        $container->setData('tenantName', $tenantName);
```
ssipasseuth: you cannot pass this information like that, this  service is aimed at providing very generic data about the the delivery.
you may want to look into examples of the best practice about configuring a test plugin, e.g. check the updater of taoQtiTest or taoAct, you got example like
```
$config = $extension->getConfig('testRunner');

           

            $config['plugins']['magnifier'] = [
                'zoomMin' => 2,
                'zoomMax' => 8,
                'zoomStep' => .5
            ];

            $extension->setConfig('testRunner', $config);
```
Call it in the extension you want this specific information to be added.

views/js/runner/config/qtiServiceConfig.js
```diff
@@ -46,7 +46,8 @@ define([
         testCompilation : true,
         serviceCallId : true,
         bootstrap : false,
-        timeout : false
+        timeout : false,
+        tenantName : false
```
ssipasseuth: same as comment above, you should not pass plugin specific info or it will be mess soon

PR #1173, opened 2017-12-08
Fix/tao 5669 round second up

ionutpad:
I decide to round the entire timestamp because, in a case that we have 123.7567890123 this will be rounded to 123.756789, but since we are converting this to seconds I think a round up it's better because if you start to count a second (and it's more that a half) the second already passed so the value should be 124.

The inside functionality of DateInterval it's converting this kind of value 123.7567890123 by round it down to 123, and it's should be 124

Because of that the time_remaining was all the time lower that expected. In a case that you have the section_timer 123456/0 the time_remaining was 00:00:01 instead of 00:00:00 (this is a simply example by all of the time_remaining's was more that expected).


krampstudio:
I can confirm that. rounding duration will have other side effects. May be the solution is to use the precise duration in the other places too. No ? The sent duration should be also using the same precision.

PR #1178, opened 2017-12-18
Feature/tao 5509 pre cache audio

krampstudio:
Requires https://github.com/oat-sa/extension-tao-test/pull/201
Requires https://github.com/oat-sa/extension-tao-item/pull/291

Precaching audio files is more difficult than images because browsers implement different behaviors for `<audio preload="auto">`, `<link rel="prelaod" as="audio">` or `<link rel="prefetch" as="audio">` and none of the previous loading ways were working consistently across browsers. 

In order to preload audio files, we load them directly as `blob` (thanks to xhr2's `responseType='blob'`). we keep the blob in memory (without using a string representation)  and then we access to the blob using `URL.createObjectURL`.
In order to change the URL, we prepend a strategy to the asset manager that resolves against the cached files.

To test the PR, please use a test with multiple items, each one with an audio interaction. Once preloaded you should be able to play the item's audio offline (by stopping your web server for instance.


krampstudio:
Yes, I'll address the polling issue in a separate PR

PR #1190, opened 2018-01-04
Tao 5769 fix session termination

hutnikau:
depends on:
https://github.com/oat-sa/extension-tao-testqti/pull/1190
https://github.com/oat-sa/extension-tao-proctoring/pull/533

### DIFF
--- a/models/classes/runner/QtiRunnerService.php
+++ b/models/classes/runner/QtiRunnerService.php
@@ -145,6 +145,7 @@ class QtiRunnerService extends ConfigurableService implements RunnerService
      * @param string $testDefinitionUri The URI of the test
      * @param string $testCompilationUri The URI of the compiled delivery
      * @param string $testExecutionUri The URI of the delivery execution
+     * @param string $userUri User identifier. If null current user will be used
      * @return QtiRunnerServiceContext
      * @throws \common_Exception
      */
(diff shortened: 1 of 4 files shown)


models/classes/runner/QtiRunnerService.php
```diff
@@ -148,13 +148,14 @@ private function loadItemData($itemRef, $path)
      * @return QtiRunnerServiceContext
      * @throws \common_Exception
      */
-    public function getServiceContext($testDefinitionUri, $testCompilationUri, $testExecutionUri)
+    public function getServiceContext($testDefinitionUri, $testCompilationUri, $testExecutionUri, $userUri = null)
```
zagovorichev: Add new param to the commentary block, please

models/classes/runner/QtiRunnerService.php
```diff
@@ -148,13 +148,14 @@ private function loadItemData($itemRef, $path)
...
     {
         // create a service context based on the provided URI
         // initialize the test session and related objects
         $serviceContext = new QtiRunnerServiceContext($testDefinitionUri, $testCompilationUri, $testExecutionUri);
         $serviceContext->setServiceManager($this->getServiceManager());
         $serviceContext->setTestConfig($this->getTestConfig());
+        $serviceContext->setUserUri($userUri);
```
zagovorichev: Would be great to have condition:
```
if ($userUri) {
$serviceContext->setUserUri($userUri);
}
```

zagovorichev: also I don't like that from now we can change user for the context, maybe you can do it in the __construct, don't you?

PR #1204, opened 2018-01-22
Feature/tao 5148 test runner storage

krampstudio:
Implement the `testStore` in the QTI Test Runner and adapt the plugins.

Please note there's a remainder. 
The `testState` which is a `cachedStore` is not yet migrated to the testStore (only async).

**Requires** 
- [ ] https://github.com/oat-sa/tao-core/pull/1621
- [ ] https://github.com/oat-sa/extension-tao-test/pull/210


views/js/runner/provider/qti.js
```diff
@@ -491,15 +501,8 @@ define([
...
                     self.dataUpdater.updatePluginsConfig(self.getPlugins());
 
-                    //check if we need to trigger a storeChange
-                    if(!_.isEmpty(storeId) && !_.isEmpty(results.lastStoreId) && results.lastStoreId !== storeId){
-
-                        /**
-                         * We have changed the local storage engine (could be a browser change)
-                         * @event runner/provider/qti#storechange
-                         */
-                        self.trigger('storechange');
```
no-chris: Is it safe to remove this event? Or could this be considered a breaking change?

scripts/update/Updater.php
```diff
@@ -1762,6 +1762,6 @@ public function update($initialVersion) {
             $this->setVersion('18.9.5');
         }
 
-        $this->skip('18.9.5', '19.0.1');
+        $this->skip('18.9.5', '19.1.0');
```
jsconan: Version conflict with the manifest:
- manifest `20.0.0`
- update `19.1.0`

views/js/runner/provider/qti.js
jsconan: I think this is ok since the plugins that listened to this event have also been updated accordingly

no-chris:
This breaks some plugin tests. Maybe the provider mock needs to be adapted?

PR #1211, opened 2018-02-05
added test store to provider mock

no-chris:


views/js/test/runner/mocks/providerMock.js
```diff
@@ -64,6 +65,18 @@ define([
...
+            /**
+             * Initialize and load the test store
+             * @returns {testStore}
+             */
+            loadTestStore : function loadTestStore(){
+                var config = this.getConfig();
+
+                //the test run needs to be identified uniquely
+                var identifier = config.serviceCallId || 'test-' + Date.now();
+                return testStoreFactory(identifier);
```
krampstudio: May be using the in-memory backend would be more suitable : 
```
 return testStoreFactory(identifier, store.backends.memory);
```

no-chris:
ok fixed

PR #1212, opened 2018-02-06
added option to load portable element directly from loaded item data

ssipasseuth:
1. added new method getItemPortableElements() to interface RunnerService.php (result in new major version)
2. load portable element data with the getItemData()


bugalot:
Seems good to me thanks !

PR #1218, opened 2018-02-13
Fix/construct servicemanager

jbout:


jbout:
Already solved by someone else & merged

PR #1236, opened 2018-02-23
tweaked the lock icon style

ssipasseuth:
tweaked the lock icon style:
1. reduce the line of the lock button
2. aligned the lock button with respect to the inputs
3. made more room for the lock button on the right by reducing the time spinner. only 2 digits are allowed anyway


ssipasseuth:
already copied and merged into the main branch

PR #1242, opened 2018-03-02
Added access for the RestPublisher role to the entry points of the RE…

zagovorichev:
…ST API
depends on https://github.com/oat-sa/tao-core/pull/1666


zagovorichev:
@ionutpad Update

PR #1252, opened 2018-03-21
clear storage array after obtaining data from storage

hutnikau:


krampstudio:
what is the issue behind this ? db size ?

PR #1255, opened 2018-03-22
Fix/tao 6006 validate authoring identifiers

krampstudio:
About https://oat-sa.atlassian.net/browse/TAO-6006

Fix and test the unique id validator within the test authoring.

If you create a test with non unique identifiers in develop and try to save it, I've added a warning to let you know an id is not unique (not perfect, but helps in addition to the field validation)

### DIFF
--- a/views/js/controller/creator/helpers/qtiTest.js
+++ b/views/js/controller/creator/helpers/qtiTest.js
@@ -106,16 +106,28 @@ define([
 
         /**
          * Gives you a validator that check if a QTI id is available
-         * @param {Array} lockedIdentifiers - the list of identifiers you cannot use anymore
+         * @param {Array} lockedIdentifiers - deprecated, get them trough the modelOverseer
+         * @param {Object} modelOverseer - let's you get the data model
          * @returns {Object} the validator
          */
-        idAvailableValidator : function idAvailableValidator(lockedIdentifiers){
+        idAvailableValidator : function idAvailableValidator(lockedIdentifiers, modelOverseer){
+            var self = this;
             return {
                 name : 'testIdAvailable',
                 message : __('is already used in the test.'),
-                validate : function(value, callback, options){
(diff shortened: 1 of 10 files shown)


views/js/controller/creator/helpers/qtiTest.js
```diff
@@ -217,8 +229,24 @@ define([
...
+                .forEach(function(count, id){
+                    if(count > 1){
+                        nonUniqueIdentifiers.push(id);
+                    }
+                });
+            if(nonUniqueIdentifiers.length === 1){
+                throw new Error(__('The identifier "%s" is not unique accross the test.', nonUniqueIdentifiers[0]));
+            }
+            if(nonUniqueIdentifiers.length > 1){
+                throw new Error(__('The following identifiers "%s" are not unique accross the test.', nonUniqueIdentifiers.join('", "')));
```
ssipasseuth: Could be more useful, if the error message indicates elements is faulty by highlighting it. But since it is an edge case, could be a separate story, we should as product owner.

ssipasseuth:
As discussed with @markmolenaar , both issues can be addressed later as part of other tasks.

views/js/controller/creator/helpers/qtiTest.js
ssipasseuth: As a rule of thumb, when displaying an error message, the next step for the user should be clear, obvious and easily doable.

krampstudio: Yes I was thinking to it also, but there's no link between the ids and the elements in the current version. 
And I was balancing 2 hours of developments against an error that _should_ never occur, the only way to get there is to have created a test with the bug in the previous version. 
This error message is just a last barrier to not save a broken test.

krampstudio:
Maybe we should revisit how we handle errors, more generally.

PR #1256, opened 2018-03-23
Fix/tao 5036/sprint73/fix exit message

tikhanovichA:


views/js/runner/helpers/stats.js
```diff
@@ -47,6 +47,8 @@ define([
 
             } else if (isItemCurrentlyAnswered && !context.itemAnswered) {
                 stats.answered++;
+            } else if(!isItemCurrentlyAnswered && !context.itemAnswered) {
+                stats.answered++;
```
jsconan: I don't understand the purpose here. The result of this code will be, no matter the value of `isItemCurrentlyAnswered`:

```javascript
if (!context.itemAnswered) {
    stats.answered++;
}
```

I don't think this is a wanted behavior, as the item will always be considered answered. This will introduce regression.

PR #1261, opened 2018-03-27
Tao 5862 exit from test in case of server error

hutnikau:
depends on https://github.com/oat-sa/extension-tao-delivery/pull/324


hutnikau:
@krampstudio just updated PR according to your comments

PR #1263, opened 2018-03-29
Feature/tao 5974 configurable progressbar

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-5974

This PR aims to make the progress bar plugin more configurable.
To do so, the progress computation algorithms have been put in a proper helper. This helper provides 3 main entries:
- `computeStats`: computes stats about the current progression within the wanted scope
- `computeIndicator`: computes the indicator data to feed the renderer, according the wanted type
- `computeProgress`: computes the current progress based on the provided context and config

An improvement is brought by this helper, as every kind of progressbar should be now scopable, so the `percentage` indicator type can be scoped as well as other types. This is why the updater take care of that by ensuring the scope is well configured to keep consistency with installs that are using this type with a wrong scope config.

In the same time a new algorithm has been added to take care of the feature required by the story: `sections`. It takes care of each reached non informational section. That means the progression is related to the viewed number of sections that contain at least one answerable item.

The rendering of the progress bar is now delegated to an external and configurable renderer. For now 2 renderers are available:
- `percentage`: this is the standard view, with a linear bar that reflects the current percentage of progress based on the current context
- `position`: this is a new way of representing the progression. It displays one case by element in the current scope. This renderer should be used along with the new `sections` indicator type. This could be also used with any type of indicator, but in that case a high number of elements could be an issue.

The plugin is also now wrapped into a factory that should allow to force a particular config. This will be needed for the second part of this story, when the progress bar mode will be set from a delivery feature set (a particular flavor of the progressbar will be wrapped with the wanted config).


views/js/runner/helpers/progress.js
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2018 (original work) Open Assessment Technologies SA ;
+ */
+
+/**
+ * Compute the current progress in the Test Runner
```
krampstudio: Is this _helper_ used by other component than the progress plugin ?

views/js/runner/plugins/controls/progressbar/plugin.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Returns the configured plugin
+     * @param {String} pluginName - The name of the plugin
+     * @param {Function} [configurator] - A callback that will customize the config
+     * @param {Object} [renderers] - A list of renderers to assign
+     * @returns {plugin} Returns a tweaked progressbar plugin
+     */
+    return function progressbarPluginFactory(pluginName, configurator, renderers) {
```
krampstudio: I think this pattern is the inverse of what it should be.
I think the plugin itself shouldn't be wrapped in a factory. The plugin remains the same, it react to the same events, etc. 
However during the `install` or the `init` step the plugin selects the progress component it should use based on the configuration.

My concern is that adds a new step in the plugin loading mecanism. Here you want to configure another rendering of the plugin, but this can be anything else, so the feature configuration can be just setting a configuration option.

krampstudio:
Just a simple question. We have the case where we want to count the progress only of elements that have a given category. How do you think we could handle this ?

views/js/runner/plugins/controls/progressbar/plugin.js
jsconan: The plugin already uses configuration options, the way you describe it.
The wrapper/factory adds the ability to package the plugin. The goal is to bring the ability to package plugin with forced options. This is to be used along with the `feature set` that can be chosen from the delivery (check boxes). In that context we may only select which plugin to enable/disable, not the options we want.

jsconan:
For the categories, we could add a new indicator type, that match the wanted categories, and also augment the progress stats with a new entry that targets the count of matched categories

views/js/runner/helpers/progress.js
jsconan: For now, this helper is only used by the progress plugin.

krampstudio: I was wondering why the stats of the tests aren't sufficient to compute the progress : 
1. if this is the case and this helper is only about formatting the stats, then it should be moved inside the plugin folder too ?
2. OR  It fill a gap in the testMap that contains inaccurate data (ie. `viewed` values are not up to date), in that situation shouldn't we update directly the testMap ?

jsconan: Correct, this is for the first case: the regular stats are not sufficient to compute the progress.
I placed this helper as a common helper because I was wondering if other plugins could benefit of it. If you think this is too much specific to the progress bar plugin and useless for others, I can move it to the plugin folder.

krampstudio: My main concern is that it looks like the place the progression is computed, globally and can lead to misunderstanding. 
It's safer in my opinion to move it into the plugin folder. If we have a case that needs it we will generalize it.

krampstudio:
When testing I've also found a regression, before your PR in a vanilla version of TAO, compare the result of skipping items. By default (on develop) the progression continues, but with your branch, the progress remains the same

views/js/runner/helpers/progress.js
jsconan: ACK, I will move it

jsconan:
Damn! I will check that

PR #1265, opened 2018-04-03
Fix/fix/tao 5036/sprint74/fix exit message

tikhanovichA:


views/js/runner/helpers/stats.js
```diff
@@ -48,6 +49,10 @@ define([
             } else if (isItemCurrentlyAnswered && !context.itemAnswered) {
                 stats.answered++;
             }
+
+            if (sync && !isItemCurrentlyAnswered && !context.itemAnswered) {
+                stats.answered++;
+            }
```
jsconan: Wouldn't it be better/simpler to do instead:
```javascript
        if (!item.informational) {
            isItemCurrentlyAnswered = currentItemHelper.isAnswered(runner);
            if (!isItemCurrentlyAnswered && context.itemAnswered) {
                stats.answered--;
            } else if ((isItemCurrentlyAnswered || sync) && !context.itemAnswered) {
                stats.answered++;
            }
        }
```

views/js/runner/helpers/messages.js
```diff
@@ -31,17 +31,18 @@ define([
      * @param {String} message - custom message that will be appended to the unanswered stats count
      * @param {String} scope - scope to consider for calculating the stats
      * @param {Object} runner - testRunner instance
+     * @param {Boolean} sync - flag for sync the unanswered stats in exit message and the unanswered stats in the toolbox. Default false
      * @returns {String} Returns the message text
      */
-    function getExitMessage(message, scope, runner) {
+    function getExitMessage(message, scope, runner, sync = false) {
```
jsconan: I missed that, but could you remove the default value from the function declaration? The browsers we are supposed to support don't have the support of this feature (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)

PR #1278, opened 2018-04-26
Moving to the next item without an alert

zagovorichev:


zagovorichev:
@dietertaotesting Update

PR #1281, opened 2018-05-04
Fix/tao 6186 progress percentage

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-6186

The percentage indicator now reflects the actual progress as it only takes care of the questions, and ignore the informational items.
On a **non linear** test part, it should only count the answered items, unless when they are informational.
On a **linear** test part, it should count every question, not matter whether answered or not, but should still ignore the informational items.

The progress helper has been slightly refactored to allow such behavior. The unit test has also been updated, especially since not all features were properly tested.

The handling of the item answered state is also linked to the linear mode of the current test part.

You should find some sample tests attached to the ticket.


krampstudio:
This behavior seems to be expected

PR #1290, opened 2018-05-17
fix/TAO-6262-deliveryexecution-item-state-deletion-route

bugalot:
2 Pull Requests are related, but not absolutely required. They were developed for testing but will certainly be of a good help to other developers and devops in the future.

https://github.com/oat-sa/extension-tao-delivery/pull/339
https://github.com/oat-sa/extension-tao-delivery-rdf/pull/195


models/classes/cat/GetDeliveryExecutionsItems.php
```diff
@@ -74,17 +74,25 @@ public function __construct(
      */
     public function getItemsRefs()
     {
-        $sections = $this->assessmentTestSession->getAssessmentTest()->getComponentsByClassName('assessmentSection');
-        $items = [];
-        foreach ($sections as $section) {
-            $assessmentItemsRef = $this->getSectionItems($section, $this->deliveryExecution->getDelivery());
```
ionutpad: It may be out of topic, but can you please remote the method `getSectionItems` because it's not used anymore.

PR #1297, opened 2018-05-28
Feature/tao 5701 previewer

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-5701

Requires: 
 - [x] https://github.com/oat-sa/tao-core/pull/1586
 - [x] https://github.com/oat-sa/tao-core/pull/1750

Apply some changes required for the new item previewer.
The implementation of the previewer has been moved to [taoQtiTestPreviewer](https://github.com/oat-sa/extension-tao-testqti-previewer/pull/1)

___
Implements the item previewer using the architecture of the new test runner. It registers an adapter that will be loaded by the taoItems previewer.

**Notes:**
The new previewer expects the item data to be provided in JSON, that is ok for the items that are built for the new TestRunner. The item built for the old TestRunner are provided in XML, and therefore need to be translated to JSON on the fly. To do so we apply the item packer, quite the same way it is done at compilation time. That works fine, unless we delete the source item. To be able to translate the XML item into JSON, we do need the source item. There is no way to achieve that process differently, unless we rewrite the item packer, or unless we fork it completely to be able to only use the XML version.

Also, please note that only the compiled items are taken into account. The backend cannot provided data for source item right now (preview item in authoring).

The console has been implemented client side, but was not on the backend. This shouldn't be a problem for the preview of results, as the console is disabled, but it needs to be available for the preview in authoring. It should be done in another story.

Also, the device size emulator still needs to be implemented in the new previewer.


krampstudio:
@jbout we need your advice, I'm wondering if the "Test∕Previewer" needs to be put in a dedicated extension.

actions/class.Previewer.php
```diff
@@ -0,0 +1,225 @@
...
+            $code = $this->getErrorCode($e);
+        }
+
+        $this->returnJson($response, $code);
+    }
+
+    /**
+     * Stores the state object and the response set of a particular item
+     */
+    public function submitItem()
```
krampstudio: Is it needed for the previewer ?

models/classes/ItemPreviewer.php
```diff
@@ -0,0 +1,279 @@
...
+use oat\taoItems\model\pack\ItemPack;
+use oat\taoItems\model\pack\Packer;
+use oat\taoQtiItem\model\qti\Service;
+use oat\taoQtiItem\model\QtiJsonItemCompiler;
+use oat\taoQtiTest\models\container\QtiTestDeliveryContainer;
+use oat\taoResultServer\models\classes\ResultServerService;
+use Zend\ServiceManager\ServiceLocatorAwareInterface;
+use Zend\ServiceManager\ServiceLocatorAwareTrait;
+
+class ItemPreviewer implements ServiceLocatorAwareInterface
```
krampstudio: I'm not sure this modeling is correct. It may not be  linked to the previewer, I can see other uses cases when we can load the compiled item data.

models/classes/ItemPreviewer.php
```diff
@@ -0,0 +1,279 @@
...
+    public function __construct()
+    {
+        $this->fileStorage = \tao_models_classes_service_FileStorage::singleton();
+    }
+
+    /**
+     * @param string $resultId
+     * @return ItemPreviewer
+     */
+    public function setResultId($resultId)
```
krampstudio: Should we set the results ? If the result id is used to retrieve the test taker and then the language, in this extension we should better set the language (which is computed at the place this is called/instantiated)

models/classes/ItemPreviewer.php
```diff
@@ -0,0 +1,279 @@
+<?php
```
krampstudio: Missing header

models/classes/ItemPreviewer.php
```diff
@@ -0,0 +1,279 @@
...
+        if (empty($this->itemHrefs)) {
+            $this->loadItemHrefs();
+        }
+
+        return $this->itemHrefs[2];
+    }
+
+    private function loadItemHrefs()
+    {
+        $runtimeService = $this->getServiceLocator()->get(RuntimeService::SERVICE_ID);
```
krampstudio: The link with the delivery needs to be validated in terms of architecture

scripts/install/RegisterTestRunnerPreviewers.php
```diff
@@ -0,0 +1,67 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
```
krampstudio: Wrong header (date)

views/js/previewer/provider/item.js
```diff
@@ -0,0 +1,388 @@
...
+        /**
+         * Initialize and load the test store
+         * @returns {testStore}
+         */
+        loadTestStore : function loadTestStore(){
+            var config = this.getConfig();
+
+            //the test run needs to be identified uniquely
+            var identifier = config.serviceCallId || 'test-' + Date.now();
+            return testStoreFactory(identifier);
```
krampstudio: do you need a test store for the item previewer ?

views/js/previewer/provider/item.js
```diff
@@ -0,0 +1,388 @@
...
+        loadPersistentStates : function loadPersistentStates() {
+            var self = this;
+            var config = this.getConfig();
+            var persistencePromise = cachedStore('test-states-' + config.serviceCallId, 'states');
+
+            persistencePromise.catch(function(err) {
+                self.trigger('error', err);
+            });
+
+            return persistencePromise
```
krampstudio: do you need persistent states also ?

views/js/previewer/config/item.js
```diff
@@ -0,0 +1,200 @@
...
+             */
+            getTimeout : function getTimeout() {
+                return storage.timeout;
+            },
+
+            /**
+             * Gets the config for the communication channel
+             * @returns {Object}
+             */
+            getCommunicationConfig : function getCommunicationConfig() {
```
krampstudio: Do we need a communication confifg for the previewer ?

views/js/previewer/provider/item.js
```diff
@@ -0,0 +1,388 @@
...
+                })
+                .on('enableitem', function () {
+                    this.trigger('enabletools');
+                })
+                .on('error', function () {
+                    this.trigger('disabletools enablenav');
+                })
+                .on('finish leave', function () {
+                    this.trigger('disablenav disabletools');
+                    this.trigger('endsession');
```
krampstudio: The 2 events `endsession` and `beforeunloaditem` are artifacts kept only for compatibility (as far as I remember in the test runner provider), not sure the item previewer should support them

views/js/previewer/proxy/item.js
```diff
@@ -0,0 +1,324 @@
...
+define([
+    'jquery',
+    'lodash',
+    'i18n',
+    'core/promise',
+    'core/promiseQueue',
+    'core/communicator',
+    'taoQtiTest/previewer/config/item',
+    'util/httpErrorParser'
+], function($, _, __, Promise, promiseQueue, communicatorFactory, configFactory, httpErrorParser) {
```
krampstudio: Shouldn't we have a very simple proxy that handles the only 2 requests we need ?

models/classes/ItemPreviewer.php
```diff
@@ -0,0 +1,279 @@
...
+
+namespace oat\taoQtiTest\models;
+
+use oat\taoDelivery\model\RuntimeService;
+use oat\taoItems\model\pack\ItemPack;
+use oat\taoItems\model\pack\Packer;
+use oat\taoQtiItem\model\qti\Service;
+use oat\taoQtiItem\model\QtiJsonItemCompiler;
+use oat\taoQtiTest\models\container\QtiTestDeliveryContainer;
+use oat\taoResultServer\models\classes\ResultServerService;
```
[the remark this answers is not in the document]
gyszucs: The `ResultServerService` is needed to get the result storage implementation to fetch the test taker. Is there any other way to do that? Would adding `taoResultServer` as a dependency  in manifest be an issue?

actions/class.Previewer.php
gyszucs: For the time being no, we can remove it.

models/classes/ItemPreviewer.php
gyszucs: You mean, setting a more generic name?

models/classes/ItemPreviewer.php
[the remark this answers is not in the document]
krampstudio: I'm agree, but shouldn't we limit in this extension to set the language using a setter or a method param ? So retrieving the test taker and then it's lang is done by the caller code (in the result extension)

models/classes/ItemPreviewer.php
gyszucs: Maybe, that can be a option but in that case the code of language computing will have to be moved into the controller which would make it more "crammed".

models/classes/ItemPreviewer.php
gyszucs: Sorry, what you mean?

models/classes/ItemPreviewer.php
krampstudio: yes if you have some inspiration (I don't). I can see multiple use case where you could want to load the item data alone.

models/classes/ItemPreviewer.php
krampstudio: The fact the implementation rely on the DeliveryContainer used. If @joel can have a look to check if it's ok

views/js/previewer/provider/item.js
jsconan: Roughly I would say no, but some plugins do need it. Especially the theme switcher, that is mandatory if wish to keep the same features level as for the previous previewer.

views/js/previewer/provider/item.js
jsconan: Some plugins may need it unfortunately

views/js/previewer/config/item.js
jsconan: Probably not. I kept it for compatibility reasons. I will test if I can get rid of it.

views/js/previewer/proxy/item.js
jsconan: Since the previewer is based on the test runner architecture and life cycle, I am not sure we can simplify the proxy so much

jsconan:
I moved the previewer to the extension taoQtiTestPreviewer and improved the link with taoOutcomeUi that is now made by setting a config (default item type to preview).

PR #1299, opened 2018-06-01
Fix/tao 6325/respect min constraint in choice interactions

uncleempty:


views/js/runner/helpers/currentItem.js
```diff
@@ -158,15 +206,20 @@ define([
...
+
+                _.forEach(declarations, function (declaration) {
                     var attributes = declaration.attributes || {};
                     var response = responses[attributes.identifier];
                     var baseType = attributes.baseType;
                     var cardinality = attributes.cardinality;
 
                     count++;
-                    if (!currentItemHelper.isQuestionAnswered(response, baseType, cardinality, declaration.defaultValue)) {
+                    if (!currentItemHelper.isQuestionAnswered(response, baseType, cardinality, declaration.defaultValue, constraintValues[attributes.identifier])) {
```
jsconan: Should not it be better to work on `qtiClass` instead?
Since the list of constraints is indexed by `qtiClass`, it should be more generic.
Otherwise what will happen next time when we will need to add another check? Add another parameter?

The purpose would be to:
- map the declaration with `qtiClass`
- pass the `qtiClass` to the `isQuestionAnswered` helper
- get the constraint from the index based on the provided `qtiClass`

views/js/runner/helpers/currentItem.js
```diff
@@ -158,15 +206,20 @@ define([
             var count = 0;
             var empty = 0;
 
+            var declarations;
+
             if (itemRunner) {
-                _.forEach(currentItemHelper.getDeclarations(runner), function (declaration) {
+                declarations = currentItemHelper.getDeclarations(runner);
+                var constraintValues = currentItemHelper.guessInteractionConstraintValues(runner);
```
jsconan: please declare var on top

PR #1302, opened 2018-06-07
Fix/tao 6360/progress counter

tikhanovichA:


hutnikau:
I tested it but progress is always `Item 0/0`:
image

PR #1309, opened 2018-06-13
Fix/tao 6382/sync timer with client

tikhanovichA:


config/default/testRunner.conf.php
```diff
@@ -555,5 +555,10 @@
     /**
      * Enables to run automatic navigation on items when timeLimits.minTime = timeLimits.maxTime
      */
-    'guidedNavigation' => false
+    'guidedNavigation' => false,
+
+    /**
+     * Sync the timer with client after reloading
+     */
+    'sync-timer-with-client' => true,
```
krampstudio: could you please find another name. There's no _sync_... `restore-timer-from-client` could be an option. 
Please comment the fact that it MUST be consistent with https://github.com/oat-sa/extension-tao-testqti/pull/1309/files#diff-5892ad10b1300de059d4e9fd1cfc2cc6L339. If the value is `client` then this new config must be `true`

PR #1315, opened 2018-06-19
Act 210 fix timers with extra time

hutnikau:
Steps to reproduce issue:
1. launch test with 10 seconds timer and with `extended_time=2` lti parameter. Total time should be 20 seconds.
2. When on timer will be less than 10 seconds (main time is up, and user uses extended time) go to the next item. On new item timer will show 20+ seconds again.

### DIFF
--- a/views/js/runner/plugins/controls/timer/plugin.js
+++ b/views/js/runner/plugins/controls/timer/plugin.js
@@ -58,20 +58,29 @@ define([
              * @param {Object} config - the current config, especially for the warnings
              * @return {Promise<Object[]>} the list of timers for the current context
              */
-            this.loadTimers = function loadTimers(timeStore, timeConstraints, isLinear,  config){
-                var timers = timersFactory(timeConstraints, isLinear, config);
+            this.loadTimers = function loadTimers(timeStore, testContext, config){
+                var isLinear =  !!testContext.isLinear;
+                var timeConstraints = testContext.timeConstraints;
+                var extraTime = testContext.extraTime;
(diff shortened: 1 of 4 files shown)


views/js/runner/plugins/controls/timer/plugin.js
```diff
@@ -58,20 +58,29 @@ define([
              * @param {Object} config - the current config, especially for the warnings
              * @return {Promise<Object[]>} the list of timers for the current context
              */
-            this.loadTimers = function loadTimers(timeStore, timeConstraints, isLinear,  config){
-                var timers = timersFactory(timeConstraints, isLinear, config);
+            this.loadTimers = function loadTimers(timeStore, testContext, config){
```
krampstudio: please don't give the whole testContext as parameter
```js
this.loadTimers = function loadTimers(imeStore, timeConstraints, isLinear, config, extraTime)
```
and please update the function documentation accordingly

views/js/runner/plugins/controls/timer/timers.js
```diff
@@ -85,9 +85,10 @@ define([
      * @param {Object} [config] - timers config
      * @param {Object[]} [config.warnings] - the warnings to apply to the timers (max only for now)
      * @param {Object[]} [config.warnings] - the warnings to apply to the timers (max only for now)
+     * @param {Object[]} extraTime - as defined in the testContext
```
krampstudio: Please update the documented type accordingly `{Object]` instead of `{Object[]}`

PR #1321, opened 2018-06-25
Feature/tao 6221 allow late submission

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-6221

Requires: 
 - [ ] https://github.com/oat-sa/extension-tao-test/pull/243

Handle the `allowLateSubmission` option in the TestRunner.

The implemented behavior is:

GIVEN that late allow late submission option is enabled
WHEN the timer expire
THEN the test taker can still interact with the item he is currently displaying for an unlimited amount of time
AND the item response is saved when the test taker submit his response
AND the test taker is not able to move to any other item of the timed section (section, test part or test)

### DIFF
--- a/models/classes/runner/QtiRunnerService.php
+++ b/models/classes/runner/QtiRunnerService.php
@@ -1083,10 +1083,11 @@ class QtiRunnerService extends ConfigurableService implements RunnerService
      * @param RunnerServiceContext $context
      * @param $scope
      * @param $ref
+     * @param $late
      * @return boolean
      * @throws \common_Exception
      */
-    public function timeout(RunnerServiceContext $context, $scope, $ref)
+    public function timeout(RunnerServiceContext $context, $scope, $ref, $late = false)
     {
         if ($context instanceof QtiRunnerServiceContext) {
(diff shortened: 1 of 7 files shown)


models/classes/runner/QtiRunnerService.php
```diff
@@ -1097,7 +1098,20 @@ public function timeout(RunnerServiceContext $context, $scope, $ref)
             }
             try {
                 $session->closeTimer($ref, $scope);
-                $session->checkTimeLimits(false, true, false);
+                if ($late) {
```
krampstudio: I am wondering how difficult it is to get the timeConstraint from the scope and ref to know if the timer has the parameter `allowLateSubmission`. If possible it would be better than relying on a parameter

jsconan: Yes, it could be possible, but it will come with some cost. And we cannot get rid of the parameter, as we we cannot simply check for the `allowLateSubmission` since the processing should not be done on each call but only when the `timeout` is actually done (i.e. when the TT submit the responses)

PR #1323, opened 2018-06-26
Act 210 store consumed timer value on client side

hutnikau:
We need to store on client side not remaining time but consumed time otherwise we also need to store which extra time values has been applied.
In Nccer there is ability to add/remove extra time not only during start of session but on the fly (by proctor).


krampstudio:
Even if the approach is correct changing the saved value breaks the compatibility, especially with ongoing tests.
I suggest you to keep saving the remainingTime and check if the `extraTime` value from the `testContext` has changed. It's a bit more of calculation but that's will keep the backward compat.

PR #1330, opened 2018-07-12
Fix/tao 6416/inject theme namespace on test init

uncleempty:
Depends on oat-sa/tao-core#1775 && oat-sa/extension-tao-delivery-rdf#208 && oat-sa/extension-tao-act#1037 && oat-sa/tao-core#1777 && oat-sa/extension-tao-itemqti#1172
Basically, the problem is that on non-LTI launches theme set on delivery edit form isn't applied to items.
To test it:
* create a delivery on a project which uses theme-switcher/has option to switch theme on delivery edit form;
* launch the delivery
* see if the item theme got applied


actions/class.Runner.php
```diff
@@ -306,6 +314,27 @@ public function init()
...
+//                $config["activeNamespace"] = $launchData->getVariable(LtiThemeSwitcher::LTI_VARIABLE);
+//            }
+//        }
+
+        /** @var \oat\taoDelivery\model\execution\DeliveryExecution $deliveryExecution */
+        $deliveryExecution = ServiceProxy::singleton()->getDeliveryExecution($executionUri);
+
+        $delivery = $deliveryExecution->getDelivery();
+
+        $themeNamespace = $delivery->getOnePropertyValue(new Property(DeliveryThemeDetailsProvider::DELIVERY_THEME_ID_URI));
```
krampstudio: @jbout don't we try to limit the access to the ontology during a test ?

actions/class.Runner.php
```diff
@@ -306,6 +314,27 @@ public function init()
         $this->returnJson($response, $code);
     }
 
+    private function guessThemeNamespace($executionUri)
```
krampstudio: Same problem, this method shouldn't belong to a controller

views/js/runner/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -74,6 +86,18 @@ define([
                 return themesConfig && _.size(themesConfig.available) > 1;
             }
 
+            function reloadThemes() {
+                var themeConfig = themeHandler.get('items');
```
krampstudio: Please don't use the same variable name than in a upper scope. Didn't you run ESLint  on your code (this should have triggered a warning).

actions/class.Runner.php
jbout: Correct, already discussed this with him, this should not be done here

models/classes/runner/config/QtiRunnerConfig.php
```diff
@@ -212,4 +224,33 @@ protected function getCategories(RunnerServiceContext $context)
...
+        $config = \common_ext_ExtensionsManager::singleton()->getExtensionById('taoTests')->getConfig('test_runner_plugin_registry');
+
+        if (array_key_exists(self::TOOL_ITEM_THEME_SWITCHER_KEY, $config) && $config[self::TOOL_ITEM_THEME_SWITCHER_KEY]['active']) {
+            return true;
+        }
+    }
+
+    private function guessTestTheme()
+    {
+        $request = \Context::getInstance()->getRequest();
```
krampstudio: You should never access a request (controller stuff) within a service. 
A service needs to be stateless and I/O oriented.

krampstudio: Sorry I didn't see the `buildConfig` doesn't have access to the session/context.
 -  use the `options` object to send the active theme namespace, see https://github.com/oat-sa/extension-tao-testqti/pull/1330/files#diff-b0de81bf25f20349c0102da237a83ac7L160

krampstudio: And please check if the backend unit tests needs to be update also

PR #1336, opened 2018-07-20
Fix/tao 6556 pause button not working

bugalot:
Please see TAO-6556 for the reason of this :) !


jbout:
Problematic due to direct check of extension

PR #1343, opened 2018-08-03
Feature/tao 6573 config based category presets

krampstudio:
**Related to :** 
 - https://oat-sa.atlassian.net/browse/TAO-6573
 - https://oat-sa.atlassian.net/browse/TAO-6355 

### Feature 

The list of "Tools" in the test creator links to categories are generated by the service `\oat\taoQtiTest\models\TestCategoryPresetProvider`. I've added a new method `getAvailablePresets` (in order to keep backward compat) that let's you retrieve the list of category presets but, this time, based on a configuration.
This let's you define for each preset the linked feature flag. If a preset has the field `featureFlag => "enable-foo"`, the preset will be available only if the configuration contains the flag `"enable-foo" => true`.

So in order to achieve our goal, I've added the field  `'featureFlag' => 'enable-allow-skipping'` to the `Section skipping` preset, so if the configuration contains this flag and this flag is `false` we filter out the preset. 

I've added this mechanism since there are other presets that could be linked to the configuration.

### How to test

 - Run the unit test : `taoQtiTest/test/TestCategoryPresetProviderTest.php`
 - Change the value of `enable-allow-skipping` in `config/taoQtiTest/testRunner.conf.php`
 - Check within the _test creator_  :arrow_right:  _section  properties_ :arrow_right: _Test Navigation_ :  the `Section skipping` option should displayed based on the value of the config.

### DIFF
--- a/models/classes/TestCategoryPreset.php
+++ b/models/classes/TestCategoryPreset.php
@@ -59,6 +59,12 @@ class TestCategoryPreset implements JsonSerializable
      */
     private $pluginId = '';
 
+    /**
+     * @var string $featureFlag - the name of a config flag,
+     * the preset will be deactivated based on this optional value.
+     */
+    private $featureFlag;
+
 
     /**
      * Create a test category preset
@@ -93,6 +99,9 @@ class TestCategoryPreset implements JsonSerializable
         if(isset($data['pluginId'])) {
             $this->pluginId = (string) $data['pluginId'];
         }
+        if(isset($data['featureFlag'])) {
+            $this->featureFlag = (string) $data['featureFlag'];
+        }
     }
 
     public function getId()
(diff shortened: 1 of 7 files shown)


models/classes/TestCategoryPreset.php
```diff
@@ -125,6 +134,11 @@ public function getPluginId()
         return $this->pluginId;
     }
 
+    public function getFeatureFlag()
+    {
+        return isset($this->featureFlag) ? $this->featureFlag : false;
```
gyszucs: No really need for `isset` here. Just return the property and make sure the return value is a string. Returning a bool can be confusing, because the initial value is a string. So, for example: `return (string) $this->featureFlag;`

models/classes/TestCategoryPresetProvider.php
```diff
@@ -86,6 +87,65 @@ public function getPresets() {
...
+     *
+     * @param TestCategoryPreset $preset the preset to test 
+     * @param array $config the configuration
+     * @return boolean true if available
+     */
+    private function isPresetAvailable(TestCategoryPreset $preset, $config = [])
+    {
+        if(!is_null($preset)) {
+            $flag = $preset->getFeatureFlag();
+            if(is_string($flag) && isset($config[$flag]) && $config[$flag] != true){
```
gyszucs: Instead of `is_string($flag)` you can use only `$flag` to check whether it has a value or not.

PR #1350, opened 2018-08-14
Fix/tao 6651 remove double spaces

siwane:


locales/ja-JP/messages.lang.php
```diff
@@ -277,7 +268,7 @@
 $GLOBALS['__l10n']["Loaded"]="読み込み完了";
 $GLOBALS['__l10n']["Loaded at Startup"]="起動時に実行";
 $GLOBALS['__l10n']["Loading"]="読み込み中";
-$GLOBALS['__l10n']["Log in"]="ログイン";
+$GLOBALS['__l10n']["Log in"]="ログインID";
```
ionutpad: 'ID' has been added any reason?

locales/it-IT/messages.po
```diff
@@ -422,8 +452,8 @@ msgstr "Se richiesto, deve apparire almeno una volta nella selezione"
 msgid "If set, it causes the order of the child elements to be randomized, otherwise it uses the order in which the child elements are defined."
 msgstr "Se impostato, provoca l'ordine degli elementi figlio da randomizzare, altrimenti utilizza l'ordine in cui sono definiti gli elementi figlio."
 
-msgid "If the candidate can skip the item, without submitting a response."
-msgstr "Se il candidato può ignorare l'item, senza presentare una risposta."
+msgid "If the candidate can skip the item, without submitting a response (default is true)."
+msgstr ""
```
ionutpad: Italian translation disappeared.

locales/it-IT/messages.po
```diff
@@ -1181,7 +1289,7 @@ msgstr ""
 msgid "You are about to submit the test. You will not be able to access this test once submitted. Click OK to continue and submit the test."
 msgstr "Hai terminato questa parte della prova. Se clicchi su Sì non potrai più tornare a questa sezione."
 
-msgid "You are encountering a prolonged connectivity loss"
+msgid "You are encountering a prolonged connectivity loss. "
```
ionutpad: Extra space added after point `.`

locales/ja-JP/messages.lang.php
siwane: yes, because of https://github.com/oat-sa/tao-core/blob/master/locales/ja-JP/messages.po#L653. I'm not able to say if it's an error or not

locales/it-IT/messages.po
siwane: nope, it's not the same `(default is true).` Need to be retranslated in all languages

PR #1358, opened 2018-08-30
Fix/tao 6288 revert timer label change on install

no-chris:
@ionutpad reverted the change in the update script (https://github.com/oat-sa/extension-tao-testqti/pull/1346) but forgot to do it at install. 

Might be worth checking if this impact the customer that requested that functionnality, probably the defaut config with the timer label configuration needs to be copied to the NCSBN extension? Otherwise they will have a regression on re-install.


uncleempty:
Haven't seen any issues with timer on fresh NCSBN installation. Works like it should.

PR #1362, opened 2018-09-07
Feature/si 144/qti cli exporter

zagovorichev:


zagovorichev:
@Alroniks This is just a console QTI - test exporter (as a default but will be exported to the servers folder)

zagovorichev:
@tikhanovichA Update

PR #1371, opened 2018-10-11
Feature/tao 6908/item metadata compiled

tikhanovichA:
Depends on https://github.com/oat-sa/extension-tao-itemqti/pull/1191


manifest.php
```diff
@@ -38,10 +38,10 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '27.0.7',
+    'version'     => '27.1.0',
```
ionutpad: We will need big bum I think, interface changed.

scripts/tools/RecompileItemsElements.php
```diff
@@ -0,0 +1,217 @@
...
+     */
+    protected function recompileItems()
+    {
+        $extManager = $this->getServiceLocator()->get(\common_ext_ExtensionsManager::SERVICE_ID);
+
+        if ($extManager->isInstalled('taoDeliveryRdf') === true && $extManager->isEnabled('taoDeliveryRdf') === true) {
+            $compiledDeliveryClass = $this->getClass(DeliveryAssemblyService::CLASS_URI);
+            if ($compiledDeliveryClass->exists() === true) {
+                foreach ($compiledDeliveryClass->getInstances(true) as $compiledDelivery) {
+                    $this->getAssessmentsFromDelivery($compiledDelivery);
```
ionutpad: You will have to add a catch of exception here in case the delivery it's not valid compile to not stop the execution. I get this exception in the described
```
An exception occured while running "\oat\taoQtiTest\scripts\tools\RecompileItemsElements"
  Property (http://www.tao.lu/Ontologies/TAODelivery.rdf#AssembledDeliveryRuntime) of resource (https://nccersso.taocloud.org/nccer.rdf#i15393348749718555) should not be empty

```

PR #1372, opened 2018-10-12
Fix/tao 6943/qtirunner popup ordering

zagovorichev:


krampstudio:
Looks ok as a quick fix. please ensure the callback isn't trigger at both events (since `nav-next` triggers the `move`)

zagovorichev:
@krampstudio Callback isn't triggered twice, because without success we can't go further and when we can pass first, then the second can be passed, as well. I mean that we have the same condition for both of the triggers

PR #1382, opened 2018-11-05
remove theme ID fetching from generation of Runner config

peetya:


models/classes/runner/config/QtiRunnerConfig.php
```diff
@@ -38,7 +38,14 @@ class QtiRunnerConfig extends ConfigurableService implements RunnerConfig
 
     const OPTION_CONFIG = 'config';
 
+    /**
```
alexeyvy: No usages within oat-sa repos, maybe we just remove it in this version?

https://github.com/search?q=org%3Aoat-sa+TOOL_ITEM_THEME_SWITCHER&type=Code

models/classes/runner/config/QtiRunnerConfig.php
```diff
@@ -38,7 +38,14 @@ class QtiRunnerConfig extends ConfigurableService implements RunnerConfig
...
+     * @deprecated since version 29.2.0, to be removed in 30.0.0. Use QtiRunnerService::TOOL_ITEM_THEME_SWITCHER instead
+     */
     const TOOL_ITEM_THEME_SWITCHER = 'itemThemeSwitcher';
+
+    /**
```
alexeyvy: Same thing: https://github.com/search?q=org%3Aoat-sa+TOOL_ITEM_THEME_SWITCHER_KEY&type=Code

models/classes/runner/config/QtiRunnerConfig.php
peetya: Community bundles could use it so it's more safe to mark it as deprecated.

models/classes/runner/config/QtiRunnerConfig.php
peetya: Same comment ;) So we cannot be sure that no other community extension is using that so it's more safe to mark it as deprecated.

peetya:
I added some unit tests, which covers the updated part of the class.

PR #1385, opened 2018-11-27
Allow runner to load extra routes defined by customer extensions

oatymart:


manifest.php
```diff
@@ -38,7 +38,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '29.2.0',
+    'version'     => '29.2.1',
```
jsconan: Is it not a feature?

PR #1379, opened 2018-10-29
Backend for preserving runners tools data

alexeyvy:

PR #1389, opened 2018-11-29
Breaking/tao 7156 load tool state

krampstudio:
Related to : 
 - https://oat-sa.atlassian.net/browse/TAO-7156
 - https://oat-sa.atlassian.net/browse/TAO-6748 

Requires : 
 - [x] #1379 (test in an integration branch if needed
 - [x] https://github.com/oat-sa/extension-tao-test/pull/264 

Changes : 
 - refactor the highlighter plugin to be able to store it's value in the testStore and to be able to use the toolState by extension
 - added a new module `provider/toolStateBridge` to handle setting and restoring the states of the tools (a tool is a plugin, but each plugin isn't a tool; with `toolName == pluginName`) from/to the `testStore`.
- implementation of the `toolStateBridge` directly in the qti provider

How to test : 
 - configure a tool (the highlighter to match the AC) in `config/taoQtiTest/testRunner.conf.php`
 - Install the storage, Rds for example `sudo -u www-data php index.php "oat\taoQtiTest\scripts\install\InstallRdsToolsStateStorage"`
 - Define the storage as RDS in `config/taoQtiTest/toolStateStorage.conf.php`
 - take a test with the highlighter configured
 - check the highlighting state is kept on the same execution when you reload the page AND resume the test with another browser


krampstudio:
@zagovorichev The dependencies are merged, could you please have a look

krampstudio:
@zagovorichev could you please review it again

PR #1398, opened 2018-12-13
Feature/tao 7239/allow skip ahead with review panel

uncleempty:


models/classes/QtiCategoryPresetProvider.php
```diff
@@ -205,6 +188,69 @@ public function registerPresets(TestCategoryPresetProvider $presetService)
...
             ]
         );
+
+        //TAO-7239, register new feature category only if it is enabled in config
+        /** @var \common_ext_ExtensionsManager $tt */
+        $extManager = $presetService->getServiceLocator()->get(\common_ext_ExtensionsManager::SERVICE_ID);
+        $testRunnerConfig = $extManager->getExtensionById('taoQtiTest')->getConfig('testRunner');
+
+        $isSkipaheadEnabled = isset($testRunnerConfig['test-taker-review-skipahead']) && $testRunnerConfig['test-taker-review-skipahead'];
+        if ($isSkipaheadEnabled) {
```
jsconan: Having a new option to disclose the feature, why not, but do you really need to duplicate the other options? Is it for the sort order?
It should be better to avoid such duplicate, especially since this introduces some mess and is error prone.
Why not something like:
```php
        $groupNavigation = [
            TestCategoryPreset::fromArray([
                'id'            => 'reviewScreen',
                'label'         => __('Enable Review Screen'),
                'qtiCategory'   => 'x-tao-option-reviewScreen',
                'description'   => __('Enable the item review screen / navigator.'),
                'order'         => 100,
                'pluginId'      => 'review'
            ]),
            TestCategoryPreset::fromArray([
                'id'            => 'markReview',
                'label'         => __('Enable Mark for Review'),
                'qtiCategory'   => 'x-tao-option-markReview',
                'description'   => __('Enable mark for review of items. Requires the Review Screen option.'),
                'order'         => 200,
                'pluginId'      => 'review'
            ]),
            TestCategoryPreset::fromArray([
                'id'            => 'informational',
                'label'         => __('Informational Item Usage'),
                'qtiCategory'   => 'x-tao-itemusage-informational',
                'description'   => __('Force the item to be considered as informational and not taken into account in (un)answered / flagged counters.'),
                'order'         => 300,
                'pluginId'      => 'review'
            ]),
            TestCategoryPreset::fromArray([
                'id'            => 'nextSection',
                'label'         => __('Allow Section Skipping'),
                'qtiCategory'   => 'x-tao-option-nextSection',
                'description'   => __('Allow skipping of the current section.'),
                'order'         => 400,
                'pluginId'      => 'nextSection',
                'featureFlag'   => 'next-section'
            ])
        ];
        if ($isSkipaheadEnabled) {
            $groupNavigation[] = TestCategoryPreset::fromArray([
                'id'            => 'skipAhead',
                'label'         => __('Enable Skipping Ahead'),
                'qtiCategory'   => 'x-tao-option-review-skipahead',
                'description'   => __('Enables skipping to items within this section. Requires the review screen option.'),
                'order'         => 250,
                'pluginId'      => 'review',
                'featureFlag'   => 'skip-ahead'
        }
        $presetService->register(
            TestCategoryPresetProvider::GROUP_NAVIGATION,
            $groupNavigation
        );
```

views/js/runner/plugins/navigation/review/review.js
```diff
@@ -215,6 +215,25 @@ define([
...
+                var reviewConfig = testConfig.review || {};
+
+                var isEnabled = (reviewConfig.allowSkipahead || false ) && (options.reviewSkipahead || false);
+
+                return isEnabled;
+            }
+
+            if (isSkipaheadEnabled()) {
+                navigatorConfig.skipaheadEnabled = true;
+                navigatorConfig.scope = 'testPart';
```
jsconan: Why do you force the scope to `testPart`?

views/js/runner/helpers/map.js
```diff
@@ -302,8 +303,17 @@ define([
             }
 
             // update the stats to reflect the scope
-            if (section) {
-                section.stats = this.computeItemStats(section.items);
+            if (context && context.options && context.options.reviewSkipahead) {
+                partSections = part.sections || {};
+                if (partSections) {
+                    _.each(partSections, function(partSection) {
```
jsconan: `_.each` is depreacted, please use `_.forEach` instead.

views/js/runner/helpers/map.js
```diff
@@ -436,6 +446,9 @@ define([
                 if (item.viewed) {
                     acc.viewed++;
                 }
+                if (item.categories && item.categories.includes('x-tao-option-review-skipahead')) {
+                    item.canBeSkipped = true;
```
jsconan: is it really needed? the `prevent-unseen` option already covers that.

views/js/runner/plugins/navigation/review/navigator.js
```diff
@@ -301,10 +305,15 @@ define([
...
 
+                this.setState('skipahead-enabled', this.config.skipaheadEnabled);
                 this.setState('prevents-unseen', this.config.preventsUnseen);
-                if (this.config.preventsUnseen) {
-                    // disables all unseen items to prevent the test taker has access to.
-                    this.controls.$tree.find(_selectors.unseen).addClass(_cssCls.disabled);
+                if (!this.config.skipaheadEnabled) {
+                    this.controls.$tree.find(_selectors.skippable).removeClass(_cssCls.skippable);
+
+                    if (this.config.preventsUnseen) {
```
jsconan: is it really needed? the `prevent-unseen` option already covers that. All you have to do is: `if (this.config.preventsUnseen && !options.reviewSkipahead) {`

views/js/runner/helpers/map.js
uncleempty: Hmmm? I can change this, but can you point me on the deprecation info?
Because neither at [Lodash deprecations](https://github.com/lodash/lodash/wiki/Deprecations), nor at [Lodash changelog](https://github.com/lodash/lodash/wiki/Changelog) I see it deprecated. Just only found that `each` is an alias for `forEach`. Quick googling also didn't gave any results for that.

jsconan: it is an alias now, the exact API is `forEach`: https://lodash.com/docs/2.4.2#forEach

So, yes, please don't use the alias as it may disappear in later versions

jsconan: But, anyway, this part of the code should go IMO. The `prevent-unseen` option already does the job. The only thing you have to do is add another condition to allow the disclose of unseen item for a particular category.

views/js/runner/plugins/navigation/review/navigator.js
uncleempty: I'm not sure where to answer better about scopes and "similarity" to `prevents-unseen`, so would just comment this one.
First of all `skipahead` is not so similar to `prevents-unseen` as it may seem. `prevents-unseen` will allow to navigate to any item not yet seen. And `skipahead` should allow navigating to the items in sections which has `Enable Skipping Ahead` mark checked in test creator. So, for example, when seeing 4 sections with only 3 of them allowed to be skipped ahead: a) `prevents-unseen` will allow to go to any item of any section, while b) `skipahead` will allow to navigate to any item of only those sections who are allowed to be skipped ahead.

That's why, in general, I decided to not rely on the existing option, but to move it to separate one, even if the functionality partially covered by existing one.

Then, about forcing scope to `testPart`, this one was discussed with Mark/Tom. As `skipahead` is fully a, some kind of, "debug" feature it should give a little more convenience to work with review panel (and because most customers has scope of review panel set to `testSection` which almost fully cancel the new feature use - in this case the one needs this, he could only use `prevents-unseen`).

So, these are the main reasons, to move it all to a different feature, to allow this working separately. 
p.s.
Also, as said, `skipahead` should work in any case it is enabled, no matter what `prevents-unseen` is set to.

views/js/runner/helpers/map.js
uncleempty: Aw, yes, understood about using alias, will move it to use the `forEach`.

About `prevents-unseen` - please, see the comment to the last one of yours in here

models/classes/QtiCategoryPresetProvider.php
uncleempty: Indeed, should do the job, will change this

views/js/runner/plugins/navigation/review/navigator.js
jsconan: I see your point. However, I disagree with the scope change. Even if it is a debug feature, it will break somehow the main behavior. What happen if the customer has set the scope to the whole test? suddenly it will be restricted to a `testPart`. Having customer specifics disturbing the main component is not acceptable.

Regarding the `skipahead` option, the AC of the ticket clearly mention as long as a section (so each items of the section) has the option, the user can reach any of the other items of this section. This is the purpose of the `prevent-unseen` too (in reverse). No need to specify which item can be reached. The feature is more to define from which item the "move-to-any" is allowed. Or did I misread the ticket?

views/js/runner/plugins/navigation/review/navigator.js
```diff
@@ -281,8 +281,9 @@ define([
                 total: 0
             };
             var totalQuestions = this.getProgressionTotal(progression, 'questions');
+            var activeSection, isSkipaheadEnabled;
 
-            this.map = map;
+                this.map = map;
```
jsconan: typo?

views/js/runner/helpers/map.js
```diff
@@ -276,6 +303,7 @@ define([
             var scopeMap = _.cloneDeep(map || {});
             var part;
             var section;
+            var partSections;
```
jsconan: is this variable still used?

views/js/runner/plugins/navigation/review/navigator.js
```diff
@@ -341,7 +348,7 @@ define([
             item.active = true;
 
             // adjust each item with additional meta
-            return mapHelper.each(scopedMap, function(itm) {
+            return mapHelper.each(scopedMap, function(itm, scopedSection) {
```
jsconan: is the parameter used?

PR #1401, opened 2018-12-18
Possible fix for the problem with wrong timestamps for current timed …

uncleempty:
~~Now the `firstTimestamp` and `lastTimestamp` values should be more accurate as they are taken from the timeline by the identifier of current `source`

**Not to be merged yet! Just a possible solution to share**~~

Can be merged.

Code removal was needed - it was unneeded in case of `client` timer as the timer values come from browser, and also unneeded in case of `server` timer - values are re-calculated in another place

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -38,7 +38,7 @@ return array(
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '29.7.1',
+    'version'     => '29.7.2',
     'author'      => 'Open Assessment Technologies',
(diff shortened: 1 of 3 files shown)


krampstudio:
Related to https://oat-sa.atlassian.net/browse/TAO-7295 https://oat-sa.atlassian.net/browse/TAO-7192 and https://oat-sa.atlassian.net/browse/TAO-7327

SashaTAO:
Could you please specify JIRA ticket id in the branch name?

krampstudio:
@uncleempty could you please redo the PR with the correct branch name. This fix is "highest priority"

PR #1406, opened 2018-12-31
Feature/tao 7016 export offline actions

oatymart:
A part of the work for https://oat-sa.atlassian.net/browse/TAO-7016

Adds a method to export a copy of the actions from the actionStore so they can be downloaded in a file by the "disconnectedTestSaver" TR plugin.


views/js/runner/proxy/cache/proxy.js
```diff
@@ -389,6 +389,29 @@ define([
...
+                var actions;
+                return this.queue.serie(function(){
+                    return self.actiontStore.flush().then(function(data){
+                        actions = data;
+                        _.forEach(actions, function (action) {
+                            self.actiontStore.push(action.action, action.parameters);
+                        });
+                        return data;
+                    })
+                    .catch(function(err){
```
krampstudio: this is not necessary to `catch` and `throw` since you already return the Promise

oatymart: Ok, removed.

PR #1407, opened 2018-12-31
Fix/duration controls wrapping

oatymart:
Fix for a TAO-wide visual issue in the authoring sidebar:
The `h:m:s` fields were always wrapping onto a second line.

duration-before
duration-after-fix

I've used a `flex-wrap: nowrap` container to correct the bug.


views/scss/creator.scss
```diff
@@ -98,6 +98,12 @@
         @include responsive('width', $sideWidthSmall, $sideWidthMedium, $sideWidthWide);
         @include vendor-prefix(flex, 0 1 auto, property, (-ms-, -webkit-, ''));
         height: 100%;
+
+        .duration-group {
+            display: flex;
```
krampstudio: shouldn't you use the prefixed version for IE 11?

PR #1419, opened 2019-01-11
Covered actions with unit tests

Skyld:
Scope:

- Covered all actions with unit tests
- Added phpunit xml dist file to help running tests
- Added `var` folder to gitignore for temporary files such as phpunit coverage report


phpunit.xml.dist
```diff
@@ -0,0 +1,17 @@
+<phpunit bootstrap="vendor/autoload.php">
```
siwane: @jbout What do u think adding this file on each extensions? as @Skyld said: ```the xml itself is in gitignore,
the dist file serves as a bluprint for the configuration with the minimal settings```. Can also include report for extension.

PR #1420, opened 2019-01-14
trigger event when emptyitemwarningmodal enabled

peetya:


views/js/controller/runtime/testRunner.js
```diff
@@ -1216,13 +1216,18 @@ function (
         };
 
         var config = module.config();
+        var _isEmptyItemWarningModalEnabled = function() {
+            return config
+                && config.qtiTools
+                && config.qtiTools.emptyItemWarningModal;
+        };
```
jsconan: Is it a check for an external component? Not sure this is the better way. It introduces coupling, which is bad.
A more generic way should be possible.

views/js/controller/runtime/testRunner.js
```diff
@@ -1307,7 +1312,17 @@ function (
 
                 $controls.$forwardButtons.click(function () {
                     if (!$(this).hasClass('disabled')) {
-                        TestRunner.moveForward();
+                        if (_isEmptyItemWarningModalEnabled()) {
+                            $doc
+                                .off('testRunner:moveForward')
+                                .on('testRunner:moveForward', function() {
```
jsconan: I am afraid the event name does not comply to our coding rules. Namespace separator is the dot `.`.  And usually we tend to use dashes as word separator, all in lowercase: `testrunner-moveforward`.

peetya:
Closing the PR as it's not related anymore.

PR #1427, opened 2019-01-30
Feature/tao 7419 next item warning plugin

oatymart:
Plugin for https://oat-sa.atlassian.net/browse/TAO-7419

This plugin depends on two platform-level Test Runner configuration options: `force-enable-next-item-warning` and `enable-next-item-warning-checkbox`.

It can be tested in its 3 operating modes ("always on", "always off", "on with disable checkbox") by setting the appropriate Boolean values in `config/taoQtiTest/testRunner.conf.php`

~~The plugin listens for some new events: `nav-next`, `nav-skip` and `nav-nextsection`. Some refactoring of the skip.js and nextSection.js plugins was necessary to match the structure of the next.js plugin more closely.~~ It now listens only to `move` and `skip`, without modifying any existing plugins.

Note that the plugin will be registered *inactive* for now in plain TAO. It will be activated and configured for the specific customer.


config/default/testRunner.conf.php
```diff
@@ -575,4 +575,12 @@
      * Specifies runner tools that should keep its states in backend storage
      */
     'tool-state-server-storage' => [],
+
+    /**
+     * Defines whether to always show a warning dialog before moving to next item
+     * Also, if shown, should the dialog contain a "don't show again" checkbox
```
krampstudio: Please indicate this option affects only the `Linear` test parts.

config/default/testRunner.conf.php
```diff
@@ -575,4 +575,12 @@
...
+     */
+    'force-enable-next-item-warning' => false,
+    'enable-next-item-warning-checkbox' => true,
```
krampstudio: It looks obvious for us what this option is supposed to do, at least add a comment describing what the checkbox is supposed to do and this option makes sense only when the previous is activated

views/js/runner/plugins/navigation/next/dialogConfirmNext.js
```diff
@@ -0,0 +1,104 @@
...
+                label : __('Cancel'),
+                close: true
+            },
+            {
+                id : 'ok',
+                type : 'info',
+                label : __('Go to next item'),
+                close: true
+            }],
+            onOkBtn: function() {
```
krampstudio: by convention is ES5 code, we use _named function expression_ : 
```js
onOkBtn: function onOkBtn () {
```

views/js/runner/plugins/navigation/next/nextItemWarning.js
```diff
@@ -0,0 +1,155 @@
...
+
+            //plugin behavior
+            /**
+             * Checks configuration, shows a dialog asking to confirm the nav action
+             *
+             * @param {String} action - 'next' or 'skip'
+             * @returns {Promise} - resolves if dialog accepted or not shown, rejects if dialog cancelled
+             */
+            function doNextWarning(action) {
+                //var context = testRunner.getTestContext();
```
krampstudio: could you please remove any comment

views/js/runner/plugins/navigation/next/nextItemWarning.js
```diff
@@ -0,0 +1,155 @@
...
+                var checkboxParams = null;
+
+                testRunner.trigger('disablenav');
+
+                return new Promise(function(resolve, reject) {
```
krampstudio: Please chain promises : 
```js
return testStore.getStore(self.getName()).then(function(store) {
    return store.getItem('dontShowNextItemWarning').then(function(checkboxValue) {
        // Show the warning unless user has turned it off:
        if (checkboxValue !== true) {
            // Define checkbox only if enabled by config:
            if (testConfig.enableNextItemWarningCheckbox) {
                checkboxParams = {
                    checked: checkboxValue,
                    submitChecked: function() {
                        store.setItem('dontShowNextItemWarning', true);
                    },
                    submitUnchecked: function() {
                        store.setItem('dontShowNextItemWarning', false);
                    },
                };
                // use a wrapping promise only when necessary, since you're already in a chain,
                // 
                return new Promise(function(resolve, reject) {
                    dialogConfirmNext(
                        __('Go to the next item?'),
                        getCustomNextMessage(action),
                        resolve, // if the test taker accepts
                        function cancel() {
                           //this is an internal convention to differentiate errors, and rejection that cancel an aciton
                            reject({
                                cancel: true
                            }); 
                        },
                        checkboxParams
                    );
                });
            }
        }
    });
})
.catch(function(err) {
    //if the rejection is due to an error, please rethrow it
    if (err && err instanceof Error) {
        throw err;
    }
    if (err && err.cancel === true) {
        testRunner.trigger('enablenav');
    }
});

```

views/js/runner/plugins/navigation/next/nextItemWarning.js
```diff
@@ -0,0 +1,155 @@
...
+                    testStore.getStore(self.getName()).then(function(store) {
+                        store.setItem('dontShowNextItemWarning', null);
+                    });
+                })
+                .before('move skip', function(e, type, scope) {
+                    var context = testRunner.getTestContext();
+                    if (context.isLinear) {
+                        // Do nothing if nextSection warning imminent:
+                        if (scope === 'section' && context.options.nextSectionWarning) {
+                            return Promise.resolve();
```
krampstudio: this is not necessary to return a Promise if you don't need to block the event. You should handle only the cases where you'll block the event

views/js/runner/plugins/navigation/next/nextItemWarning.js
```diff
@@ -0,0 +1,155 @@
...
+                if (! itemPartiallyAnswered) {
+                    customNextMessage = __('Are you sure you want to go to the next item? You will not be able to go back and provide an answer.');
+                }
+                else if (action === 'next') {
+                    customNextMessage = __('Are you sure you want to go to the next item? You will not be able to go back and change your answer.');
+                }
+                else if (action === 'skip') {
+                    customNextMessage = __('Are you sure you want to clear your answer and go to the next item? You will not be able to go back and provide an answer.');
+                }
+                return customNextMessage;
```
krampstudio: shouldn't we have a default value ? or if `action` is required, please throw a `TypeError` if the function is called with incorrect arguments

views/js/runner/plugins/navigation/next/nextItemWarning.js
```diff
@@ -0,0 +1,155 @@
...
+    'taoQtiTest/runner/helpers/currentItem',
+    'taoQtiTest/runner/plugins/navigation/next/dialogConfirmNext'
+], function ($, _, __, pluginFactory, currentItemHelper, dialogConfirmNext){
+    'use strict';
+
+    /**
+     * Returns the configured plugin
+     */
+    return pluginFactory({
+        name : 'nextItemWarning',
```
krampstudio: I'm thinking if we shouldn't rename it to `linearNextItemWarning` so people doesn't get mistaken (with the other plugins that handle warnings)

oatymart: I've gone ahead with the name change to `linearNextItemWarning` in all appropriate places.

views/js/runner/plugins/navigation/next/nextItemWarning.js
oatymart: Implemented following this model.

NB: I am still returning a rejected promise after catching the user's rejection, otherwise the move takes place (which is not desired).

manifest.php
```diff
@@ -38,12 +39,12 @@
...
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '30.4.0',
+    'version'     => '30.5.0',
     'author'      => 'Open Assessment Technologies',
     'requires'    => array(
         'taoQtiItem' => '>=18.0.0',
         'taoTests'   => '>=8.3.0',
-        'tao'        => '>=24.0.0',
+        'tao'        => '>=24.2.0',
```
krampstudio: the version isn't correct

oatymart: Fixed now.

PR #1434, opened 2019-02-19
Feature/tao 7703/scientific calculator tweaks

ricproenca-tao:
related to https://oat-sa.atlassian.net/browse/TAO-7703

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -39,7 +39,7 @@ return array(
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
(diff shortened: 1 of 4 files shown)


manifest.php
```diff
@@ -39,7 +39,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '30.5.3',
+    'version'     => '30.6.3',
```
jsconan: Caution, the version increase should produce an aligned number: `30.5.3` => `30.6.0`

PR #1435, opened 2019-02-20
Fix/tao 7714 informational item nowarning

oatymart:
https://oat-sa.atlassian.net/browse/TAO-7714

Added an if-statement to bypass the linearNextItemWarning dialog if the item is informational.
This works with the 'Informational Item Usage' choice in Test authoring. The Test Part must be in linear mode.

As a reminder, this TR plugin is inactive by default (activated only for Infosign), so you may need to activate it in `config/taoTests/test_runner_plugin_registry.conf.php` and set the `force-enable-linear-next-item-warning` setting in `config/taoQtiTest/testRunner.conf.php`

### DIFF
--- a/views/js/test/runner/plugins/navigation/next/linearNextItemWarning/test.js
+++ b/views/js/test/runner/plugins/navigation/next/linearNextItemWarning/test.js
@@ -29,12 +29,6 @@ define([
     var providerName = 'mock';
     runnerFactory.registerProvider(providerName, providerMock());
 
-    // mock the item we will get (informational or not)
-    mapHelper.getItemAt = function(map, isInfo) {
-        return {
-            informational: isInfo
-        };
-    };
 
     /**
      * The following tests applies to all plugins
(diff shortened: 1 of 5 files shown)


views/js/test/runner/plugins/navigation/next/linearNextItemWarning/test.js
```diff
@@ -0,0 +1,305 @@
...
+    'taoQtiTest/runner/helpers/map'
+], function(Promise, runnerFactory, providerMock, pluginFactory, mapHelper) {
+    'use strict';
+
+    var pluginApi;
+    var providerName = 'mock';
+    runnerFactory.registerProvider(providerName, providerMock());
+
+    // mock the item we will get (informational or not)
+    mapHelper.getItemAt = function(map, isInfo) {
```
krampstudio: I had difficulties to figure out you were using the `position` property from the `testContext` in order to flag the item as informational. 
I'd suggest to make this part more explicit, for example : 

```js
QUnit.cases([
        {
            title: 'when the next part warning is set',
            context: {
                enableAllowSkipping: false,
                allowSkipping: false,
                options: {
                    nextPartWarning: true,
                    nextSectionWarning: false
                }
            },
            item : {
               informational : false
            }
        },
        {
            title: 'when the item is informational',
            context: {
                isLinear: true,
                options: {
                    nextPartWarning: false,
                    nextSectionWarning: false
                }
            },
            item : {
               informational : false
            }
    ])
    .asyncTest('No dialog is triggered ', function(caseData, assert) {
        var runner = runnerFactory(providerName);
        var plugin = pluginFactory(runner, runner.getAreaBroker());

        // mock test store init
        runner.getTestStore = function() {
            return {
                setVolatile: function() {}
            };
        };
        // mock config
        runner.getTestData = function() {
            return {
                config: caseData.testConfig
            };
        };
        // mock the item we will get (informational or not)
        mapHelper.getItemAt = function(map) {
           return caseData.item;
        };

        QUnit.expect(1);

        //...
    });
```

views/js/test/runner/plugins/navigation/next/linearNextItemWarning/test.js
```diff
@@ -0,0 +1,305 @@
...
+
+    /**
+     * Specific tests for this plugin
+     */
+    QUnit.module('Behavior');
+
+    QUnit.cases([
+        {
+            title: 'when the next part warning is set',
+            context: {
```
krampstudio: would you mind rename it `testContext` so we know where it will be used

views/js/test/runner/plugins/navigation/next/linearNextItemWarning/test.js
oatymart: Thanks, I didn't spot that way of doing it. Much clearer now.

PR #1436, opened 2019-02-20
Add offline test runner initialization endpoint

peetya:


models/classes/runner/OfflineQtiRunnerService.php
```diff
@@ -0,0 +1,131 @@
...
+        $items = [];
+
+        foreach ($this->getItemIdentifiersFromTestMap($testMap) as $itemIdentifier) {
+            $itemRef = $runnerService->getItemHref($serviceContext, $itemIdentifier);
+
+            $items[$itemIdentifier] = [
+                'baseUrl' => $runnerService->getItemPublicUrl($serviceContext, $itemRef),
+                'itemData' => $runnerService->getItemData($serviceContext, $itemRef),
+                'itemState' => $runnerService->getItemState($serviceContext, $itemIdentifier),
+                'itemMetaData' => $runnerService->getFeedbacks($serviceContext, $itemRef),
```
jsconan: Is this key accurate? should not it be `feedbacks` instead?

actions/class.OfflineRunner.php
```diff
@@ -0,0 +1,57 @@
...
+     * @return void
+     */
+    public function init()
+    {
+        try {
+            $serviceContext = $this->getRunnerService()->initServiceContext($this->getServiceContext());
+            $response = $this->getInitResponse();
+
+            $response['items'] = $this->getOfflineRunnerService()->getItems($serviceContext);
+            $response['testBranchingRules'] = $this->getOfflineRunnerService()->getBranchingRules($serviceContext);
```
jsconan: Is the `testBranchingRules` entry injected in the `testMap` on the client side?

scripts/update/Updater.php
```diff
@@ -1747,5 +1747,11 @@ public function update($initialVersion) {
             $this->setVersion('30.6.0');
         }
 
+        if ($this->isVersion('30.6.0')) {
+            AclProxy::applyRule(new AccessRule('grant', 'http://www.tao.lu/Ontologies/TAO.rdf#DeliveryRole', array('ext'=>'taoQtiTest', 'mod' => 'OfflineRunner')));
+
+            $this->setVersion('30.7.0');
```
jsconan: Shouldn't you register `OfflineQtiRunnerService` too?

peetya: Fixed.

actions/class.OfflineRunner.php
peetya: I moved the branching rules into the `testMap` as we discussed.

actions/class.OfflineRunner.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * @throws common_Exception
+     * @return void
+     */
+    public function init()
+    {
+        try {
+            $serviceContext = $this->getRunnerService()->initServiceContext($this->getServiceContext());
+            $response = $this->getInitResponse();
```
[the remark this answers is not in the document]
peetya: Fixed.

actions/class.Runner.php
```diff
@@ -1065,4 +1035,81 @@ protected function getRawRequestParameter($name)
...
+            && $this->getRequestParameter('clientState') === 'paused'
+        ) {
+            $this->getRunnerService()->pause($serviceContext);
+            $this->getRunnerService()->check($serviceContext);
+        }
+
+        $result = $this->getRunnerService()->init($serviceContext);
+        $this->getRunnerService()->persist($serviceContext);
+
+        return $this->getInitSerializedResponse($result, $serviceContext);
```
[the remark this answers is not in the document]
peetya: I didn't change the order of the commands, just split into separate functions, so the functionality is not changed. Previous version: https://github.com/oat-sa/extension-tao-testqti/blob/master/actions/class.Runner.php#L289-L300

models/classes/runner/OfflineQtiRunnerService.php
peetya: The `getFeedbacks` method returns the whole content of the `variableElements.json` file and as I saw there are other data than just the feedbacks. You're right, the naming is not correct, but not sure that I need to rename the method or the object key. What do you think?

krampstudio: Both `feedbacks` and `itemMetaData` names aren't accurate and refer to something else. If this method returns the "variable elements" of an item (feedbacks, response delcaration and response processing, and conceptually anything that can help a test taker during the test), then why not naming it `getItemVariableElements` ? Or `getItemPrivateContent` , etc.

Regarding the response format, I don't think the `variableElements` should be sent separately from the `itemData`. Shouldn't they be merged merged upfront ? Remember those variable elements are only parts moved away from the original item definition. 
Then we can check if the response declaration is set or not (or we can have a flag).
What do you think ?

actions/class.Runner.php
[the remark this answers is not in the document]
jsconan: Hmmm... Actually the order changed, as long as the following code is moved to a function that is called later: https://github.com/oat-sa/extension-tao-testqti/blob/master/actions/class.Runner.php#L281-L286

models/classes/runner/OfflineQtiRunnerService.php
peetya: I created a new function called `getItemVariableElementsData()` and marked the `getFeedbacks()` as deprecated. Also I mapped the variable elements into the `itemData` as you suggested. Could you please take a look?

models/classes/runner/TestDefinitionSerializerService.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * @param string $filePath
+     * @return array
+     */
+    private function parseXmlToArray($filePath)
+    {
+        $xml = simplexml_load_file($filePath);
+        $json = json_decode(json_encode($xml), true);
```
Skyld: This is an associative array, not a json

models/classes/runner/OfflineQtiRunnerService.php
```diff
@@ -0,0 +1,147 @@
...
+            $result = [];
+
+            foreach ($array[$identifier] as $key => $value) {
+                $result[] = $this->getSubIdentifiersRecursively(
+                    $array[$identifier][$key],
+                    $identifiers
+                );
+            }
+
+            return array_merge(...$result);
```
Skyld: I don't know if we still support it, but this won't work below php 5.6

peetya: As far as I know we should support PHP5.6 or higher. @siwane can you please confirm or not? :) Thanks!

models/classes/runner/TestDefinitionSerializerService.php
peetya: Fixed.

actions/class.Runner.php
```diff
@@ -1065,4 +1040,80 @@ protected function getRawRequestParameter($name)
...
+    /**
+     * @param bool $isInitialized
+     * @param QtiRunnerServiceContext $serviceContext
+     * @return array
+     * @throws \oat\oatbox\service\exception\InvalidServiceManagerException
+     * @throws common_Exception
+     * @throws common_exception_InvalidArgumentType
+     * @throws common_ext_ExtensionException
+     */
+    private function getInitSerializedResponse($isInitialized, $serviceContext)
```
Skyld: I would remove the flag parameter and generate only the successful response here. Opinion?

actions/class.Runner.php
```diff
@@ -1065,4 +1040,80 @@ protected function getRawRequestParameter($name)
...
+    }
+
+    /**
+     * Checks the storeId request parameter and returns the last store id if set, false otherwise
+     *
+     * @param QtiRunnerServiceContext $serviceContext
+     * @return string|boolean
+     * @throws common_exception_InvalidArgumentType
+     */
+    private function getClientStoreId($serviceContext)
```
Skyld: Missing type hint

actions/class.Runner.php
```diff
@@ -1065,4 +1040,80 @@ protected function getRawRequestParameter($name)
...
+    {
+        if (
+            $this->hasRequestParameter('storeId')
+            && preg_match('/^[a-z0-9\-]+$/i', $this->getRequestParameter('storeId'))
+        ) {
+            return $this->getRunnerService()->switchClientStoreId(
```
Skyld: I'm not familiar with the underlying behaviour here, but the method name `getClientStoreId` indicates to retrieve something, not switching it. The naming could be misleading. What do you think?

actions/class.OfflineRunner.php
```diff
@@ -0,0 +1,86 @@
...
+     * Attaches the branching rules to the test map
+     * @param $testMap
+     * @param $serviceContext
+     * @return array
+     * @throws common_exception_Error
+     * @throws common_exception_InconsistentData
+     * @throws core_kernel_persistence_Exception
+     * @throws \oat\tao\model\websource\WebsourceNotFound
+     */
+    private function attachBranchingRulesToResponse($testMap, $serviceContext)
```
Skyld: Missing typehints

models/classes/runner/OfflineQtiRunnerService.php
```diff
@@ -0,0 +1,147 @@
...
+     * Returns an array of items, containing also confident data, like branching and response processing rules
+     *
+     * @param RunnerServiceContext $serviceContext
+     * @return array
+     * @throws \common_Exception
+     * @throws \common_exception_Error
+     * @throws \common_exception_InconsistentData
+     * @throws \common_exception_InvalidArgumentType
+     */
+    public function getItems($serviceContext)
```
Skyld: Missing typehint

models/classes/runner/OfflineQtiRunnerService.php
```diff
@@ -0,0 +1,147 @@
...
+    {
+        $this->getRunnerService()->assertQtiRunnerServiceContext($serviceContext);
```
Skyld: There are a lot of ServiceLocator calls. I know that it's legacy stuff, but maybe it would be more performant to cache it as a class property. Opinion?

actions/class.Runner.php
peetya: Seems strange, but the `switchClientStoreId()` returns with the `lastStoreId`.

models/classes/runner/OfflineQtiRunnerService.php
peetya: I've checked and the services get cached in the ServiceManager and the `get()` call will load them from the config only for the first time.

models/classes/runner/OfflineQtiRunnerService.php
peetya: Fixed.

actions/class.OfflineRunner.php
peetya: Fixed.

actions/class.Runner.php
peetya: Fixed.

actions/class.Runner.php
peetya: Fixed.

models/classes/runner/QtiRunnerService.php
```diff
@@ -1795,7 +1793,7 @@ protected function getTransmissionId(QtiRunnerServiceContext $context, $itemId =
      * @param RunnerServiceContext $context
      * @throws \common_exception_InvalidArgumentType
      */
-    protected function assertQtiRunnerServiceContext(RunnerServiceContext $context)
+    public function assertQtiRunnerServiceContext(RunnerServiceContext $context)
```
Skyld: Shouln't the `OfflineQtiRunnerService` inherit from `QtiRunnerService`, since they are sharing a lot of logic?

models/classes/runner/OfflineQtiRunnerService.php
siwane: I agree to use it, it is compatible with PHP5.6 and we support lower than this version.

models/classes/runner/TestDefinitionSerializerService.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * @param string $testCompilationUri
+     * @return string
+     * @throws \oat\tao\model\websource\WebsourceNotFound
+     * @throws \common_exception_InconsistentData
+     */
+    private function getPrivateDirectoryPath($testCompilationUri)
+    {
+        $privateDirectoryId = explode('|', $testCompilationUri)[0];
```
Skyld: Maybe it would be nice to check if the result is an empty array to avoid undefined index exceptions

models/classes/runner/map/TestMapBranchRuleExtender.php
```diff
@@ -0,0 +1,80 @@
...
+    private function addBranchRuleToTestMap($testMap, $testDefinitionSubObject, $testMapIdentifier, $id)
+    {
+        $testMap[$testMapIdentifier][$id][self::KEY_BRANCH_RULE] =
+            array_key_exists(self::KEY_BRANCH_RULE, $testDefinitionSubObject)
+                ? $testDefinitionSubObject[self::KEY_BRANCH_RULE]
+                : new \stdClass();
+
+        return $testMap;
+    }
+}
```
Skyld: Missing empty line at the end of file

models/classes/runner/map/TestMapBranchRuleExtender.php
```diff
@@ -0,0 +1,80 @@
...
+     * @param string $testMapIdentifier
+     * @param string $id
+     * @return array
+     */
+    private function addBranchRuleToTestMap($testMap, $testDefinitionSubObject, $testMapIdentifier, $id)
+    {
+        $testMap[$testMapIdentifier][$id][self::KEY_BRANCH_RULE] =
+            array_key_exists(self::KEY_BRANCH_RULE, $testDefinitionSubObject)
+                ? $testDefinitionSubObject[self::KEY_BRANCH_RULE]
+                : new \stdClass();
```
Skyld: Why stdClass?

peetya: To have an empty json object (`{}`) when returning the JSONResponse.

models/classes/runner/TestDefinitionSerializerService.php
peetya: `explode` won't return an empty array in any case.

models/classes/runner/map/TestMapBranchRuleExtender.php
peetya: Fixed.

models/classes/runner/QtiRunnerService.php
peetya: I was planning to handle them as separate services to be able to extend it by other extensions more easily.

PR #1439, opened 2019-02-25
Feature/tao 7305 use new request module

oatymart:
For https://oat-sa.atlassian.net/browse/TAO-7304 and https://oat-sa.atlassian.net/browse/TAO-7305

Depends on https://github.com/oat-sa/tao-core/pull/2006

Contains the changes necessary in the Test Runner to start using the `core/request` module.


oatymart:
Merged into https://github.com/oat-sa/extension-tao-testqti/pull/1465

PR #1447, opened 2019-03-06
protected method in the controller

zagovorichev:


actions/class.Runner.php
```diff
@@ -53,6 +53,7 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
 
     /**
      * taoQtiTest_actions_Runner constructor.
+     * @\oat\tao\model\routing\AnnotationReader\security("hide");
```
martijn-tao: Please import the class to prevent FQN usage

actions/class.TestContent.php
```diff
@@ -120,33 +120,4 @@ public function download() {
         
         tao_helpers_Http::returnFile($path);
     }
-    
-    /**
```
martijn-tao: Should this method really be deleted?

zagovorichev: according to this spike https://oat-sa.atlassian.net/browse/TAO-6784

PR #1448, opened 2019-03-06
Feature/tao 7460/tool state configuration scripts

alexeyvy:
https://oat-sa.atlassian.net/browse/TAO-7460

SetStateStorageForTools.php is for setting storage implementation
and 
SetToolsThatPreserveStates.php to register tools which should store their states on backend with help of the storage given to the first script


alexeyvy:
Hi Julien! There's no any test for the actions, so you only test the code by running SetStateStorageForTools and making sure ToolsStateStorage.conf.php updated with given persistence and storage (KvToolsStateStorage for 'kv', RdsToolsStateStorage for 'rds' and NoToolsStateStorage for 'no')

How to run the action: php index.php oat\taoQtiTest\scripts\tools\SetStateStorageForTools {STORAGE_TYPE(kv|rds|no)} {persistence_id(optional, by default it's 'default')} 

In case you chose rds, there's also a new table should be installed in the database. 

And the second action is to be used to set the value of 'tool-state-server-storage' section of testRunner.conf.php:
php index.php oat\taoQtiTest\scripts\tools\ SetToolsThatPreserveStates {comma separated list of tools to be set} 
This action also checks whether each tool exists on section 'plugins' of  testRunner.conf.php, otherwise throws an error to the terminal.

Do you think we need some test coverage for these actions (I'm starting to think so)?

scripts/tools/SetStateStorageForTools.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public function __invoke($params)
+    {
+        if (count($params) === 0) {
+            return new Report(Report::TYPE_ERROR, 'Storage type is not provided');
+        }
+
+        $storageType = $params[0];
+
+        switch ($storageType) {
```
martijn-tao: Maybe we can clean this up by dividing the switch cases into seperate methods, eg:

`$this->registerRdsStorage()`
`$this->registerKeyValueStorage()`
`$this->registerNoStorage()`

scripts/tools/SetStateStorageForTools.php
```diff
@@ -0,0 +1,106 @@
...
+     * @param string $persistenceId
+     * @return \common_persistence_Persistence|null
+     */
+    private function getPersistence($persistenceId)
+    {
+        try {
+            /** @var \common_persistence_Manager $persistenceManager */
+            $persistenceManager = $this->getServiceLocator()->get(\common_persistence_Manager::SERVICE_KEY);
+            $persistence = $persistenceManager->getPersistenceById($persistenceId);
+            return $persistence;
```
martijn-tao: This can become an inline return (`return $persistenceManager->getPersistenceById($persistenceId);`)

julien-sebire:
Hi Aliaksei, Martijn is back and taking care of the PR, I let him judge what could be necessary. Anyway, having some tests never hurts... ;-)
Keep the good job!
J.

scripts/tools/SetStateStorageForTools.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param array $params
+     * @return Report|null report if failed
+     * @throws \common_Exception
+     * @throws \oat\oatbox\service\exception\InvalidServiceManagerException
+     */
+    private function registerRdsStorage($params)
+    {
+        $persistenceId = array_key_exists(1, $params) ? $params[1] : 'default';
+        $persistence = $this->getPersistence($persistenceId);
+        if (!$persistence) {
```
martijn-tao: Since `$persistence` is not used in any other places, you could do an inline check here (`if (!$this->getPersistence($persistenceId)) {`)

scripts/tools/SetStateStorageForTools.php
```diff
@@ -0,0 +1,142 @@
...
+        if ($report instanceof Report) {
+            return $report;
+        }
+
+        return new Report(Report::TYPE_SUCCESS, 'Tool states service registered');
+    }
+
+    /**
+     * @param array $params
+     * @return Report|null report if failed
```
martijn-tao: I believe this should be `Report|void` instead

scripts/tools/SetStateStorageForTools.php
```diff
@@ -0,0 +1,142 @@
...
+        }
+        $this->getServiceManager()->register(
+            ToolsStateStorage::SERVICE_ID,
+            new RdsToolsStateStorage([ToolsStateStorage::OPTION_PERSISTENCE => $persistenceId])
+        );
+    }
+
+    /**
+     * @param array $params
+     * @return Report|null report if failed
```
martijn-tao: I believe this should be `Report|void` instead

PR #1461, opened 2019-03-22
fix tests

peetya:


test/unit/models/classes/runner/TestDefinitionSerializerServiceTest.php
```diff
@@ -115,5 +123,29 @@ public function testGetSerializedTestDefinition()
...
+         */
+        private function getResourceFactoryMock()
+        {
+            $resourceFactoryMock = $this
+                ->getMockBuilder(ResourceFactory::class)
+                ->getMock();
+
+            $resourceFactoryMock
+                ->method('create')
+                ->will($this->returnCallback([$this, 'resourceFactoryCreateCallback']));
```
Skyld: You can define the callback as an inline function, it's better imho. What do you think?

PR #1479, opened 2019-04-10
Feature/tao 7962 back btn

dietertaotesting:
Related/Used in: https://github.com/oat-sa/extension-tao-nfer/pull/115

`validateResponses` fires in both directions when used in the context of a non-linear test. This behavior may be unwanted when moving to the previous item. There is now a flag `validateOnPreviousMove` that can be used to avoid this. 

The second issue in this PR concerns a typo (_plagin vs plugin_) that had happened in a previous PR and that I spotted coincidentally. No functionality change.


views/js/runner/plugins/navigation/validateResponses.js
```diff
@@ -56,11 +56,18 @@ define([
...
+
+            var pluginConfig = this.getConfig();
+
+            this.getTestRunner().before('move', function (e, direction) {
                 var self = this;
                 var testContext = this.getTestContext();
                 var isInteracting = !this.getItemState(testContext.itemIdentifier, 'disabled');
 
+                if (!pluginConfig.validateOnPreviousMove && direction === 'previous') {
+                    return Promise.resolve();
```
zagovorichev: sorry, I don't understand this part. This method returned Promise before, but now it can return the result of the resolve method, correct?

dietertaotesting: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve
"The static `Promise.resolve` function returns a Promise that is resolved."

dietertaotesting:
added test case

scripts/update/Updater.php
```diff
@@ -1787,6 +1787,12 @@ public function update($initialVersion) {
             $this->setVersion('32.8.0');
         }
 
-        $this->skip('32.8.0', '32.10.1');
+        if ($this->isVersion('32.8.0')) {
+            $extension = $this->getServiceManager()->get(\common_ext_ExtensionsManager::SERVICE_ID)->getExtensionById('taoQtiTest');
+            $config = $extension->getConfig('testRunner');
+            $config['plugins']['validateResponses']['validateOnPreviousMove'] = true;
+            $extension->setConfig('testRunner', $config);
+            $this->setVersion('32.10.0');
```
zagovorichev: 32.11.0

PR #1454, opened 2019-03-14
Offline proxy and navigator component

peetya:
> Basic offline mode for test runner which helps to run NFER tests without internet connection.

To enable it run `php index.php "oat\taoQtiTest\scripts\install\SetOfflineTestRunnerConfig"`.

PR #1482, opened 2019-04-12
Fix/tao 7635/offline proxy component

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-7635

Fixing a broken unit test due to a last change in PR #1454.
The module `responseStore` was refactored, but the original version stayed till the end of the review, so the unit test were still working fine. As the obsolete version was removed, the test was failing then :(. I did not notice the unit test was using the obsolete version.


peetya:
thanks for the fix!

PR #1487, opened 2019-04-24
Fix/tao 7635 fix module case

krampstudio:
case issue on the dependency `core/promise` : 

On a case sensitive system (Linux for example...) the URLs and files paths are case sensitive : 

Try to run 
```
cd tao/views/build
npx grunt connect:test qunit:single --test=/taoQtiTest/views/js/test/runner/navigator/offlineNavigator/test.html
```

Without this fix the test is stuck.


anshul119:
Code only.

anshul119:
Checkout out the branch and tested, tests pass successfully

PR #1492, opened 2019-05-04
[TAO-8200] Refactor area masking tool to use dynamic component

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-8200
 
Improve movement/resizing behaviour of are masking tool, to increase user experience for touch screen devices users 
  
#### How to test
 
Create an ACT delivery with enabled area masking tool
Log in as a test taker
Run the delivery and try to resize/move area masking tool
   
#### Dependencies
 
Requires :
 - [ ] https://github.com/oat-sa/tao-core/pull/2090


views/js/runner/plugins/tools/areaMasking/mask.js
```diff
@@ -23,65 +23,102 @@
...
                 return this;
             }
         }, defaultConfig);
 
-
-        maskComponent
-            .setTemplate(areaMaskingTpl)
-            .on('render', function(){
-                var self     = this;
+        dynamicComponentInstance = dynamicComponent({}, dynamicComponentConfig)
```
jsconan: The config should be provided to the `.init()` method.

PR #1501, opened 2019-05-16
Fix/tao 8315/disable csrf for messages

oatymart:
Issue: https://oat-sa.atlassian.net/browse/TAO-8315

Related PR: https://github.com/oat-sa/tao-core/pull/2117

The short term fix for this bug will be to disable CSRF tokens for the "messages" feature used by some customers.

On the backend it was found that `checkSecurityToken()` was missing in 2 methods which should normally have it. It has been re-added.

However, until the token issue is fully understood, tokens will be turned off for messages, thus the call in `messsages()` is commented for now.

Should be tested in the Test Runner on ACT and NFER envs.

### DIFF
--- a/actions/class.Runner.php
+++ b/actions/class.Runner.php
@@ -957,13 +957,13 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
      */
     public function messages()
     {
-        // close the PHP session to prevent session overwriting and loss of security token for secured queries
-        session_write_close();
-
         $code = 200;
 
         try {
-            //$this->checkSecurityToken(); // should be on, but disabled as temporary fix for TAO-8315
+            $this->checkSecurityToken();
+
+            // close the PHP session to prevent session overwriting and loss of security token for secured queries
+            session_write_close();
 
             $input = \taoQtiCommon_helpers_Utils::readJsonPayload();
             if (!$input) {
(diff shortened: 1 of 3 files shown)


krampstudio:
I think it doesn't work because of the `session_write_close` https://github.com/oat-sa/extension-tao-testqti/pull/1501/files#diff-8883cfeb1526223f0ebf0cd07eff00deR961
which might not be necessary anymore. No ?

krampstudio:
After trying to :  
 - add the token in communicator  https://github.com/oat-sa/tao-core/pull/2117/files#diff-f2426157e0adc87c1096322346189d3aR129
 - remove the `session_write_close` https://github.com/oat-sa/extension-tao-testqti/pull/1501/files#diff-8883cfeb1526223f0ebf0cd07eff00deR961
 - uncomment the security check https://github.com/oat-sa/extension-tao-testqti/pull/1501/files#diff-8883cfeb1526223f0ebf0cd07eff00deR966

Everything looks to be working, tested quickly

SashaTAO:
I am not sure if this change (removing `session_write_close`) will not bring back concurrency issue. Maybe it would be better to turn off CSRF verification for this endpoint for the moment and work on correct fix without pressure. What do you think?

oatymart:
After discussions we will enable the tokens again but move the `session_write_close()` just after the check.

PR #1505, opened 2019-05-20
Fix/TAO-7216/Timer doesn't start after the audio is fully loaded when the connection is slow 

bziondik:
DRAFT PR!!!

Related to: https://oat-sa.atlassian.net/browse/TAO-7216
Handling warning from ItemRunner in case of errors during Item rendering

Related to https://github.com/oat-sa/extension-tao-itemqti/pull/1279


views/js/controller/runner/runner.js
```diff
@@ -171,6 +171,24 @@ define([
                 }
             };
 
+            /**
+             * Handles warnings
+             * @param {String} [displayMessage] - an alternate message to display
+             */
+            var onWarning = function onWarning(err, displayMessage) {
```
jsconan: `onError` and `onWarning` are very similar. Could you refactor them in order to share the same mechanism between the rwo?

manifest.php
```diff
@@ -39,7 +39,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '33.2.1',
+    'version'     => '33.2.2',
```
jsconan: Could you also mention the dependency with taoQtiItem? (`20.0.2`)

views/js/controller/runner/runner.js
```diff
@@ -171,6 +171,24 @@ define([
...
+                displayMessage = displayMessage || err.message;
+
+                if(!_.isString(displayMessage)){
+                    displayMessage = JSON.stringify(displayMessage);
+                }
+
+                logger.error({ displayMessage : displayMessage }, err);
```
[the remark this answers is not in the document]
bziondik: fixed

views/js/controller/runner/runner.js
bziondik: refactored

PR #1513, opened 2019-05-30
Feature/tao 8402/sync offline

handleman:
*task* : https://oat-sa.atlassian.net/browse/TAO-8402

**short description** : extend offline proxy component to make it able to synchronise the test state on every action.

**full description** : 
offline proxy enable test runner to work regardless of internet connection, saving data only at the very end of test. The problem was that this approach doesn't allow user to proceed test after interruption on other machine. 

**problem was solved** this feature made proxy to synchronise test data on every action user perform. So test is executed offline with synchronisation of all the data when connection is fine, and store all actions in the queue when there is no connection.

**how to test**
First of all you need to switch test runner to "offline mode" with command : `php index.php "oat\taoQtiTest\scripts\install\SetOfflineTestRunnerConfig"` 
After that, your test would be able to be run offline (without internet connection) .
1. run the command above. (with) connection online
2. login in your environment in some delivery 
3. start the test, wait until "init" API call finished. (use network panel in devtools)
4. turn of the wifi on your laptop, or just use "offline" checkbox on " network" panel in your developers tools.
5. Now you should be able to go through test items offline.
6. turn on the internet on some item and press "next" control.
7. You should see that "message" api call is pending holding array of all actions were performed during offline.

And you can logout at any moment and should be able to proceed from the las synced item.


views/js/runner/proxy/offline/proxy.js
```diff
@@ -131,98 +131,123 @@ define([
...
+                        } else{
                             return self.syncData().then(function() {
                                 return resolve(result);
                             }).catch(function() {
                                 return resolve({ success: false });
                             });
                         }
-                    }
-
-                    if (action === 'skip') {
```
AntonTsymuk: is this statement removed by purpose?

handleman: You were right, I've moved it back

handleman:
> * I am not sure, but Is this changes require any any unit tests updates?
> * Also please adjust proper formatting to your code(indentations, missing spaces in operators)

Changes have affected only one components method, without any change in its attributes or component api calls sequence in general. 
The behaviour itself is related closely to specific user scenario, so it makes sense to test it during acceptance/behaviour testing stage. I don't see here need for unit test change.

handleman:
@hutnikau  @AntonTsymuk  Could you do another round?

PR #1515, opened 2019-05-31
Feature/nex 81 es2015 migration

krampstudio:
Related to https://oat-sa.atlassian.net/browse/NEX-81

 - Bundles with babel
 - migrate one controller to ES2015

Requires PR : 

 - [x] https://github.com/oat-sa/tao-core/pull/2156


views/js/controller/creator/creator.js
```diff
@@ -78,14 +77,17 @@ define([
...
-            var self = this;
-            var $container = $('#test-creator');
-            var $saver = $('#saver');
-            var binder, binderOptions, modelOverseer;
+        start(options){
+            const self = this;
+            const $container = $('#test-creator');
+            const $saver = $('#saver');
+
             var creatorContext;
```
anshul119: What do you think about using let instead of var? It can be good as it prevents re declaration, usage before declaration etc

views/js/controller/creator/creator.js
```diff
@@ -78,14 +77,17 @@ define([
...
-          * @param {Boolean} [options.guidedNavigation  = false]- feature flag for the guided navigation
+          * @param {Boolean} [options.guidedNavigation=false] - feature flag for the guided navigation
           */
-        start : function(options){
-            var self = this;
-            var $container = $('#test-creator');
-            var $saver = $('#saver');
-            var binder, binderOptions, modelOverseer;
+        start(options){
+            const self = this;
```
anshul119: you dont need this anymore :)

PR #1516, opened 2019-06-03
Feature/nex 90 update renamed ui module

btamas:
Update reference to renamed ui module.

- [ ] dependency https://github.com/oat-sa/tao-core/pull/2158


krampstudio:
*Please do not merge before next sprint*

PR #1520, opened 2019-06-11
Fix/TAO-8318/remove feedback in case terminated session

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-8318

Added condition to cover the case when in catch we get a response, not Error
https://github.com/oat-sa/tao-core-sdk-fe/blob/de027f48faf88411bb276289fcc9ca720f9cf123/src/core/request.js#L187-L189

### DIFF
--- a/views/js/runner/proxy/qtiServiceProxy.js
+++ b/views/js/runner/proxy/qtiServiceProxy.js
@@ -105,9 +105,6 @@ define([
                     if (error.data && self.isConnectivityError(error.data)) {
                         self.setOffline('request');
                     }
-                    if (error instanceof Error) { //in case status 200 OK ans response.success:false we don't have Error
-                        feedback().error(error);
-                    }
                     return Promise.reject(error);
                 });
             };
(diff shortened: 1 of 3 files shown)


views/js/runner/proxy/qtiServiceProxy.js
```diff
@@ -105,7 +105,9 @@ define([
                     if (error.data && self.isConnectivityError(error.data)) {
                         self.setOffline('request');
                     }
-                    feedback().error(error);
+                    if (error instanceof Error) { //in case status 200 OK ans response.success:false we don't have Error
```
oatymart: I don't think it makes sense to check if `error instanceof Error` because ideally (from a semantic perspective), it always will be true. Since in this case we are actually breaking the semantics and trying to handle a non-Error-type rejection coming from the previous `.then()`, we could attach some unique property to the response there (L101) and check for it here.

(Also, it's questionable whether the `feedback()` is needed at all... if you are confident that all kinds of Test Runner errors are being handled in a user-friendly way (by dialogs, etc.), and they do not need to be shown as feedback errors, then you could simply remove it.)

bziondik: Removed feedback, it was added within refactoring (added core/request module, mixing old functionality of core/dataProvider/request and qtiServiceProxy) see changes in file qtiServiceProxy.js here  https://github.com/oat-sa/extension-tao-testqti/pull/1465/files
Previously there was no feedback here.

Tested on errors:
1. terminated session
2. lost connection
3. token expired

PR #1526, opened 2019-06-14
TAO-8362 - Remove error code from waiting dialog

ferenckiss89:
Related ticket: https://oat-sa.atlassian.net/browse/TAO-8362

- error code + message is not being sent anymore in case of reconnection issue
- default value has been set for the message parameter in case the error message is not passed


views/js/runner/plugins/controls/connectivity/connectivity.js
```diff
@@ -119,7 +119,7 @@ define([
 
                         //creates the waiting modal dialog
                         dialog = waitingDialog({
-                            message : __('You are encountering a prolonged connectivity loss. ') + message,
+                            message : __('You are encountering a prolonged connectivity loss. '),
```
oatymart: We could keep the message here but only include it if it is a string (see below).

oatymart: Then, we could:
L198: remove the passing of the error message: `self.displayWaitingDialog()`
L177: keep passing the (could be informational) message: no change required

oatymart: In this way, the message would be gone only in the case of connectivity error. And the change to current behaviour is minimised.

ferenckiss89: Good point. Fixing it.

PR #1530, opened 2019-06-20
[TAO-7016] NFER-16 Disconnected: file save at end of test

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-7016
 
Add missing data to json file which can be downloaded and later exported if there is no connection during delivery execution in offline mode 
 
#### How to test

- Prepare tao instance with NFER extension enabled
- Switch test runner to "offline mode" with command : php index.php "oat\taoQtiTest\scripts\install\SetOfflineTestRunnerConfig"
- Login as a test taker and start delivery execution
- After the delivery page fully loaded, disabled internet connection in browser dev tools network tab
- Finish the delivery in offline mode, the modal window with countdown and download results option will appear  
- Wait until countdown is finished and download test results
- Login into backoffice and navigate to "Tools" -> "Upload offline test completion"
- Import test results and check that it was added to delivery execution results

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -39,7 +39,7 @@ return array(
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '33.7.3',
+    'version'     => '33.7.4',
     'author'      => 'Open Assessment Technologies',
     'requires'    => array(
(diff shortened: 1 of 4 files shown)


handleman:
Problem solved. Code looks fine.

PR #1534, opened 2019-06-26
Fix/report data path

Yury-Razhkou:
Useless operation with setting path data. Few lines below there is new way of setting it.


models/classes/export/class.TestExport.php
```diff
@@ -115,8 +115,6 @@ public function export($formValues, $destination)
 
         if (!$report->containsError() && $subjectUri) {
             $this->getEventManager()->trigger(new QtiTestExportEvent(new core_kernel_classes_Resource($subjectUri)));
-
-            $report->setData($path);
```
SashaTAO: Could you please fix this test as well https://github.com/oat-sa/extension-tao-testqti/blob/master/test/integration/QtiTestExporterTest.php#L184
It still expects that $report->getData() will return path as a string

PR #1541, opened 2019-07-02
Fix/tao 8623 dynamic components

anshul119:
Ticket: https://oat-sa.atlassian.net/browse/TAO-8623

Because of the extraction of tao ui components(dynamicComponent) to a separate npm repo, the order of loading the stylesheet is different. The dynamicComponent's stylesheet is loaded after the test runner tool specific stylesheet resulting in bad UI. 

To tackle the issue I made style of test runner . tool more specific.

Companion PR:

- [ ] https://github.com/oat-sa/tao-core-ui-fe/pull/35
- [ ] https://github.com/oat-sa/tao-core/pull/2205


views/scss/inc/_magnifier.scss
```diff
@@ -111,11 +115,7 @@
                 }
 
                 &.close {
```
btamas: Why do you need the empty selector?

anshul119: I dont, removed :D

PR #1557, opened 2019-07-19
Fix/sass config override

krampstudio:
Ensure the `includePaths` options is defined at the correct level (the task and not the overriding the main config)
Remove unused entries
Reformat and use `path.join` instead of concats

### DIFF
--- a/views/build/grunt/sass.js
+++ b/views/build/grunt/sass.js
@@ -1,27 +1,61 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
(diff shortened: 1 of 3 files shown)


views/build/grunt/sass.js
```diff
@@ -1,27 +1,61 @@
...
+ *
+ * @param {Object} grunt - the grunt object (by convention)
+ */
 module.exports = function(grunt) {
     'use strict';
 
-    var root    = grunt.option('root') + '/taoQtiTest/views/';
-    var pluginDir = root + 'js/runner/plugins/';
+    const sassConfig = grunt.config('sass') || {};
+    const root       = path.join(grunt.option('root'), 'x/taoQtiTest/views/');
```
jsconan: is the `x` a typo?

PR #1560, opened 2019-07-24
[TAO-8713] NFERlightweight: Passing a test with an offline test runner leads to different and incorrect displaying of the final test score

AntonTsymuk:
Update npm packages


handleman:
looks fine

PR #1563, opened 2019-07-30
Refactor/tao 8709 enhance compilation data service

bugalot:
**Please make sure this is heavily tested, with and without CAT support**
This PR was an experimental PR to describe how a possibile solution could be. Please take this into consideration while testing and eventually, merging this PR. Thanks!

This **draft** PR aims at providing alternative delivery compilation procedures to get rid of compiled PHP files in delivery assemblies.

**1. Relying on XML `compact-test` file instead of PHP**

By enabling the `XmlCompilationDataService`, the Delivery Compilation Process will produce a test definition as XML (`compact-test.xml`) instead of a PHP one `compact-test.php`. **_Be carefull, enabling this particular Compilation Data Service will make previously compiled Deliveries not working anymore. In case we need retro-compatibility, we'll have to find another strategy._**

To enable the `XmlCompilationDataService`, please run the intended script `sudo -u www-data php index.php "oat\taoQtiTest\scripts\install\SetupXmlCompilation"`.

**2. Relying on JSON `test-meta` files instead of PHP**

All Compilation Data Services will now generate a JSON version of the former `test-meta.php` file. This is retro-compatible. Indeed, a fallback has been put in place. 

To summarize, 

* This PR doesn't hurt any running installation. Update process will be seamless as it does not change any configuration.
* Enabling the `XmlCompilationDataService` will make formerly compiled deliveries obsolete and not working anymore. So this strategy is designed for fresh install + call to `sudo -u www-data php index.php "oat\taoQtiTest\scripts\install\SetupXmlCompilation"` script.
* In terms of Central Server/VM strategy, relying on the `XmlCompilationDataService` will make the trick, **but only in case of freshly installed** Central Server / freshly built VM.


models/classes/CompilationDataService.php
```diff
@@ -47,5 +48,51 @@ public function __construct($options = []) {
...
+     * @return mixed
+     * @throws \common_Exception
+     */
+    public function readCompilationMetadata(\tao_models_classes_service_StorageDirectory $compilationDirectory)
+    {
+        try {
+            $data = $compilationDirectory->read(\taoQtiTest_models_classes_QtiTestService::TEST_COMPILED_META_FILENAME . '.json');
+            return json_decode($data, true);
+
+        } catch (\Exception $e) {
```
hutnikau: Catching general `\Exception` seems to me not really reliable solution. Is that possible to get rid of all the legacy metadata php files by recompiling them into json during update?

models/classes/cat/CatService.php
```diff
@@ -150,7 +150,7 @@ public function getAssessmentItemRefByIdentifier(\tao_models_classes_service_Sto
         $compilationDataService = $this->getServiceLocator()->get(CompilationDataService::SERVICE_ID);
         $filename = "adaptive-assessment-item-ref-${identifier}";
         
-        return $compilationDataService->readPhpCompilationData(
+        return $compilationDataService->readCompilationData(
             $privateCompilationDirectory,
             "${filename}.php",
```
hutnikau: seems like `php` extension is not needed anymore

models/classes/cat/CatService.php
```diff
@@ -191,7 +191,7 @@ public function getAssessmentItemRefsByPlaceholder(\tao_models_classes_service_S
         $compilationDataService = $this->getServiceLocator()->get(CompilationDataService::SERVICE_ID);
         $filename = "adaptive-assessment-section-${adaptiveSectionId}";
         
-        $component = $compilationDataService->readPhpCompilationData(
+        $component = $compilationDataService->readCompilationData(
             $privateCompilationDirectory,
             "${filename}.php",
```
hutnikau: seems like `php` extension is not needed anymore

models/classes/class.QtiTestCompiler.php
```diff
@@ -897,7 +897,7 @@ protected function compileAdaptive(AssessmentTest $test)
                     \common_Logger::d("QTI Adaptive Section with identifier '" . $current->getIdentifier() . "' found.");
                     
                     // Deal with AssessmentSection Compiling.
-                    $compilationDataService->writePhpCompilationData(
+                    $compilationDataService->writeCompilationData(
                         $this->getPrivateDirectory(),
                         "adaptive-assessment-section-${sectionIdentifier}.php",
```
hutnikau: extension :)

models/classes/class.QtiTestCompiler.php
```diff
@@ -910,7 +910,7 @@ protected function compileAdaptive(AssessmentTest $test)
                             $sectionPartHref = $sectionPart->getHref();
                             
                             // Deal with AssessmentItemRef Compiling.
-                            $compilationDataService->writePhpCompilationData(
+                            $compilationDataService->writeCompilationData(
                                 $this->getPrivateDirectory(),
                                 "adaptive-assessment-item-ref-${sectionPartIdentifier}.php",
```
hutnikau: extension :)

models/classes/CompilationDataService.php
bugalot: As we need a quick resolution, I don't think it is absolutely required at the present time. That can come in later on.

zagovorichev:
Checked with CAT, it works for me. Good job.

models/classes/class.QtiTestCompiler.php
zagovorichev: fixed

PR #1565, opened 2019-07-31
Feature/tao 8733/extended time round

yuryfilippovich:
Related to https://oat-sa.atlassian.net/browse/TAO-8733
1 create timer strategy service as the timer dependency
2 add new service as a relation in QtiTimer class
2 Update version
3 Create a default service config
4 Unit tests


config/default/TimerStrategyService.conf.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+use oat\taoQtiTest\models\runner\time\TimerStrategyService;
+
+return new TimerStrategyService();
```
zagovorichev: if timers defined in the taoTest and interface is used there, it would be better to move this there too

models/classes/runner/time/TimerStrategyService.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+
+namespace oat\taoQtiTest\models\runner\time;
+
+use oat\oatbox\service\ConfigurableService;
+
+class TimerStrategyService extends ConfigurableService implements TimerStrategyInterface
+{
+    const SERVICE_ID = 'taoQtiTest/TimerStrategyService';
```
zagovorichev: service_id better to move into the TimerStrategyInterface

zagovorichev: how do you think: do we need this service? can we use it without TimerService? outside of this service, probably it would be better to delete SERVICE_ID from this class and initialize it like QtiTimerService(['timer_strategy' => new TimerStrategyService()]) (it's not working code, but you have an idea)

scripts/update/Updater.php
```diff
@@ -1904,5 +1904,13 @@ public function update($initialVersion) {
         }
 
         $this->skip('34.0.0', '34.2.0');
+
+        if ($this->isVersion('34.2.0')) {
+            $this->getServiceManager()->register(
+                TimerStrategyService::SERVICE_ID,
```
zagovorichev: still not sure that we need it as a full service, maybe timer strategy provider would be enough?

models/classes/runner/time/TimerStrategyService.php
yuryfilippovich: We use it in OdsResultsPayload to generate json and in DeliveryExecutionManagerService

zagovorichev:
sorry, @yuryfilippovich , we have template of the description for the PR here: https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/11338720/Development+Workflow+with+Github [Pull request template example]

models/classes/runner/time/TimerStrategyService.php
```diff
@@ -0,0 +1,51 @@
...
+            return 0;
+        }
+
+        return $this->getNewTime($time, $multiplier) - $time;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getNewTime($time, $multiplier)
```
zagovorichev: maybe we can find another name for this method which will describe its mission better

PR #1569, opened 2019-08-08
update test runner extensions

ferenckiss89:
Version update of test runner fe extensions and newly builded css files.


manifest.php
```diff
@@ -41,7 +41,7 @@
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '34.3.1',
+    'version'     => '34.3.2',
```
krampstudio: Same remark, let's jump a feature version and enforce the dependency to `tao >= 38.5.0` and `taoTests` also

views/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "license": "GPL-2.0",
   "dependencies": {
-    "@oat-sa/tao-test-runner": "^0.2.1",
-    "@oat-sa/tao-test-runner-qti": "^0.3.2"
+    "@oat-sa/tao-test-runner": "^0.4.0",
```
krampstudio: let's remove all carrets and tilds for now

manifest.php
ferenckiss89: ok

views/package.json
ferenckiss89: yep

PR #1575, opened 2019-08-19
Fix/tao 8725/update interactjs

handleman:
**task:** https://oat-sa.atlassian.net/browse/TAO-8725
**description:** 
on hybrid laptops ( both with keyboard and touchscreen) all plugins (widgets) of test runner were unable to move through the touchscreen instead of mouse or touchpad. 

**solution**
interactjs lib which is responsible for all interface interactions was updated to newer version which supports modern " pointer events" browser API which solved the issue.

**related PR**
https://github.com/oat-sa/tao-core/pull/2251
https://github.com/oat-sa/extension-tao-itemqti/pull/1327

**related NPM releases**
https://github.com/oat-sa/tao-core-libs-fe/releases/tag/0.3.0
https://github.com/oat-sa/tao-core-ui-fe/releases/tag/0.23.0
https://github.com/oat-sa/tao-test-runner-qti-fe/releases/tag/0.7.0


handleman:
no need increase version of NPM since it was updated on develop branch

PR #1579, opened 2019-08-23
Feature/nex 71/e2e test import refactor

oatymart:
https://oat-sa.atlassian.net/browse/NEX-71

- split and parameterised the setup and cleanup commands (`importTestPackage()`, `publishTest()`, `setDeliveryForGuests()`, `deleteItem()`, `deleteTest()`, `deleteDelivery()`) so they can be reused
- ~~added 2 tests as zip files, replacing base64 style~~ base64 tests remain
- changed folder structure
- added tools test outline (no assertions yet)
- new command: `startTest()`


views/js/e2e/_helpers/_setup/setupCommands.js
```diff
@@ -37,19 +35,29 @@ Cypress.Commands.add('publishImportedTest', () => {
     // Wait until test import request finishes
     cy.wait('@testImportIndex');
 
+    // Prepare to load fixture
```
ricproenca-tao: What if I want to use base64 string instead of zip file?
I think we should allow both.

views/js/e2e/runner/tools/tools.spec.js
```diff
@@ -0,0 +1,82 @@
...
+        cy.setDeliveryForGuests('e2e Tools test');
+        cy.logout();
+    });
+
+    /**
+     * Log in & start the test
+     */
+    beforeEach(() => {
+        cy.guestLogin();
+        // cy.startTest('e2e Tools test'); // TODO:
```
ricproenca-tao: Already did this function
```
Cypress.Commands.add('startTest', (testName) => {
    cy.get(runnerSelectors.testList)
        .find(runnerSelectors.availableDeliveries)
        .contains(testName)
        .click();

    cy.wait(['@testRunnerGet', '@testRunnerPost']);
});
```
If you agree with this, please add it to your code.

views/js/e2e/_helpers/_setup/setupCommands.js
ferenckiss89: base64, as we agreed

views/js/e2e/runner/tools/tools.spec.js
oatymart: ok, added

PR #1582, opened 2019-08-26
Feature/nex 182/e2e test runner navigation tests 2

ricproenca-tao:
Related to task:
https://oat-sa.atlassian.net/browse/NEX-182

Description:
e2e test cases to cover the navigation on the test runner (next, prev, skip, end)
It covers linear and non linear tests

Requires:
tao-e2e-runner installed into your tao/views/build folder.

How to test:
Please checkout the branch and run the tests:
`npx cypress run` (run tests)
`npx cypress open` (open cypress visual tests)

Tests:
```
Navigation


    Linear Test Navigation

      First Item
        √ Has proper navigation buttons (39222ms)
        √ Move to next item (7252ms)
        √ Skip to next item (5941ms)

      Second Item
        √ Has proper navigation buttons (8423ms)
        √ Move to next item (9531ms)
        √ Skip to next item (7652ms)

      Last Item
        √ Has proper navigation buttons (7447ms)
        √ End test (8383ms)
        √ Skip and end test (7860ms)


    Non Linear Test Navigation

      First Item
        √ Has proper navigation buttons (26758ms)
        √ Move to next item (6401ms)
        √ Skip to next item (7276ms)

      Second Item
        √ Has proper navigation buttons (7716ms)
        √ Move to next item (8248ms)
        √ Skip to next item (8795ms)
        √ Move to previous item (8868ms)

      Last Item
        √ Has proper navigation buttons (7373ms)
        √ Move to previous item (8029ms)
        √ End test (7648ms)
        √ Skip and end test (8904ms)
```


views/js/e2e/_helpers/commands/navigationCommands.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Copyright (c) 2019 (original work) Open Assessment Technologies SA ;
+ */
+
+import setupSelectors from "../selectors/setupSelectors";
+
+Cypress.Commands.add('nextItem', () => {
+    cy.log('COMMAND: nextItem');
+    cy.get(setupSelectors.testNavigation.nextItem).click();
+    cy.wait(['@testRunnerGetItem', '@testRunnerInit']);
```
ferenckiss89: testRunnerInit is not requested in my env, without that it is working

views/js/e2e/_helpers/commands/navigationCommands.js
```diff
@@ -0,0 +1,50 @@
...
+});
+
+Cypress.Commands.add('previousItem', () => {
+    cy.log('COMMAND: previousItem');
+    cy.get(setupSelectors.testNavigation.previousItem).click();
+    cy.wait(['@testRunnerGetItem', '@testRunnerInit']);
```
ferenckiss89: testRunnerInit is not requested in my env, without that it is working

views/js/e2e/_helpers/commands/navigationCommands.js
```diff
@@ -0,0 +1,50 @@
...
+});
+
+Cypress.Commands.add('skipItem', () => {
+    cy.log('COMMAND: skipItem');
+    cy.get(setupSelectors.testNavigation.skipItem).click();
+    cy.wait(['@testRunnerGetItem', '@testRunnerInit']);
```
ferenckiss89: testRunnerInit is not requested in my env, without that it is working

views/js/e2e/_helpers/commands/navigationCommands.js
```diff
@@ -0,0 +1,50 @@
...
+Cypress.Commands.add('endTest', () => {
+    cy.log('COMMAND: endTest');
+    cy.get(setupSelectors.testNavigation.endTest).click();
+    cy.wait('@testRunnerInit');
+});
+
+Cypress.Commands.add('skipAndEndTest', () => {
+    cy.log('COMMAND: skipAndEndTest');
+    cy.get(setupSelectors.testNavigation.skipAndEndTest).click();
+    cy.wait('@testRunnerInit');
```
ferenckiss89: testRunnerInit is not requested in my env, without that it is working

views/js/e2e/_helpers/commands/navigationCommands.js
oatymart: I don't believe `@testRunnerInit` will ever be called for `next`, `prev`, `move`.

views/js/e2e/_helpers/commands/navigationCommands.js
```diff
@@ -0,0 +1,50 @@
...
+Cypress.Commands.add('skipItem', () => {
+    cy.log('COMMAND: skipItem');
+    cy.get(setupSelectors.testNavigation.skipItem).click();
+    cy.wait(['@testRunnerGetItem', '@testRunnerInit']);
+});
+
+Cypress.Commands.add('endTest', () => {
+    cy.log('COMMAND: endTest');
+    cy.get(setupSelectors.testNavigation.endTest).click();
+    cy.wait('@testRunnerInit');
```
oatymart: `@testRunnerInit` also doesn't get hit at the end of the test. Instead, I think you should assert that you returned to `/taoDelivery/DeliveryServer/index`.

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -43,7 +43,7 @@ Cypress.Commands.add('importTestPackage', (fileContent, fileName) => {
     cy.get(setupSelectors.testsPage.fileInput).upload(
         {
             fileContent,
-            fileName,
+            fileName: `${fileName}.zip`,
```
oatymart: Seems to be a Windows thing, but ok, we can go with this for all platforms.

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -124,5 +130,6 @@ Cypress.Commands.add('startTest', (testName) => {
         .contains(`Delivery of ${testName}`)
         .click();
 
-    cy.wait(['@testRunnerInit', '@testRunnerGetItem']);
+    cy.wait(['@testRunnerGetItem', '@testRunnerInit']);
```
oatymart: The order doesn't matter to Cypress (it's waiting for the responses, so requests made in the order `1,2` could return as `1,2` or `2,1` depending on network speed).

views/js/e2e/_helpers/selectors/setupSelectors.js
```diff
@@ -20,7 +20,7 @@
  * CSS Selectors for setup
  */
 export default {
-    resourceTree: '.resource-tree',
+    resourceTree: '.taotree',
```
oatymart: Use `testsPage: { rootTestClass:'li[data-uri="http://www.tao.lu/Ontologies/TAOTest.rdf#Test"]'`

views/js/e2e/_helpers/commands/cleanupCommands.js
```diff
@@ -37,6 +37,9 @@ Cypress.Commands.add('deleteItem', (itemName) => {
     // Delete test
     cy.get(cleanupSelectors.itemsPage.itemDeleteButton).click();
 
+    //windows workaround
+    cy.wait(1000);
```
oatymart: Would be nice to avoid these. Does it not work if you add a `.should('exist')` before the `.click()`? Cypress is supposed to automatically retry until the element is present or it times out.

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -53,6 +53,9 @@ Cypress.Commands.add('importTestPackage', (fileContent, fileName) => {
         }
     );
 
+    // windows workaround
+    cy.wait(2000);
```
oatymart: See previous comment. If there is no network request to wait for, try to do an assertion about the expected element. We can use  `.wait(2000)` if we have to, but it should be the last resort.

views/js/e2e/_helpers/commands/navigationCommands.js
ferenckiss89: and skip and end

ricproenca-tao: forgot to change this with proper routes.

views/js/e2e/_helpers/commands/navigationCommands.js
ricproenca-tao: forgot to change this with proper routes.

views/js/e2e/_helpers/commands/navigationCommands.js
ricproenca-tao: forgot to change this with proper routes.

views/js/e2e/_helpers/commands/navigationCommands.js
ricproenca-tao: forgot to change this with proper routes.

views/js/e2e/_helpers/commands/navigationCommands.js
ricproenca-tao: forgot to change this with proper routes.

views/js/e2e/_helpers/commands/navigationCommands.js
ricproenca-tao: In my opinion, I think we shouldn't make assertions inside commands.
The assertion is there in `navigation.js line:348`

views/js/e2e/_helpers/commands/setupCommands.js
ricproenca-tao: Yes, it's a Windows thing and we need it. :(

views/js/e2e/_helpers/commands/setupCommands.js
ricproenca-tao: I know @martijn-tao, but personally I prefer to put them in order.
Thanks for the warning anyway.

views/js/e2e/_helpers/selectors/setupSelectors.js
ricproenca-tao: yes, I agree with you. It's much better.

views/js/e2e/_helpers/commands/cleanupCommands.js
ricproenca-tao: I just removed it and it seems that I don't have this problem now.

views/js/e2e/_helpers/commands/setupCommands.js
ricproenca-tao: change for the proper request `cy.wait('@fileUpload');`

views/js/e2e/_helpers/commands/setupCommands.js
oatymart: That's fine. But isn't the usual order of the Test Runner `init` before `getItem`?

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -126,3 +131,4 @@ Cypress.Commands.add('startTest', (testName) => {
 
     cy.wait(['@testRunnerInit', '@testRunnerGetItem']);
```
ferenckiss89: this line seems like is not needed, and if you remove it then the route @testRunnerInit is completely unnecessary

ricproenca-tao: The purpose of this line is wait til you have an item in the DOM.
I was thinking in a way that if we don't have an `init test request` and `get me the first item request` we cannot say that we are inside a test.
I can remove it, if you think its better.

ferenckiss89: Maybe it is just my env, but sometimes it works, sometimes does not.

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -79,7 +79,7 @@ Cypress.Commands.add('publishTest', (testName) => {
 
     // Select tree node
     cy.get(setupSelectors.testsPage.rootTestClass).within(() => {
-        cy.contains(testName).click({ force: true });
+        cy.contains(testName).click();
```
oatymart: I believe we may have used `force: true` here because the list item for your desired test can be off screen, depending on the size of the viewport (and number of other tests).

ricproenca-tao: ok. I will add again.

PR #1590, opened 2019-09-06
E2E Tests for remote delivery publishing and start

ferenckiss89:
Jira: https://oat-sa.atlassian.net/browse/NEX-290

Changes:
- [x] add extra param to the `publishTest` setupCommand to be able to test remote publication (local publication by default)
   - [x] when `deliveryType` param `remote` is there, we select TAO Remote tab
   - [x] ~~we don’t have to select a tenant~~ I hardcoded a tenant which works, might change
- [x] LTI call to start remote delivery
   - [x] start the remote delivery
   - [x] check that it is running from a different URL
   - [x] finish the test and be redirected to the deliveries page
- [x] handle both TaskQueue / non-TaskQueue actions equally well
- [x] version bump

Note:
- use [nextgen-stack](https://github.com/oat-sa/nextgen-stack/tree/develop) to ensure you have all the necessary repos and extension to be able to perform the above-mentioned actions
- make sure the taskQueue is running (also good to test without it)
- make sure messenger-consume is running

Extra prerequisites for testing (changes will eventually go into a separate PR):
`cypress.env.json` values:
```
{
  "baseUrl": "https://currentgen.docker.localhost",
  "taoDeliverBackendUrl": "https://deliver.docker.localhost",
  "taoDeliverFrontendUrl": "https://testrunner.docker.localhost",
  "bypassBackOffice": "false"
}
```


views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -52,24 +130,41 @@ Cypress.Commands.add('importTestPackage', (fileContent, fileName) => {
...
 
-    // Wait until test import request finishes
-    cy.wait(['@testImportIndex', '@taskQueueWebApi', '@taskQueueWebApi'], { timeout: 15000 });
+    /*
+     * Two different processes to handle different UX
+     * for taoTaskQueue enabled or disabled
+     */
+    cy.get('@isTaskQueueWorking').then(working => {
+        if (working) {
+            cy.wait(Array(4).fill('@taskQueueWebApi'));
```
btamas: 5000 ms was not enough for my computer.

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -83,17 +178,47 @@ Cypress.Commands.add('publishTest', (testName) => {
...
+    // Publish selected test
     cy.get(setupSelectors.testsPage.testPublishButton).click();
 
+    if (deliveryType === 'remote') {
+        // Selects TAO Remote tab
+        cy.get(setupSelectors.testsPage.deliveryTypeTabs).contains('TAO Remote').click();
+    }
+
+    // Select suitable tenant
+    cy.get('.deliver-tenant-list').select('3');
```
btamas: It should happen only when I would like to publish to remote.

views/js/e2e/_helpers/commands/setupCommands.js
```diff
@@ -83,17 +178,47 @@ Cypress.Commands.add('publishTest', (testName) => {
...
+    }
+    /*
+     * Two different processes to handle different UX
+     * for taoTaskQueue enabled or disabled
+     */
+    cy.get('@isTaskQueueWorking').then(working => {
+        if (working) {
+            cy.wait('@taskQueueWebApi'); // can be 1-4 of them :(
+
+            pollTaskQueue({ taskLabel: 'Remote Publication Lookup', status: 'completed' });
```
btamas: `Remote Publication Lookup` won't happen when you do a local publication.

views/js/e2e/_helpers/commands/setupCommands.js
oatymart: I added a 20000ms timeout. If it becomes a widespread problem, we can move that value into the request defaults in `cypress.json`.

views/js/e2e/_helpers/commands/setupCommands.js
oatymart: Now wrapped in `if`.

views/js/e2e/_helpers/commands/setupCommands.js
oatymart: Now wrapped in `if`.

oatymart:
I just created this companion PR https://github.com/oat-sa/tao-e2e-runner/pull/10 to handle the external config for these tests.

I also made a final small change, replacing a hardcoded tenant id with a label passed in from the env var.

PR #1599, opened 2019-09-19
TAO-8933 key naviagtion problem solved

handleman:
Pull request summary
 
Related to: https://oat-sa.atlassian.net/browse/TAO-8933
 
After the previous changes were broken some parts of the keyNavigation:
 
#### Steps to reproduce and How to test
 - Goto the next item, then come back to the previous, you will see error message in the console (*reason*: keyNavigation creates on the page loading with the same unique id, so when you go back - keyNavigation tries to create the same id, *solution*: use `replace: true` parameter )
 - Check TAB button in the test runner: all active blocks should be in the queue.

### related NPM Release : [0.10.3](https://github.com/oat-sa/tao-test-runner-qti-fe/releases/tag/0.10.3)


zagovorichev:
Thank you, @handleman !

PR #1606, opened 2019-09-30
throw an exception if not current item submitted

hutnikau:
In case if submitted response is not response of current item, then test may get into inconsistent state and items may be skipped.

To test it:
1. disable CSRF tokens in `client_lib_config_registry.conf` by adding the following code:
```
'core/request' => [
    'noToken' => true
],
```
2. set `'csrf-token' => false,` in `testRunner.conf.php`
3. launch the same delivery execution twice (in two tabs of the same browser)
4. Press next on first tab
5. Press next on second tab.

Both submssions will send response on the first item. But during second submission current item of test session will be the second and for the second item empty response will be saved in result server.


siwane:
Hi, how can I test it?

hutnikau:
@siwane pushed version bump and added description

PR #1661, opened 2020-01-09
Mee 9/qti package endpoint

bartlomiejmarszal:
Create Service to export qti package for delivery
Create rest endpoint to download qti package  

https://oat-sa.atlassian.net/browse/MS-48

Qti Package is available using get request. It will provide qti package zip file encoded in base64.


actions/class.RestQtiTests.php
```diff
@@ -39,6 +41,29 @@ public function index()
...
+    /**
+     * Method will return qti package encoded in base64 for delivery
+     * @throws common_Exception
+     * @throws common_exception_Error
+     * @throws common_exception_NotImplemented
+     */
+    public function exportQtiPackage()
+    {
+        $params = $this->getPsrRequest()->getQueryParams();
+        if (!$deliveryId = $params[self::PARAM_TEST_URI]) {
```
siwane: shoudl be testId

actions/class.RestQtiTests.php
```diff
@@ -39,6 +41,29 @@ public function index()
...
+            $this->returnFailure(new common_exception_MissingParameter);
+        }
+        if (!$delivery = $this->getResource($deliveryId)) {
```
siwane: Should be test

actions/class.RestQtiTests.php
```diff
@@ -39,6 +41,29 @@ public function index()
...
+            $this->returnFailure(new common_exception_ResourceNotFound);
+        }
+
+        $qtiPackage = $this->getQtiPackageExporter()->exportDeliveryQtiPackage($delivery);
```
siwane: Doing it in the fly should causes http timeout, what do u think about using defferred task?

actions/class.RestQtiTests.php
```diff
@@ -39,6 +41,29 @@ public function index()
...
+
+        $data[self::PARAM_PACKAGE_NAME] = base64_encode(file_get_contents($qtiPackage['path']));
+
+        $this->returnSuccess($data);
```
siwane: This could causes out of memory problem, please prefer stream

doc/swagger.json
```diff
@@ -20,6 +20,66 @@
         }
     ],
     "paths": {
+        "/taoQtiTest/RestQtiTests/exportQtiPackage": {
```
siwane: this swagger is not right at all (method / parameters...)

helpers/QtiPackageExporter.php
```diff
@@ -0,0 +1,55 @@
...
+use taoQtiTest_models_classes_export_TestExport as TestExporter;
+use core_kernel_classes_Resource as RdfResource;
+use tao_helpers_File as FileHelper;
+
+class QtiPackageExporter extends ConfigurableService
+{
+    /**
+     * @param RdfResource $test
+     *
+     * @return mixed|null
```
siwane: mixed?

helpers/QtiPackageExporter.php
```diff
@@ -0,0 +1,55 @@
...
+     * @throws \common_Exception
+     * @throws \common_exception_Error
+     */
+    public function exportDeliveryQtiPackage(RdfResource $test)
+    {
+        $exportRaport = $this->getTestExporter()->export([
```
siwane: typo

helpers/QtiPackageExporter.php
```diff
@@ -0,0 +1,55 @@
...
+            'filename' => Util::normalizePath('qti_package_'),
```
siwane: Could u be ;ore unia to avoid collision

bartlomiejmarszal: This path string is going to be decorated with additional string in nomalizePath method so collision is avoided

actions/class.RestQtiTests.php
```diff
@@ -212,4 +237,12 @@ private function getTestClass()
     {
         return $this->getClassFromRequest(new \core_kernel_classes_Class(TaoOntology::CLASS_URI_TEST));
     }
+
+    /**
+     * @return QtiPackageExporter
+     */
+    private function getQtiPackageExporter()
```
epoxxid: Good work! One thing I'd suggest is maybe it would be better to use return type hints instead of @return here and in other places, if it allowed by target PHP version (did not found version declaration in composer.json). What do you think?

bartlomiejmarszal: I wish I could. Unfortunately because of some issues, I can't use type hinting. AFAIK

actions/class.RestQtiTests.php
```diff
@@ -39,6 +41,34 @@ public function index()
...
 
+    /**
+     * Method will return qti package encoded in base64 for delivery
+     * @throws common_Exception
+     * @throws common_exception_Error
+     * @throws common_exception_NotImplemented
+     */
+    public function exportQtiPackage()
+    {
+        if ($this->getRequestMethod() != Request::HTTP_GET) {
```
epoxxid: Maybe would be safe to use strict comparison here?

test/integration/QtiPackageExportTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $extractExportPath = sys_get_temp_dir() . '/extractedExportPackage';
+        $zip = new \ZipArchive();
+        if ($zip->open($exportTempFile)) {
+            $zip->extractTo($extractExportPath);
+            $zip->close();
+        }
+        $manifestFile = $extractExportPath . '/imsmanifest.xml';
+
+        //Assert zip content
+        $this->assertTrue(file_exists($manifestFile));
```
peetya: Would it be safer to check that all the required file exist?

actions/class.RestQtiTests.php
Haru0: And yoda code?

test/integration/QtiPackageExportTest.php
bartlomiejmarszal: We need to check if the export process was successful and zip file can be decompressed. This is the reason that I check if extracted files contain expected manifest file. The additional check would test if exporter working correctly which I think out of the scope of this test.

test/integration/QtiPackageExportTest.php
```diff
@@ -0,0 +1,116 @@
...
+{
+    public function __construct()
+    {
+    }
+
+    public function returnSuccess($rawData = [], $withMessage = true)
+    {
+        return $rawData;
+    }
+}
```
Haru0: Missing blank line.

siwane:
When I do a call wihtout testUri, I have:
```<br />
<b>Notice</b>:  Undefined index: testUri in <b>/Users/siwane/workspace/tao/package-tao/taoQtiTest/actions/class.RestQtiTests.php</b> on line <b>58</b><br />
{"success":false,"errorCode":0,"errorMsg":"At least one mandatory parameter was required but found missing in your request","version":"3.4.0-sprint119"}```

File: doc/swagger.json [hunk 13]
```diff
@@ -20,6 +20,94 @@
...
+                        }
+                    },
+                    "400": {
+                        "description": "Bad request",
+                        "schema": {
+                            "$ref": "#/definitions/errorModel"
+                        },
+                        "examples": {
+                            "application/json": {
+                                "application/json": {
```
siwane: Could u remove this please :)

siwane:
When I do with with testUri I have:
```<br />
<b>Notice</b>:  Undefined index: classUri in <b>/Users/siwane/workspace/tao/package-tao/taoQtiTest/models/classes/export/class.TestExport.php</b> on line <b>114</b><br />
{"success":true,"data":{"qtiPackage":"UEsDBBQAA```
Do u have PHP warning enabled?

PR #1696, opened 2020-03-03
Fix html entities name when insert on test

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/UDI-40

https://oat-sa.atlassian.net/browse/BOSA-37

**Description**

When you try to add some item, that have a name without English characters like `Torbørg`, to a test, the display name is not properly rendered

**How to test**

- Go to test and create one
- Insert some item which contains HTML entities in it name
- After insert, see if the name is correctly displayed

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -42,7 +42,7 @@ return [
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '35.13.2',
+    'version'     => '35.13.3',
     'author'      => 'Open Assessment Technologies',
     'requires'    => [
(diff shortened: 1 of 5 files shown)


jsconan:
I guess there is nothing to review here as it is a [WIP] PR. In that case, please avoid assigning someone before it is ready for review. You may eventually set the Draft flag when creating the PR.

Also, since you are the author of the PR you must not be the assigned. This is the responsibility of a reviewer to merge the PR, not the author's responsibility.

Regarding the description, there is a weird description. I don't understand how it is possible to have "Torbørg" turning out to "Torbjørg". At least it is ok to have a bad display, like "Torb?rg" or "Torb&amp;oslash;org" or Torb☐rg, but the injection of an extra character?

PR #1712, opened 2020-03-23
Fix/udi 97/have access to children under restricted class

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/UDI-97
depends on oat-sa/tao-core#2444


manifest.php
```diff
@@ -47,12 +47,12 @@
...
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
-    'version'     => '36.3.0',
+    'version'     => '36.4.0',
     'author'      => 'Open Assessment Technologies',
     'requires'    => [
         'taoQtiItem' => '>=20.0.2',
         'taoTests'   => '>=13.2.0',
-        'tao'        => '>=41.1.0',
+        'tao'        => '>=41.5.0',
```
wazelin: You'll probably have to bump it.
The core went beyond this point

PR #1741, opened 2020-04-14
Changed test compilation version to 2.2.

julien-sebire:
Does this resolve the Manual Scoring GapText issue?


Haru0:
Seems we can close it, right @julien-sebire ?

PR #1745, opened 2020-04-20
REF 68 - Offline mode switcher tool

Alroniks:
The task is here https://oat-sa.atlassian.net/browse/REF-68

Changes provide the new script that partially copied from the previous one, but with adjustments that make this script to switcher tool, that able to manage offline mode states between on and off.

## Testing

Еesting consists in verifying that the configuration has returned to its original state after switching the mode on and off.

To turn on offline mode need to run the script:
```
php index.php 'oat\taoQtiTest\scripts\cli\TestRunnerOfflineMode' -m offline
```

To turn off offline mode need to run the script:
```
php index.php 'oat\taoQtiTest\scripts\cli\TestRunnerOfflineMode' -m online
```

If you use Docker, need to perform these actions inside the container with the application.


scripts/cli/TestRunnerOfflineMode.php
```diff
@@ -0,0 +1,271 @@
...
+                        'extension' => null,
+                        'controller' => null,
+                        'action' => 'messages',
+                        'syncActions' => [
+                            'move',
+                            'skip',
+                            'storeTraceData',
+                            'timeout',
+                            'exitTest',
+//                            'getNextItemData', // remove
```
wazelin, line 207: If it's not part of the `disable` action, shouldn't it be removed from `enable` as well, or the other way around?

Alroniks: Yes, sorry, forget about it. I'll remove. Actually enable action set this parameter, but disable should not. But as there is array_merge used, it is enough just to remove this value from the list.

bziondik:
I checked
first, enabled and all works well
then disabled and when I tried to run assessment there was an error
image

scripts/cli/TestRunnerOfflineMode.php
```diff
@@ -0,0 +1,271 @@
...
+
+        if ($e !== null && $d !== null) {
+            throw new ScriptException('Only one action (enable or disable) can be specified per one script call.');
+        }
+
+        if ($e) {
+            return $this->enable();
+        } elseif ($d) {
+            return $this->disable();
+        } else {
```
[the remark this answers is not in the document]
Alroniks: Without it script just gracefully finishes without any notify about parameters. It is not required but increases usability.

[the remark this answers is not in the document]
wazelin: I think @SergiiTao meant the actual `else`, not the body of it ;)

Alroniks:
I simplified the cli interface of the script and added some improvement to avoid CodeClimate swearing.

Alroniks:
@bziondik I fixed issue with a missed provider.

scripts/install/SetOfflineTestRunnerConfig.php
```diff
@@ -24,126 +24,18 @@
...
 namespace oat\taoQtiTest\scripts\install;
 
-use oat\oatbox\AbstractRegistry;
-use oat\oatbox\service\ConfigurableService;
-use oat\taoTests\models\runner\plugins\PluginRegistry;
-use oat\taoTests\models\runner\providers\ProviderRegistry;
-use oat\taoTests\models\runner\providers\TestProvider;
-use oat\tao\model\modules\DynamicModule;
+use common_ext_action_InstallAction;
+use common_report_Report as Report;
```
wazelin, line 28:
```suggestion
use common_report_Report as Report;
use oat\taoQtiTest\scripts\cli\TestRunnerOfflineMode;
```

PR #1742, opened 2020-04-15
Apply French language contributions

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/BSA-97

**Description**

Change translations here:

- https://github.com/oat-sa/extension-tao-testqti/blob/6a4e7598078e99e05c87acab77d43632c572bdb6/views/js/controller/creator/templates/testpart.tpl#L11

image

- Button `New Section` in the test, to create new section part
image

- Remove Test Part

- Move Up

- Move Down

PR #1751, opened 2020-04-22
BOSA 97 - Additional fix to French translations

habemuscode:
**Related to**

https://github.com/oat-sa/extension-tao-testqti/pull/1742

**How to test**

imagen


habemuscode:
Ok, so update its only if change on key, not on value, thanks @bziondik

PR #1759, opened 2020-04-26
Update test-runner component

AnastasiaArcadia:
**Related to**
https://oat-sa.atlassian.net/browse/TCA-483

**Description**
KB actions for All / Unanswered / Flagged tabs are incorrectly announced

**How to test**
1. as a user start a test
2. enable JAWS
3. navigate to All / Unanswered / Flagged tabs
4. set focus to each tab, listen to SR's announcements

It should be no announcement of navigation via Arrow keys

#### Dependencies 

Requires :
 - [ ] https://github.com/oat-sa/tao-test-runner-qti-fe/pull/181


scripts/update/Updater.php
```diff
@@ -2028,7 +2028,7 @@ public function update($initialVersion)
 
             $this->setVersion('37.3.0');
         }
-
+
```
ampaveliev, line 2031: Extra spaces

ampaveliev:
**For the PR assignee**. Make sure that the package-lock file has the correct `@oat-sa/tao-test-runner-qti `version

PR #1767, opened 2020-04-30
npm package version bump

lecosson:
**Related to**
https://oat-sa.atlassian.net/browse/TCA-381

**Description**
test taker wants to have the TTS tool defaulted to on when I begin the test and stay on (unless turn it off) across all items of the test.

**How to test**
1. prepare ACT instance
2. link or replace package with this branch
3. navigate to test with some APIP items
4. check APIP plugin state and position across items

**Related PR's**
https://github.com/oat-sa/tao-test-runner-qti-fe/pull/185


lecosson:
> Seems you forgot to bump the version. Also it'd be more convenient, if you add the git link to your branch instead of version which hasn't been released yet

test example added to jira

PR #1801, opened 2020-05-27
Fix/tca 427/lock existing test session for readonly mode

SashaTAO:
JIRA ticket: https://oat-sa.atlassian.net/browse/TCA-427

**Problem**
Request fails when proctor tries to reactivate terminated test session via UI interface. 
When test session is loaded initially in readOnly mode and we try to load it for modification, it uses the same session but creates a new testSessionStorage with new BinaryAssessmentTestSeeker implementation. Because of that QTI SDK fails trying to look for current position - `AssessmentTestSeeker::seekPosition()` method.

**Solution**
When session was loaded in readOnly mode and after that in scope of the same request access mode changes to write, we need to reload session to get latest updates before locking, and reinitialize new related components (storage and seeker). 

**How to test** (requires https://github.com/oat-sa/extension-tao-delivery/pull/469 for testing)
- start proctored LTI delivery execution (wait for authorization and start the test)
- as a proctor terminate the session
- as a proctor (requires two roles TeachingAssistant,Administrator) try to reactivate the session.


helpers/class.TestSessionStorage.php
```diff
@@ -131,17 +131,14 @@ public function setUserUri($userUri)
...
     {
-        if (empty(static::$session) || static::$session->getSessionId() !== $sessionId) {
-            $this->setLastError(-1);
-            static::$session = parent::retrieve($test, $sessionId);
-            static::$session->setReadOnly($forReadingOnly);
-            if (!$forReadingOnly) {
-                $this->lockSession(static::$session);
-            }
+        if (static::$session && static::$session->getSessionId() === $sessionId) {
+            return static::$session;
```
hutnikau, line 135: if first call of this method will be with `$forReadingOnly = true` the second call with `$forReadingOnly = false` will not lock the session.

models/classes/TestSessionService.php
```diff
@@ -163,12 +163,11 @@ public function getTestSession(DeliveryExecution $deliveryExecution, $forReading
...
         $sessionId = $deliveryExecution->getIdentifier();
 
-        //load session if no such session or access mode changed
-        if (
-            !$this->hasTestSession($sessionId) ||
-            (!$forReadingOnly && self::$cache[$sessionId][self::SESSION_PROPERTY_SESSION]->isReadOnly())
-        ) {
+        if (!$this->hasTestSession($sessionId)) {
             $this->loadSession($deliveryExecution, $forReadingOnly);
+        } elseif ($this->accessChangedToReadOnly($forReadingOnly, $sessionId)) {
```
hutnikau, line 168: Looks like the name of the `accessChangedToReadOnly` is not precise. It should be something like: `accessModeChandedToWrite`

SashaTAO: Fixed.

helpers/class.TestSessionStorage.php
SashaTAO: This is the case which happens when proctor tries to reactivate terminated session. At first session is loaded in read only mode and later when we set state we change more to `readOnly = false`. I tested it, worked fine.

test/unit/TestSessionServiceTest.php
```diff
@@ -47,19 +48,25 @@
 class TestSessionServiceTest extends TestCase
 {
 
-    public function testGetTestSession()
+    public function testGetTestSessionLoadsNewSession()
```
zagovorichev, line 51: I guest that this test is `testGetTestSessionLoadsSameSession`

helpers/class.TestSessionStorage.php
```diff
@@ -179,10 +176,11 @@ public function persist(AssessmentTestSession $assessmentTestSession)
     /**
      * @param AssessmentTestSession $session
      */
-    protected function lockSession(AssessmentTestSession $session)
+    public function lockSession(AssessmentTestSession $session)
     {
         $lock = $this->createLock('AssessmentTestSession_' . $session->getSessionId(), 30);
         $lock->acquire(true);
+        $session->setReadOnly(false);
```
zagovorichev, line 183: why session locking means that it become writeable?

test/unit/TestSessionServiceTest.php
SashaTAO: Fixed

helpers/class.TestSessionStorage.php
SashaTAO: Fixed

helpers/class.TestSessionStorage.php
SashaTAO: Good question. I think it was used to make it more clear from client point of view, you set access mode (readOnly) instead of saying lock/don't lock.

helpers/class.TestSessionStorage.php
```diff
@@ -131,17 +131,14 @@ public function setUserUri($userUri)
...
     public function retrieve(AssessmentTest $test, $sessionId, $forReadingOnly = false)
     {
-        if (empty(static::$session) || static::$session->getSessionId() !== $sessionId) {
-            $this->setLastError(-1);
-            static::$session = parent::retrieve($test, $sessionId);
-            static::$session->setReadOnly($forReadingOnly);
-            if (!$forReadingOnly) {
-                $this->lockSession(static::$session);
-            }
+        if (static::$session && static::$session->getSessionId() === $sessionId) {
```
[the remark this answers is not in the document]
SashaTAO: Good catch. Fixed.

test/unit/TestSessionServiceTest.php
SashaTAO: Fixed

models/classes/TestSessionService.php
```diff
@@ -287,4 +286,14 @@ public function deleteDeliveryExecutionData(DeliveryExecutionDeleteRequest $requ
...
     }
+
+    /**
+     * @param $forReadingOnly
+     * @param string $sessionId
+     * @return bool
+     */
+    private function accessModeChangedToWrite($forReadingOnly, string $sessionId): bool
+    {
+        return !$forReadingOnly && self::$cache[$sessionId][self::SESSION_PROPERTY_SESSION]->isReadOnly();
```
wazelin, line 297: I realize the method is private, and if one uses it they probably know what they're doing, but just to be absolutely safe, I would've done something like the following.
```suggestion
        return $this->hasSession($sessionId)
            && !$forReadingOnly
            && self::$cache[$sessionId][self::SESSION_PROPERTY_SESSION]->isReadOnly();
```

Alternatively, introducing a method, getting the session, or loading it if necessary, may be an option as well.

SashaTAO: At the moment this method is/should be called only if test session exists. But you are right, to be safe it's better to check it as well.

helpers/class.TestSession.php
```diff
@@ -202,6 +207,14 @@ public function isReadOnly()
         return $this->readOnly;
     }
 
+    /**
+     * @return bool
+     */
+    public function isLocked()
+    {
+        return $this->lock instanceof LockInterface;
```
hutnikau, line 215:
```suggestion
        return $this->lock instanceof LockInterface && $this->lock->isAcquired();
```

SashaTAO: Fixed

helpers/class.TestSessionStorage.php
zagovorichev: It is clear, I'm wondering : why the method which named as lockSession can change state of the session, not the lock only, but also it can setReadOnly to false

zagovorichev:
- [ ] New code is respecting code style rules
- [ ] Acceptance criteria are respected
- [ ] Code is self documented

PR #1810, opened 2020-06-02
[BSA-129] Fix: don't fetch all unnecessary privileges 

SergiiTao:
There were extra DB requests, which impacted performance.


models/classes/creator/TreeItemLookup.php
```diff
@@ -72,23 +73,30 @@ public function getItems(
...
+        $resourceService = $this->getSecureResourceService();
 
         $treeData[0]['children'] = array_filter(
             $treeData[0]['children'],
-            static function (array $item) use ($readableResourcesMap): bool {
-                return $item['type'] !== 'instance' || isset($readableResourcesMap[$item['uri']]);
+            static function (array $item) use ($resourceService): bool {
+                try {
+                    if ($item['type'] === 'instance') {
+                        $resourceService->validatePermission($item['uri'], ['READ']);
```
wazelin, line 93: BTW, what do you think of using `PermissionHelper` in order to fetch all the child permissions at once?
In the following manner.
```php
$nodeIds = [];

foreach ($treeData[0]['children'] as $child) {
    if ($child['type'] === 'instance') {
        $nodeIds[] = $child['uri'];
    }
}

$accessibleNodes = $this->getServiceLocator()->get(PermissionHelper::class)
    ->filterByPermission(
        $nodeIds,
        PermissionInterface::RIGHT_READ
    );

$treeData[0]['children'] = array_filter(
    $treeData[0]['children'],
    static function (array $item) use ($accessibleNodes): bool {
        return $item['type'] !== 'instance' || in_array($item['uri'], $accessibleNodes, true);
    }
);
```

models/classes/creator/TreeItemLookup.php
```diff
@@ -72,23 +73,30 @@ public function getItems(
...
 
-    private function filterTreeData(core_kernel_classes_Class $root, array $treeData): array
+    private function filterTreeData(array $treeData): array
     {
         if (empty($treeData) || empty($treeData[0]['children'])) {
             return $treeData;
         }
 
-        /** @noinspection PhpUnhandledExceptionInspection */
-        $readableResourcesMap = $this->getSecureResourceService()->getAllChildren($root);
```
wazelin, line 86: Nice catch! Totally my bad on that one.

models/classes/creator/TreeItemLookup.php
SergiiTao: didn't know about PermissionHelper
yep. it should work faster if we don't need cache.

PR #1816, opened 2020-06-05
TAO-10312 Fix: allow creating a new test under a non ascii labeled class

wazelin:
- [TAO-10312](https://oat-sa.atlassian.net/browse/TAO-10312) chore: clean up `taoQtiTest_models_classes_QtiTestService`'s static analyzer warning causes
- [TAO-10312](https://oat-sa.atlassian.net/browse/TAO-10312) fix: sanitize a test's identifier during its creation by means of QTI SDK's `Format::sanitizeIdentifier()`
- [TAO-10312](https://oat-sa.atlassian.net/browse/TAO-10312) chore(version): bump a fix one

The actual fix is [here](https://github.com/oat-sa/extension-tao-testqti/pull/1816/commits/66df3109af42be36faa17a63b7b7ac40dfaac75d), the rest was mostly about static analyzer warning fixes.

### How to test
- Create a test class (directory) with non-ascii letters in its label
- Try to create a test instance under it

### DIFF
--- a/models/classes/class.QtiTestService.php
+++ b/models/classes/class.QtiTestService.php
@@ -415,12 +415,14 @@ class taoQtiTest_models_classes_QtiTestService extends TestService
 
             foreach ($report as $r) {
                 $data = $r->getData();
-                $overwrittenItemsIds = array_keys($data->overwrittenItems);
 
                 // -- Rollback all items.
                 // 1. Simply delete items that were not involved in overwriting.
                 foreach ($data->newItems as $item) {
-                    if (!$item instanceof MetadataGuardianResource && !in_array($item->getUri(), $overwrittenItemsIds)) {
+                    if (
+                        !$item instanceof MetadataGuardianResource
+                        && !array_key_exists($item->getUri(), $data->overwrittenItems)
+                    ) {
                         common_Logger::d("Rollbacking new item '" . $item->getUri() . "'...");
                         @$itemService->deleteResource($item);
                     }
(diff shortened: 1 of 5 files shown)


wazelin:
> TestSessionServiceTest

image
Works for me, I ran all the unit tests locally.

Could it be that some of the dependencies is not up to date on your instance?

ktsirkou:
> > TestSessionServiceTest
> 
> image
> Works for me, I ran all the unit tests locally.
> 
> Could it be that some of the dependencies is not up to date on your instance?

hmm, here what I get

Screenshot from 2020-06-05 10-01-31

models/classes/class.QtiTestService.php
```diff
@@ -1096,7 +1083,7 @@ public function createContent(core_kernel_classes_Resource $test, $createTestFil
             $doc->documentElement->setAttribute('title', $test->getLabel());
 
             //generate a valid qti identifier
-            $identifier = tao_helpers_Display::textCleaner($test->getLabel(), '*', 32);
+            $identifier = Format::sanitizeIdentifier($test->getLabel());
```
SergiiTao, line 1086: do we have a test for that?

wazelin: No, at least not yet.

Contributions are welcome ;)

SergiiTao: no no. it doesn't work like that :)) 
please add it if it is possible

wazelin: What do you mean, "it doesn't work like that"?

While it is important to have the test, especially since the behavior was buggy, it's more important to deliver the fix, and to work on the new features, requested by our customers. And if I to spend my time, working on a good test for this particular method, I may fail to deliver the rest.

Since we all work as a team, I invite you to contribute (because of the matters mentioned above). And it is your right to refuse and work on something else. But it certainly does work like that.

wazelin: Anyway. It seems like tomorrow I should be able to dedicate some time to that test.

Thanks for raising that concern!

PR #1827, opened 2020-06-12
Feature/tao 10137/export qti test package to file in shared filesystem

SashaTAO:
JIRA ticket: https://oat-sa.atlassian.net/browse/TAO-10137

According to prepared implementation plan we have to be able to publish delivery to remote environment(s) as a separate action on Deliveries page after delivery is created. We publish QTI test package between two environments and recreate delivery on the target environment. To make sure that we publish exactly the same delivery we have to publish exactly the same test. Because of that when delivery is created we have to  export snapshot of origin test from which delivery was created and store it in shared file system to be able to use later for remote publishing.
Added method to export QTI test to specified file path on shared file system. Refactored exporter class to be able to provide all dependencies via constructor.

How to test:
Not possible to test via UI yet.
Call to `exportQtiTestPackageToFile` with existing test uri, file system ID and new file path -> test must be exported to this file.

Requires:
- [ ] https://github.com/oat-sa/tao-core/pull/2543

### DIFF
--- a/actions/class.RestQtiTests.php
+++ b/actions/class.RestQtiTests.php
@@ -66,7 +66,7 @@ class taoQtiTest_actions_RestQtiTests extends AbstractRestQti
             return $this->returnFailure(new common_exception_ResourceNotFound('Resource not found'));
         }
 
-        $qtiPackage = $this->getQtiPackageExporter()->exportDeliveryQtiPackage($test);
+        $qtiPackage = $this->getQtiPackageExporter()->exportDeliveryQtiPackage($test->getUri());
 
         $data[self::PARAM_PACKAGE_NAME] = base64_encode(file_get_contents($qtiPackage['path']));
 
--- a/actions/class.RestQtiTests.php
+++ b/actions/class.RestQtiTests.php
@@ -66,9 +66,9 @@ class taoQtiTest_actions_RestQtiTests extends AbstractRestQti
             return $this->returnFailure(new common_exception_ResourceNotFound('Resource not found'));
         }
 
-        $qtiPackage = $this->getQtiPackageExporter()->exportDeliveryQtiPackage($test->getUri());
+        $exportReport = $this->getQtiPackageExporter()->exportDeliveryQtiPackage($test->getUri());
 
-        $data[self::PARAM_PACKAGE_NAME] = base64_encode(file_get_contents($qtiPackage['path']));
+        $data[self::PARAM_PACKAGE_NAME] = base64_encode(file_get_contents($exportReport->getData()['path']));
 
         return $this->returnSuccess($data);
     }
--- a/actions/class.RestQtiTests.php
+++ b/actions/class.RestQtiTests.php
@@ -250,7 +250,7 @@ class taoQtiTest_actions_RestQtiTests extends AbstractRestQti
     /**
      * @return QtiPackageExporter
      */
-    private function getQtiPackageExporter()
+    private function getQtiPackageExporter(): QtiPackageExporter
     {
         return $this->getServiceLocator()->get(QtiPackageExporter::SERVICE_ID);
     }
--- a/actions/class.RestQtiTests.php
+++ b/actions/class.RestQtiTests.php
@@ -252,6 +252,6 @@ class taoQtiTest_actions_RestQtiTests extends AbstractRestQti
      */
     private function getQtiPackageExporter()
     {
-        return $this->getServiceLocator()->get(QtiPackageExporter::class);
+        return $this->getServiceLocator()->get(QtiPackageExporter::SERVICE_ID);
     }
 }
--- a/helpers/QtiPackageExporter.php
+++ b/helpers/QtiPackageExporter.php
@@ -27,7 +27,6 @@ use oat\oatbox\filesystem\FileSystemService;
 use oat\tao\helpers\FileHelperService;
 use oat\tao\model\service\InjectionAwareService;
 use taoQtiTest_models_classes_export_TestExport22 as TestExporter;
-use core_kernel_classes_Resource as RdfResource;
 
 class QtiPackageExporter extends InjectionAwareService
 {
@@ -54,36 +53,40 @@ class QtiPackageExporter extends InjectionAwareService
     }
 
     /**
-     * @param RdfResource $test
+     * @param string $testUri
      *
      * @return array
      * @throws common_Exception
      * @throws common_exception_Error
      */
-    public function exportDeliveryQtiPackage(RdfResource $test)
+    public function exportDeliveryQtiPackage(string $testUri)
     {
         $exportReport = $this->testExporter->export(
             [
             'filename' => self::QTI_PACKAGE_FILENAME,
-            'instances' => $test->getUri(),
-            'uri' => $test->getUri()
+            'instances' => $testUri,
+            'uri' => $testUri
             ],
             $this->fileHelperService->createTempDir()
         );
+
+        if ($exportReport->containsError()) {
+            throw new common_Exception('QTI Test package export failed.');
+        }
 
         return $exportReport->getData();
     }
 
     /**
-     * @param RdfResource $test
+     * @param string $testUri
      * @param string $fileSystemId
      * @param string $filePath
      * @return File
      * @throws common_Exception
      */
-    public function exportQtiTestPackageToFile(RdfResource $test, string $fileSystemId, string $filePath): File
+    public function exportQtiTestPackageToFile(string $testUri, string $fileSystemId, string $filePath): File
     {
-        $result = $this->exportDeliveryQtiPackage($test);
+        $result = $this->exportDeliveryQtiPackage($testUri);
 
         return $this->moveFileToSharedFileSystem($result['path'], $fileSystemId, $filePath);
     }
--- a/helpers/QtiPackageExporter.php
+++ b/helpers/QtiPackageExporter.php
@@ -63,9 +63,9 @@ class QtiPackageExporter extends InjectionAwareService
     {
         $exportReport = $this->testExporter->export(
             [
-            'filename' => self::QTI_PACKAGE_FILENAME,
-            'instances' => $testUri,
-            'uri' => $testUri
+                'filename' => self::QTI_PACKAGE_FILENAME,
+                'instances' => $testUri,
+                'uri' => $testUri
             ],
             $this->fileHelperService->createTempDir()
         );
@@ -87,6 +87,9 @@ class QtiPackageExporter extends InjectionAwareService
     public function exportQtiTestPackageToFile(string $testUri, string $fileSystemId, string $filePath): File
     {
         $result = $this->exportDeliveryQtiPackage($testUri);
+        if (!isset($result['path']) || !is_string($result['path'])) {
+            throw new common_Exception('Export report does not contain path to exported QTI package: ' . json_encode($result));
+        }
 
         return $this->moveFileToSharedFileSystem($result['path'], $fileSystemId, $filePath);
     }
--- a/helpers/QtiPackageExporter.php
+++ b/helpers/QtiPackageExporter.php
@@ -22,6 +22,7 @@ namespace oat\taoQtiTest\helpers;
 
 use common_Exception;
 use common_exception_Error;
+use common_report_Report;
 use oat\oatbox\filesystem\File;
 use oat\oatbox\filesystem\FileSystemService;
 use oat\tao\helpers\FileHelperService;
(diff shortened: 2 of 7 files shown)


helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
+    public function exportDeliveryQtiPackage(string $testUri)
     {
-        $exportReport = $this->getTestExporter()->export(
+        $exportReport = $this->testExporter->export(
             [
-            'filename' => Util::normalizePath('qti_package_'),
-            'instances' => $test->getUri(),
-            'uri' => $test->getUri()
+                'filename' => self::QTI_PACKAGE_FILENAME,
+                'instances' => $testUri,
```
luisfn, line 68: Do we need this `instances` key? Does it make sense to be returning the `uri` 2 times?

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
+    /** @var FileSystemService */
+    private $fileSystemService;
+
+    /** @var FileHelperService */
+    private $fileHelperService;
+
+    public function __construct(
+        TestExporter $testExporter,
+        FileSystemService $fileSystemService,
+        FileHelperService $fileHelperService
```
Skyld, line 49: I think it would be better if the helper service (which should not exist by the way) would be used by the FileSystemService

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
             ],
-            FileHelper::createTempDir()
+            $this->fileHelperService->createTempDir()
         );
 
+        if ($exportReport->getType() === common_report_Report::TYPE_ERROR) {
+            throw new common_Exception('QTI Test package export failed.');
+        }
+
         return $exportReport->getData();
```
Skyld, line 78: Why not returning with the report itself?

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
     {
-        $exportReport = $this->getTestExporter()->export(
+        $exportReport = $this->testExporter->export(
             [
-            'filename' => Util::normalizePath('qti_package_'),
-            'instances' => $test->getUri(),
-            'uri' => $test->getUri()
+                'filename' => self::QTI_PACKAGE_FILENAME,
+                'instances' => $testUri,
+                'uri' => $testUri
```
Skyld, line 69: Both of them are test uris?

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
+     * @param string $filePath
+     * @return File
+     * @throws common_Exception
+     */
+    public function exportQtiTestPackageToFile(string $testUri, string $fileSystemId, string $filePath): File
+    {
+        $result = $this->exportDeliveryQtiPackage($testUri);
+        if (!isset($result['path']) || !is_string($result['path'])) {
+            throw new common_Exception('Export report does not contain path to exported QTI package: ' . json_encode($result));
+        }
```
luisfn, line 93: Maybe I missed that and it's done in another level, but shouldn't we also verify if path exist?

Skyld, line 93: This validation should happen in `exportDeliveryQtiPackage` method. If you don't have a path here, it should mean the export was unsuccesful.

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
+     * @throws common_Exception
      */
-    protected function getTestExporter()
+    private function moveFileToSharedFileSystem(string $sourceFilePath, string $fileSystemId, string $destinationFilePath): File
     {
-        return new TestExporter();
+        $source = $this->fileHelperService->readFile($sourceFilePath);
+
+        $file = $this->fileSystemService->getDirectory($fileSystemId)
+            ->getFile($destinationFilePath);
```
Skyld, line 110:
```
$this->fileSystemService
    ->getDirectory($fileSystemId)
    ->getFile($destinationFilePath);
```

scripts/install/RegisterQtiPackageExporter.php
```diff
@@ -0,0 +1,48 @@
...
+use taoQtiTest_models_classes_export_TestExport22;
+use oat\oatbox\extension\InstallAction;
+
+class RegisterQtiPackageExporter extends InstallAction
+{
+    public function __invoke($params)
+    {
+        $serviceManager = $this->getServiceManager();
+        $qtiPackageExporter = new QtiPackageExporter(
+            new taoQtiTest_models_classes_export_TestExport22(),
```
Skyld, line 37: What is this class?

test/unit/helpers/QtiPackageExporterTest.php
```diff
@@ -0,0 +1,144 @@
...
+        $this->fileSystemServiceMock->method('getDirectory')
+            ->willReturn($directoryMock);
+
+        $result = $this->subject->exportQtiTestPackageToFile($testUri, $fileSystemId, $filePath);
+
+        self::assertInstanceOf(
+            File::class,
+            $result,
+            'Method must return instance of File in case of successful export.'
+        );
```
Skyld, line 124: You are not verifying if what you receive is a valid file or not or if it is the file what you would expect. This whole test case can be omitted just by adding return type hint to the `exportQtiTestPackageToFile` method.

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
+     * @param string $testUri
      *
      * @return array
-     * @throws \common_Exception
-     * @throws \common_exception_Error
+     * @throws common_Exception
+     * @throws common_exception_Error
      */
-    public function exportDeliveryQtiPackage(RdfResource $test)
+    public function exportDeliveryQtiPackage(string $testUri)
```
Skyld, line 63: Missing return type, and also returning with generic array is not nice

test/unit/helpers/QtiPackageExporterTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+    public function testExportDeliveryQtiPackage_ThrowsExceptionWhenExportFails(): void
+    {
+        $testUri = 'FAKE_TEST_URI';
+
+        $expectedReport = common_report_Report::createFailure('FAKE ERROR MESSAGE');
+        $this->exporterMock->method('export')
+            ->willReturn($expectedReport);
+
+        $this->expectException(common_Exception::class);
```
Skyld, line 71: I would suggest to assert on the exception message as well

test/unit/helpers/QtiPackageExporterTest.php
```diff
@@ -0,0 +1,144 @@
...
+use oat\oatbox\filesystem\Directory;
+use oat\oatbox\filesystem\File;
+use oat\oatbox\filesystem\FileSystemService;
+use oat\tao\helpers\FileHelperService;
+use taoQtiTest_models_classes_export_TestExport22;
+use oat\generis\test\MockObject;
+use oat\generis\test\TestCase;
+use oat\taoQtiTest\helpers\QtiPackageExporter;
+
+class QtiPackageExporterTest extends TestCase
```
Skyld, line 34: Your test cases are not asserting on the expected positive behaviour of your service:
- What is filename?
- What is the filepath?
- What about the file content?

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,99 @@
...
+use oat\oatbox\filesystem\File;
+use oat\oatbox\filesystem\FileSystemService;
+use oat\tao\helpers\FileHelperService;
+use oat\tao\model\service\InjectionAwareService;
 use taoQtiTest_models_classes_export_TestExport22 as TestExporter;
-use core_kernel_classes_Resource as RdfResource;
-use tao_helpers_File as FileHelper;
 
-class QtiPackageExporter extends ConfigurableService
+class QtiPackageExporter extends InjectionAwareService
```
andreluizmachado, line 32: InjectionAwareService: It's good to know about it.

helpers/QtiPackageExporter.php
SashaTAO: I agree with you that it looks as excessive information. But it was there and both keys are used in taoQtiTest_models_classes_export_TestExport::export() method:
https://github.com/oat-sa/extension-tao-testqti/blob/develop/models/classes/export/class.TestExport.php#L76
https://github.com/oat-sa/extension-tao-testqti/blob/develop/models/classes/export/class.TestExport.php#L114

helpers/QtiPackageExporter.php
SashaTAO: I don't know. It looked strange  for me as well.

helpers/QtiPackageExporter.php
SashaTAO: Yes, and both of them are used and required in test export  class. Answered here - https://github.com/oat-sa/extension-tao-testqti/pull/1827#discussion_r441058300

scripts/install/RegisterQtiPackageExporter.php
SashaTAO: It looks like this class extends test export class to add ability to export QTI 2.2 (original PR where it was created - https://github.com/oat-sa/extension-tao-testqti/pull/544) Previously it was instantiated directly in QtiPackageExporter here - https://github.com/oat-sa/extension-tao-testqti/pull/1827/files#diff-4995a42cda3b7d9303f573b95c6d790cL57. I added it as a constructor dependency in this script to be able to write tests.

test/unit/helpers/QtiPackageExporterTest.php
SashaTAO: There is already return type declaration. I had the same feeling about this test case. Removed.

helpers/QtiPackageExporter.php
SashaTAO: Fixed

test/unit/helpers/QtiPackageExporterTest.php
SashaTAO: I can add it, but I don't like assertions on exception messages. When message changes test will file, but it will not affect business logic if exception type is the same.

test/unit/helpers/QtiPackageExporterTest.php
SashaTAO: I think first two cases should be testes in test for FileSystemService. Here if I create a mock for File object it will return filename and filepath which will be configured for this mock. 
We can test that `getDirectory` and `getFile` methods of a mock will be called with expectedvalues. 

Regarding file content - it's generated in another class. For the moment added assertion that `File::put()` method will be called with expected content returned by another mock. But I need to check if local file system can be registered to eliminate all  this logic (at the moment related method is marked as deprecated).

helpers/QtiPackageExporter.php
SashaTAO: Unfortunately it requires custom script to register a class if your dependencies are something  bigger than scalar or simple class `new DependencyClass()`

helpers/QtiPackageExporter.php
SashaTAO: I need to clarify  why the [method to register local file system](https://github.com/oat-sa/generis/blob/develop/common/oatbox/filesystem/FileSystemService.php#L151) with flysystem abstraction is registered as deprecated. If we can use it, we won't need this helper at all.

helpers/QtiPackageExporter.php
SashaTAO: Moved to `exportDeliveryQtiPackage` method.

helpers/QtiPackageExporter.php
SashaTAO: Fixed

helpers/QtiPackageExporter.php
```diff
@@ -20,40 +20,102 @@
...
+        $this->testExporter = $testExporter;
+        $this->fileSystemService = $fileSystemService;
+        $this->fileHelperService = $fileHelperService;
+    }
+
     /**
-     * @param RdfResource $test
+     * @param string $testUri
      *
      * @return array
```
andreluizmachado, line 59: Please fix the PHPDoc

actions/class.RestQtiTests.php
```diff
@@ -252,6 +252,6 @@ private function getTestClass()
      */
     private function getQtiPackageExporter()
```
andreluizmachado, line 253: Is it possible to add return type hints?

helpers/QtiPackageExporter.php
SashaTAO: Updated to return report.

helpers/QtiPackageExporter.php
SashaTAO: Done

actions/class.RestQtiTests.php
SashaTAO: Added

PR #1838, opened 2020-06-17
fix verifyItemPermissions, formatted exception output

AndreyNegrov:
__Related to task:__ https://oat-sa.atlassian.net/browse/INV-173
__Description:__ XML Editor - Item permission

### DIFF
--- a/models/classes/class.QtiTestService.php
+++ b/models/classes/class.QtiTestService.php
@@ -1246,7 +1246,7 @@ class taoQtiTest_models_classes_QtiTestService extends TestService
         foreach ($array['testParts'] ?? [] as $testPart) {
             foreach ($testPart['assessmentSections'] ?? [] as $assessmentSection) {
                 foreach ($assessmentSection['sectionParts'] ?? [] as $item) {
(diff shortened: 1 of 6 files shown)


models/classes/class.QtiTestService.php
```diff
@@ -1246,7 +1246,7 @@ private function verifyItemPermissions(string $json): void
         foreach ($array['testParts'] ?? [] as $testPart) {
             foreach ($testPart['assessmentSections'] ?? [] as $assessmentSection) {
                 foreach ($assessmentSection['sectionParts'] ?? [] as $item) {
-                    if (isset($item['href'], $item['label']) && $item['qti-type'] ?? '' === self::XML_ASSESSMENT_ITEM_REF) {
```
AndreyNegrov, line 1249: this is not necessarily the wrong check
was made to ignore permission checking when importing a test
now when importing a test this method is not used

uncleempty: Sorry, I still don't get why the check for not empty `$item['label']` was removed. Can you explain it more detailed?

AndreyNegrov: This check was done to bypass the permissions check when importing tests. Now this check is not needed, because now when importing tests this function is not used

PR #1844, opened 2020-06-18
Fix NEX 1184 - Selected contrast remains chosen

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/NEX-1184

**Description**

When the user selects any contrast color in Preview and quits Preview mode, this selected color remains when the next Preview mode is opened, however when devices selector is change from default, every new Preview it is back to default.

**Steps to reproduce**

- Open an item (or shared stimulus) in Preview mode
- Change contrast color to 'White on Black'
- Quit Preview mode
- Open another item (or shared stimulus) in Preview mode

**Actual result:** Next Preview mode opened has the contrast color as was set previously.

**Expected result:** Every new opening of Preview mode has contrast color set to default 'Black on white'.


scripts/update/Updater.php
```diff
@@ -2124,7 +2124,7 @@ public function update($initialVersion)
             $this->setVersion('38.13.0');
         }
 
-      $this->skip('38.13.0', '38.14.0');
+      $this->skip('38.13.0', '38.16.1');
```
atsymuk, line 2127: i think now it is prohibited to update this file

habemuscode: done

habemuscode:
@atsymuk NPM package was updated to the next version on merge, no longer needed

bziondik:
> @atsymuk NPM package was updated to the next version on merge, no longer needed

Please, update version in  `manifest.php`

habemuscode:
Done @bziondik

wazelin:
It looks like `@oat-sa/tao-test-runner-qti`'s `lock` entry has already been updated in one of the previous PRs.
So, this PR seems to be doing nothing.

Kindly consider closing it.

PR #1860, opened 2020-07-07
[UNO-399] Fix: error when there are no resources to show

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/UNO-399

If thee didn't contain instances the error appeared.

### DIFF
--- a/models/classes/creator/ListItemLookup.php
+++ b/models/classes/creator/ListItemLookup.php
@@ -64,16 +64,16 @@ class ListItemLookup extends ConfigurableService implements ItemLookup
             $limit
         );
 
+        if (empty($result['nodes'])) {
+            return $result;
+        }
+
         $nodeIds = array_map(
             static function (array $node): string {
                 return $node['uri'];
             },
             $result['nodes']
         );
-
-        if (empty($result['nodes'])) {
-            return $result;
-        }
 
         $accessible = $this->getPermissionHelper()->filterByPermission($nodeIds, PermissionInterface::RIGHT_READ);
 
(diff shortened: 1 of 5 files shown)


models/classes/creator/TreeItemLookup.php
```diff
@@ -91,6 +91,10 @@ private function filterTreeData(array $treeData): array
             }
         }
 
+        if (empty($nodeIds)) {
```
wazelin, line 94: Could you please apply a similar condition to `ListItemLookup` as well?

```
--- a/models/classes/creator/ListItemLookup.php
+++ b/models/classes/creator/ListItemLookup.php
@@ -66,6 +66,10 @@ class ListItemLookup extends ConfigurableService implements ItemLookup
             $limit
         );
 
+        if (empty($result['nodes'])) {
+            return $result;
+        }
+
         $nodeIds = array_map(
             static function (array $node): string {
                 return $node['uri'];
```

models/classes/creator/ListItemLookup.php
```diff
@@ -73,6 +71,10 @@ static function (array $node): string {
             $result['nodes']
         );
 
+        if (empty($result['nodes'])) {
+            return $result;
+        }
+
```
wazelin, line 77: This is better moved above the `array_map` call.

models/classes/creator/TreeItemLookup.php
SergiiTao: it is a bit more rare case. but sure. thanks

models/classes/creator/ListItemLookup.php
SergiiTao: it is true

PR #1865, opened 2020-07-09
Fix/tca 384/disable tools via lti

augustas:
Overridden test category removal by fuzzy matching.
 
Related to : https://oat-sa.atlassian.net/browse/TCA-384
 
Test authors can specify any categories they wish for the test items. f.e. `x-tao-option-line-reader`. Such added category enables Line Reader plugin. Correct option key is in camelCase (`x-tao-option-lineReader`), but frontend is fuzzy matching the strings and the previous version is valid as well. https://github.com/oat-sa/tao-test-runner-qti-fe/blob/0a25a0c8b2895426514ef66037b5ddceb8124836/src/helpers/map.js#L233
This PR modifies how test options are overriden in backend, categories to be removed are found by fuzzy matching as well.
  
#### How to test
 
 - create a test with category `x-tao-option-line-reader` assigned to the section, publish a delivery.
 - launch this test via LTI and verify that line reader tool is enabled
 - launch same test via LTI, with added custom parameter `x_tao_tools={"lineReader":false}`. Verify that line reader tool is disabled
 
```
./vendor/bin/phpunit taoQtiTest/test/unit/models/classes/runner/map/QtiRunnerMapTest.php
```


models/classes/runner/map/QtiRunnerMap.php
```diff
@@ -508,4 +508,23 @@ private function getOverriddenOptionsRepository(): OverriddenOptionsRepositoryIn
         /** @noinspection PhpIncompatibleReturnTypeInspection */
         return $this->getServiceLocator()->get(OverriddenOptionsRepositoryInterface::SERVICE_ID);
     }
+
+    /**
+     * Tool/option activation status in frontend is checked by fuzzy matched categories. This step was added
+     * in order to have the same behavior for tool/option deactivation through overrides
+     */
+    private function removeFuzzyMatchedCategories(array $categoriesMap, string $categoryId): array
```
zagovorichev, line 516: maybe filterCategories, or something else? not sure that fuzzy matched is defined definition

augustas: Fuzzy matched, because it's matched disregarding case/prefix/whitespace/dashes/underscores. The used term is the same as in FE. `filterCategories` name is not specific enough in my opinion. Do you have other ideas?

zagovorichev:
everything works for me, just PHPUnit test

augustas:
> everything works for me, just PHPUnit test

Unit test added

PR #1882, opened 2020-07-30
[TCA-731] Issues with SR and MathItems

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-731
 
Update package.json
  
#### How to test
 
- prepare an instance of TAO with taoAct extension
- prepare a delivery with passage and accessibility mode enabled
- execute delivery as a test taker
- check that test item section has `aria-labelledby` attribute
- check that passage has `aria-label` attribute
 
Requires :
 - [ ] https://github.com/oat-sa/tao-test-runner-qti-fe/pull/290

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -50,7 +50,7 @@ return [
     'label'       => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license'     => 'GPL-2.0',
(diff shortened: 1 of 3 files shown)


atsymuk:
Please double check that package-lock.json properly update after npm release

PR #1900, opened 2020-08-18
Fix/inv 252/qti blocker

AndreyNegrov:
__Related to task:__ https://oat-sa.atlassian.net/browse/INV-252
__Description:__ XML editor not working on save
__What is fixed:__ the lookupClass method has been fixed
classes with the 'Operator' postfix were searched incorrectly

__How to test:__ 
1) need test which in its xml contains a tag 'preCondition' which has conditions 'or' or 'and'
2) go to the Test section, select this test and open the 'XML Editor'
3) click Save
tests xml should be saved successfully

### DIFF
--- a/models/classes/class.QtiTestConverter.php
+++ b/models/classes/class.QtiTestConverter.php
@@ -41,6 +41,12 @@ use qtism\data\View;
  */
 class taoQtiTest_models_classes_QtiTestConverter
 {
+
+    /**
+     * operators for which qtsm classes are postfix
+     *
+     * @var array $operatorClassesOperatorPostfix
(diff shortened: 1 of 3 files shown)


models/classes/class.QtiTestConverter.php
```diff
@@ -460,6 +460,19 @@ private function lookupClass($name)
             'qtism\\data\\rules\\',
             'qtism\\data\\state\\'
         ];
+
+        $operatorClassesPostfix = [
```
wazelin, line 464: The collection is better off defined either as a static variable or a private constant.

PR #1915, opened 2020-09-17
Fix/tao 10532/all added test parts have max attempts set to 1

AndreyNegrov:
**Related to task:**  https://oat-sa.atlassian.net/browse/TAO-10532
**Description:** Test Authoring - All added test parts have Max Attempts set to 1
**How to test:** 
1. Login as admin
2. Go to Tests
3. Create a new test and go to Authoring
4. Add several test parts
**Expected Result:**
all test part Max Attempts value is 0 by default.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -52,7 +52,7 @@ return [
     'label' => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
-    'version' => '39.8.0',
+    'version' => '39.8.1',
     'author' => 'Open Assessment Technologies',
     'requires' => [
         'taoQtiItem' => '>=24.0.0',
--- a/views/js/controller/creator/templates/testpart-props.tpl
+++ b/views/js/controller/creator/templates/testpart-props.tpl
@@ -81,7 +81,7 @@
                     <label for="testpart-max-attempts">{{__ 'Max Attempts'}}</label>
                 </div>
                 <div class="col-6">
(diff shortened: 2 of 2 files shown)


habemuscode (approved):
Code review only

zagovorichev (changes requested):
waiting for the approvement from qti team, not sure, that this value should be changed

zagovorichev:
Here’s some feedback from IMS:

“The effect of navigation actions on item states in nonlinear mode is left as a matter for delivery system implementers, and delivery systems currently vary on this point. In some systems navigating away from an item in non-linear mode terminates the current attempt and returning to the item later starts a new attempt.”

That’s what TAO current Gen currently does.

“In other systems, inter-item navigation in a non-linear test part has no effect on the count of item attempts, and in such systems a candidate may return multiple times, with the same attempt continuing until the user takes an action to submit the item, such as clicking a "submit" button, or making a change to the items and clicking "next". Just visiting an item doesn't usually create a new attempt. It may restart the 'duration' timer etc. but the candidate could navigate to/from the item multiple times within the same attempt.”

That’s what the NextGen Test Runner implements, and it seems pretty easier to deal with explicit response submission, skipping, feedback, etc.

In the first case, it seems unfair to have only one attempt, so as a fix I suggest that we have a mechanism to initialize each new item with customer preferences (i.e. maxAttempts = 0 for unlimited navigation).

zagovorichev:
@AndreyNegrov update the PR , please

AndreyNegrov:
it was fixed in https://github.com/oat-sa/extension-tao-testqti/pull/1916

PR #1928, opened 2020-10-22
Act 483/fix/check if current item exists

hutnikau:
At time of saving item response there is a check that item identifier in response is equal current item in the test session. 

In case if test session is already closed (timeout, terminated by proctor or script etc.) attempt to save item response causes PHP fatal error because `getCurrentAssessmentItemRef` returns `false`:
```
PHP Fatal error: Uncaught Error: Call to a member function getIdentifier() on boolean in /var/www/html/taoplatform/taoQtiTest/models/classes/runner/RunnerParamParserTrait.php:165
Stack trace:
#0 /var/www/html/taoplatform/taoQtiTest/models/classes/runner/synchronisation/action/Move.php(72): oat\taoQtiTest\models\runner\synchronisation\TestRunnerAction->saveItemResponses(false)
#1 /var/www/html/taoplatform/taoQtiTest/models/classes/runner/synchronisation/synchronisationService/ResponseGenerator.php(157): oat\taoQtiTest\models\runner\synchronisation\action\Move->process()
#2 /var/www/html/taoplatform/taoQtiTest/models/classes/runner/synchronisation/SynchronisationService.php(67): oat\taoQtiTest\models\runner\synchronisation\synchronisationService\ResponseGenerator->getActionResponse(Object(oat\taoQtiTest\models\runner\synchronisation\action\Move), 1596114943.7091, 0, Object(oat\taoQtiTest\models\runner\QtiRunnerServiceContext))
```

### DIFF
--- a/models/classes/runner/RunnerParamParserTrait.php
+++ b/models/classes/runner/RunnerParamParserTrait.php
@@ -162,7 +162,8 @@ trait RunnerParamParserTrait
                 ? json_decode($this->getRequestParameter('itemResponse'), true)
                 : null;
 
-            if ($serviceContext->getCurrentAssessmentItemRef()->getIdentifier() !== $itemDefinition) {
+            $currentAssessmentItemRef = $serviceContext->getCurrentAssessmentItemRef();
+            if ($currentAssessmentItemRef && $serviceContext->getCurrentAssessmentItemRef()->getIdentifier() !== $itemDefinition) {
                 throw new QtiRunnerItemResponseException(__('Item response identifier does not match current item'));
             }
 
--- a/manifest.php
+++ b/manifest.php
@@ -53,7 +53,7 @@ return [
     'label' => 'QTI test model',
     'description' => 'TAO QTI test implementation',
     'license' => 'GPL-2.0',
-    'version' => '40.1.3',
+    'version' => '40.1.4',
     'author' => 'Open Assessment Technologies',
     'requires' => [
         'taoQtiItem' => '>=24.0.0',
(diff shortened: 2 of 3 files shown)


Alroniks (approved):

models/classes/runner/RunnerParamParserTrait.php
```diff
@@ -162,7 +162,8 @@ protected function saveItemResponses($emptyAllowed = true)
                 ? json_decode($this->getRequestParameter('itemResponse'), true)
                 : null;
 
-            if ($serviceContext->getCurrentAssessmentItemRef()->getIdentifier() !== $itemDefinition) {
+            $currentAssessmentItemRef = $serviceContext->getCurrentAssessmentItemRef();
+            if ($currentAssessmentItemRef && $serviceContext->getCurrentAssessmentItemRef()->getIdentifier() !== $itemDefinition) {
```
zagovorichev, line 166: I don't like this solution :( it seems that now everyone can change or add responses after the test finished...

zagovorichev (changes requested):

models/classes/runner/RunnerParamParserTrait.php
hutnikau: the check of session state is done earlier, at time of retrieving of testContext:
https://github.com/oat-sa/extension-tao-testqti/blob/develop/models/classes/runner/QtiRunnerService.php#L1322

This `QtiRunnerItemResponseException` was added only to make sure that item response corresponds current item id (we had a case when response was submitted for wrong item.

This fix fixed the case when session was closed after `\oat\taoQtiTest\models\runner\QtiRunnerService::check()` called.

zagovorichev: Understood, thank you

zagovorichev (approved):

PR #1930, opened 2020-10-23
Fix/ga 185/user is asked to validate responses

AndreyNegrov:
**Related to task:**  https://oat-sa.atlassian.net/browse/CGA-185
**Description:** User is asked to Validate Responses even though there is no setting to require it so

### DIFF
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -661,7 +661,7 @@ return [
     'enable-linear-next-item-warning-checkbox' => true,
 
     /**
-     * If 'validate-response' is true, by default system requires a mandatory response
+     * If 'default-validate-response' is true, by default system requires a mandatory response
      * @type boolean
      */
     'default-validate-response' => true,
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -660,4 +660,10 @@ return [
      */
     'enable-linear-next-item-warning-checkbox' => true,
 
+    /**
+     * If 'validate-response' is true, system requires a mandatory response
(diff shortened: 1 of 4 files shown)


helpers/class.TestRunnerUtils.php
```diff
@@ -277,7 +287,9 @@ public static function beginCandidateInteraction(AssessmentTestSession $session)
      */
     public static function doesAllowSkipping(AssessmentTestSession $session)
     {
-        $doesAllowSkipping = true;
+        $runtimeConfig = self::getRuntimeConfig();
```
alexeyvy, line 290: I guess this code should be in the method `doesValidateResponses`

helpers/class.TestRunnerUtils.php
```diff
@@ -65,6 +65,16 @@ public static function getExtendedStateService()
         return self::getServiceManager()->get(ExtendedStateService::SERVICE_ID);
     }
 
+    /**
+     * Get the runtime config
+     * @return array the configuration
+     */
+    protected static function getRuntimeConfig()
```
alexeyvy, line 72: private maybe?

config/default/testRunner.conf.php
```diff
@@ -660,4 +660,10 @@
      */
     'enable-linear-next-item-warning-checkbox' => true,
 
+    /**
+     * If 'validate-response' is true, system requires a mandatory response
```
alexeyvy: please add a comment that this value is just default and can be overridden by a QTI test / section / item setting

alexeyvy (approved):
Code looks good

PR #1947, opened 2020-11-26
Fixed incorrect item duration calculation if user has multi attempts per item

emgolubev:
Related to: https://oat-sa.atlassian.net/browse/TR-101

- changed timer method in `initItemTimer` (start => end, actually it is a partial rollback of https://github.com/oat-sa/extension-tao-testqti/pull/1632)
- replaced `initItemTimer` method by `startItemTimer` in one place

### DIFF
--- a/actions/class.Runner.php
+++ b/actions/class.Runner.php
@@ -608,7 +608,6 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
             $this->saveItemResponses(false);
             $this->saveToolStates();
 
-            $serviceContext->getTestSession()->initItemTimer();
             $result = $this->getRunnerService()->move($serviceContext, $direction, $scope, $ref);
 
             $response = [
--- a/actions/class.Runner.php
+++ b/actions/class.Runner.php
@@ -608,6 +608,7 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
             $this->saveItemResponses(false);
             $this->saveToolStates();
 
+            $serviceContext->getTestSession()->initItemTimer();
             $result = $this->getRunnerService()->move($serviceContext, $direction, $scope, $ref);
 
             $response = [
--- a/models/classes/runner/QtiRunnerService.php
+++ b/models/classes/runner/QtiRunnerService.php
@@ -295,7 +295,7 @@ class QtiRunnerService extends ConfigurableService implements RunnerService
             $session->resume();
         }
 
-        $session->startItemTimer();
+        $session->initItemTimer();
         if ($session->isTimeout() === false) {
             TestRunnerUtils::beginCandidateInteraction($session);
         }
--- a/models/classes/runner/QtiRunnerService.php
+++ b/models/classes/runner/QtiRunnerService.php
@@ -295,7 +295,7 @@ class QtiRunnerService extends ConfigurableService implements RunnerService
             $session->resume();
         }
 
-        $session->initItemTimer();
+        $session->startItemTimer();
         if ($session->isTimeout() === false) {
             TestRunnerUtils::beginCandidateInteraction($session);
         }
--- a/models/classes/runner/session/TestSession.php
+++ b/models/classes/runner/session/TestSession.php
@@ -204,7 +204,7 @@ class TestSession extends taoQtiTest_helpers_TestSession implements UserUriAware
         // try to close existing time range if any, in order to be sure the test will start or restart a new range.
         // if the range is already closed, a message will be added to the log
         $tags = $this->getItemTags($this->getCurrentRouteItem());
-        $this->getTimer()->start($tags, $timestamp)->save();
+        $this->getTimer()->end($tags, $timestamp)->save();
     }
 
     /**
--- a/models/classes/runner/session/TestSession.php
+++ b/models/classes/runner/session/TestSession.php
@@ -190,18 +190,6 @@ class TestSession extends taoQtiTest_helpers_TestSession implements UserUriAware
     }
 
     /**
-     * Initializes the timer for the current item in the TestSession
-     *
-     * @param $timestamp
-     * @deprecated Need to use startItemTimer() method
-     * @throws \oat\taoTests\models\runner\time\InvalidDataException
-     */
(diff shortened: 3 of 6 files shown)


models/classes/runner/session/TestSession.php
```diff
@@ -189,24 +189,6 @@ public function getItemAttemptTag(RouteItem $routeItem)
         return $itemId . '#' . $occurrence . '-' . $itemSession['numAttempts']->getValue();
     }
 
-    /**
-     * Initializes the timer for the current item in the TestSession
-     *
-     * @param $timestamp
-     * @throws \oat\taoTests\models\runner\time\InvalidDataException
-     */
-    public function initItemTimer($timestamp = null)
```
augustas, line 198: I see this method is overridden in `taoAct` extension, https://github.com/oat-sa/extension-tao-act/blob/master/model/runner/ActTestSession.php#L57, this change should address that.

emgolubev: I got back this method for backward compatibility, I suggest to create a tech dept task.

m-lagarde (approved):
- [x] Code is self-documented

tikhanovichA (approved):
- [x] Code is self-documented

models/classes/runner/QtiRunnerService.php
```diff
@@ -295,7 +295,7 @@ public function init(RunnerServiceContext $context)
             $session->resume();
         }
 
-        $session->initItemTimer();
+        $session->startItemTimer();
```
augustas, line 298: `init` method should not start the item timer, `getItem` will do that

models/classes/runner/session/TestSession.php
```diff
@@ -193,18 +193,12 @@ public function getItemAttemptTag(RouteItem $routeItem)
...
     public function initItemTimer($timestamp = null)
     {
-        if (is_null($timestamp)) {
-            $timestamp = microtime(true);
-        }
-
-        // try to close existing time range if any, in order to be sure the test will start or restart a new range.
-        // if the range is already closed, a message will be added to the log
-        $tags = $this->getItemTags($this->getCurrentRouteItem());
-        $this->getTimer()->start($tags, $timestamp)->save();
```
augustas, line 207: The purpose of `initItemTimer` is not to start a timer, but to close a previously opened timer interval in case there is one. As you have already identified, it was modified [here](https://github.com/oat-sa/extension-tao-testqti/pull/1632/files#diff-d55499ceda4a30cf6d0b11c132f9440cc8aa9603c56a387bfed87cf22e6aaf2eR204) and that problem should be addressed as well. I agree the name of the method is not clear.

models/classes/runner/synchronisation/action/Move.php
```diff
@@ -75,7 +75,6 @@ public function process()
                 $this->setOffline();
             }
 
-            $serviceContext->getTestSession()->initItemTimer($this->getTime());
```
augustas, line 78: The purpose of this call is described in this PR, I assume it should remain here: https://github.com/oat-sa/extension-tao-testqti/pull/559

augustas (changes requested):
I think these changes fix the described problem in the ticket, but does not make it clearer and potentially introduces another bug.

actions/class.Runner.php
```diff
@@ -608,7 +608,6 @@ public function move()
             $this->saveItemResponses(false);
             $this->saveToolStates();
 
-            $serviceContext->getTestSession()->initItemTimer();
```
jsconan, line 611: I wonder if the timeline is still correct after this change. The time range has been closed at line [602](https://github.com/oat-sa/extension-tao-testqti/pull/1947/files#diff-fea95e24360d827ee2d0e75d5c4d18bc5d43134b522acb2f205ad18f7be19a67R602). A new time range needs to be started, otherwise, time won't be properly counted, and issues may occur.

Maybe would it be better to apply a similar condition instead:
```php
if (!$this->getRunnerService()->isTerminated($serviceContext)) {
    $serviceContext->getTestSession()->initItemTimer();
}
```

models/classes/runner/QtiRunnerService.php
jsconan: It looks like `startItemTimer()` and `initItemTimer()` are behaving the same.
https://github.com/oat-sa/extension-tao-testqti/blob/develop/models/classes/runner/session/TestSession.php#L215:L222

jsconan (commented):
I'm not sure the proposed fix will solve the issue.
I'd rather suggest investigating the way the item duration is computed. Like here: https://github.com/oat-sa/extension-tao-testqti/blob/develop/models/classes/runner/session/TestSession.php#L279

To track the item duration, a timeline is used. It is composed of a succession of time ranges, defined by a start and an end timestamp, corresponding to each action. Normally, only one time range should exists for an item attempt. However, if the page is refreshed or if intermediate actions are taken, like a heartbeat, several time ranges will be created.
Then, these time ranges are tagged to be linked to particular item, attempt, section, test part or test. I have the feeling the issue is probably on this link. Mabe the tagging is not scoped enough.

augustas (approved):
The related bug seems to be fixed. But I'm not completely sure if linked [original issue](https://github.com/oat-sa/extension-tao-testqti/pull/1632) will not be affected. F.e. if the connection is lost between `move` and `getItem` requests (and `move` did not receive `start=true` query param from FE), then server timer would remain stopped. QA should test in a qa environment based on the config of reporting customer, since the test runner configuration was changed since then. 

tikhanovichA (approved):
**Review Checklist**
- [] New code is covered by tests (if applicable)
- [x] The feature is working correctly on my local machine (if applicable)

m-lagarde (approved):

models/classes/runner/QtiRunnerService.php
jsconan: Oops, this needs to be reverted. With the last change, both methods are not behaving the same.
Actually, the `startItemTimer()` will behave the same as the previous version of `initItemTimer()`. At the end, this PR does not change so much the existing behavior.

PR #1975, opened 2021-02-15
fix: Fixed breaking test, added observers from qti-sdk.

julien-sebire:
Fixed CI tests by adding a missing property in a marshalled assessmentItem in QTI-SDK.


gitromba (approved):
Thanks for the fix @julien-sebire! Tests are green again 😌 🚀

PR #2012, opened 2021-04-30
fix: clean tmp folder after publishing the delivery

voknil:
https://oat-sa.atlassian.net/browse/ADF-237

There were .xml files after publishing the delivery in /tmp folder which contain an info about items and sections in the test.
Also some empty folders (e.g. tmp1335343).

### DIFF
--- a/helpers/QtiPackageExporter.php
+++ b/helpers/QtiPackageExporter.php
@@ -60,15 +60,19 @@ class QtiPackageExporter extends InjectionAwareService
      * @throws common_Exception
      * @throws common_exception_Error
      */
-    public function exportDeliveryQtiPackage(string $testUri): common_report_Report
+    public function exportDeliveryQtiPackage(string $testUri, ?string $pathToTempDir = null): common_report_Report
     {
+        if (null === $pathToTempDir) {
+            $pathToTempDir = $this->fileHelperService->createTempDir();
+        }
+
         $exportReport = $this->testExporter->export(
             [
                 'filename' => self::QTI_PACKAGE_FILENAME,
                 'instances' => $testUri,
                 'uri' => $testUri
             ],
-            $this->fileHelperService->createTempDir()
+            $pathToTempDir
         );
 
         if ($exportReport->getType() === common_report_Report::TYPE_ERROR) {
@@ -92,9 +96,14 @@ class QtiPackageExporter extends InjectionAwareService
      */
     public function exportQtiTestPackageToFile(string $testUri, string $fileSystemId, string $filePath): File
     {
-        $result = $this->exportDeliveryQtiPackage($testUri)->getData();
+        $pathToTempDir = $this->fileHelperService->createTempDir();
+        $result = $this->exportDeliveryQtiPackage($testUri, $pathToTempDir)->getData();
 
-        return $this->moveFileToSharedFileSystem($result['path'], $fileSystemId, $filePath);
+        $file = $this->moveFileToSharedFileSystem($result['path'], $fileSystemId, $filePath);
+
+        $this->fileHelperService->removeDirectory($pathToTempDir);
+
+        return $file;
     }
 
     /**
--- a/models/classes/QtiTestFileCleaner.php
+++ b/models/classes/QtiTestFileCleaner.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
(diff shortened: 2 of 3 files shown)


models/classes/QtiTestFileCleaner.php
```diff
@@ -0,0 +1,78 @@
...
+use oat\oatbox\filesystem\FileSystemService;
+use oat\oatbox\service\ConfigurableService;
+use qtism\data\TestPart;
+
+/**
+ * Cleaner service for cleanup the leftovers.
+ */
+class QtiTestFileCleaner extends ConfigurableService
+{
+    const SERVICE_ID = 'taoQtiTest/QtiTestCleaner';
```
gabrielfs7, line 38: You should not put `SERVICE_ID` if there is no configurations settings. You can remove it. If you keep it, you need to create a configuration file

models/classes/QtiTestFileCleaner.php
```diff
@@ -0,0 +1,78 @@
...
+                    if($childComponentLvl2 instanceof ExtendedAssessmentSection) {
+                        foreach ($childComponentLvl2->getComponents() as $childComponentLvl3) {
+                            if($childComponentLvl3 instanceof ExtendedAssessmentItemRef) {
+                                $filePaths[] = $itemResolver->resolve($childComponentLvl3->getHref());
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
gabrielfs7, line 69: You can reduce the method complexity by reducing the nested calls and creating private methods.

models/classes/QtiTestFileCleaner.php
```diff
@@ -0,0 +1,78 @@
...
+     * @return string[] paths of leftover xml files (e,g. '/tmp.3937826056e0543c6233a83d9964a59d.xml')
+     */
+    private function getDocumentItemsFilePaths(QtiDocument $document, Resolver $itemResolver): array
+    {
+        $filePaths = [];
+        foreach ($document->getDocumentComponent()->getComponents() as $childComponentLvl1) {
+            if($childComponentLvl1 instanceof TestPart) {
+                foreach ($childComponentLvl1->getComponents() as $childComponentLvl2) {
+                    if($childComponentLvl2 instanceof ExtendedAssessmentSection) {
+                        foreach ($childComponentLvl2->getComponents() as $childComponentLvl3) {
```
gabrielfs7, line 61: Are we sure we only have 3 levels?

models/classes/class.QtiTestCompiler.php
```diff
@@ -1097,4 +1098,9 @@ protected function useCssScoping()
     {
         return $this->settingCssScope;
     }
+
+    protected function getQtiTestFileCleaner(): QtiTestFileCleaner
+    {
+        return $this->getServiceLocator()->get(\oat\taoQtiTest\models\QtiTestFileCleaner::SERVICE_ID);
```
gabrielfs7, line 1104: lets import namespace.

models/classes/QtiTestFileCleaner.php
```diff
@@ -0,0 +1,78 @@
...
+use qtism\data\TestPart;
+
+/**
+ * Cleaner service for cleanup the leftovers.
+ */
+class QtiTestFileCleaner extends ConfigurableService
+{
+    const SERVICE_ID = 'taoQtiTest/QtiTestCleaner';
+
+    public function cleanAfterCompilation(QtiDocument $document, Resolver $itemResolver): void
```
gabrielfs7, line 40:
```suggestion
    public function clear(QtiDocument $document, Resolver $itemResolver): void
```

I think it is not needed to explain in the method name that it was used after compilation. Wdyt?

models/classes/QtiTestFileCleaner.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        $fss = $this->getFileSystem();
+
+        $filePaths = $this->getDocumentItemsFilePaths($document, $itemResolver);
+
+        foreach ($filePaths as $filePath) {
+            $fss->getDirectory('/tmp')->getFile($filePath)->delete();
```
gabrielfs7, line 47: We already have the file paths, so no need to specify the '/tmp'. Also we do not want more than one place getting accountable for the paths details. Wdyt?

models/classes/QtiTestFileCleaner.php
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    /**
+     * @return string[] paths of leftover xml files (e,g. '/tmp.3937826056e0543c6233a83d9964a59d.xml')
+     */
+    private function getDocumentItemsFilePaths(QtiDocument $document, Resolver $itemResolver): array
```
gabrielfs7, line 54:
```suggestion
    private function getDocumentFilePaths(QtiDocument $document, Resolver $itemResolver): array
```

You could get paths for media, etc so better to just return file paths to exclude. The comment is not necessary in this case since method is self explainable

helpers/QtiPackageExporter.php
```diff
@@ -92,9 +96,14 @@ public function exportDeliveryQtiPackage(string $testUri): common_report_Report
      */
     public function exportQtiTestPackageToFile(string $testUri, string $fileSystemId, string $filePath): File
     {
-        $result = $this->exportDeliveryQtiPackage($testUri)->getData();
+        $pathToTempDir = $this->fileHelperService->createTempDir();
```
gabrielfs7, line 99: Why removing the tmp dir here if we can remove on the method `exportDeliveryQtiPackage`? 

We also do not need the extra parameter `pathToTempDir`

helpers/QtiPackageExporter.php
```diff
@@ -60,15 +60,19 @@ public function __construct(
      * @throws common_Exception
      * @throws common_exception_Error
      */
-    public function exportDeliveryQtiPackage(string $testUri): common_report_Report
+    public function exportDeliveryQtiPackage(string $testUri, ?string $pathToTempDir = null): common_report_Report
```
gabrielfs7, line 63: Why the extra param?

PR #2030, opened 2021-06-08
fix: TR-858 settings menu separators issue due test running

lecosson:
**Related to:** https://oat-sa.atlassian.net/browse/TR-858

**Changes:**
- update req's for extension

**Requires:**
 - none

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
     "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
     "qtism/qtism": "~0",
     "oat-sa/generis" : ">=14.0.0",
-    "oat-sa/tao-core" : ">=47.0.0",
+    "oat-sa/tao-core" : ">=v48.11.0",
     "oat-sa/extension-tao-item" : ">=11.0.0",
     "oat-sa/extension-tao-itemqti" : ">=27.0.0",
     "oat-sa/extension-tao-test" : ">=15.0.0",
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
     "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
     "qtism/qtism": "~0",
     "oat-sa/generis" : ">=14.0.0",
-    "oat-sa/tao-core" : ">=v48.11.0",
+    "oat-sa/tao-core" : ">=48.9.0",
     "oat-sa/extension-tao-item" : ">=11.0.0",
     "oat-sa/extension-tao-itemqti" : ">=27.0.0",
     "oat-sa/extension-tao-test" : ">=15.0.0",
--- a/views/js/controller/runner/runner.js
+++ b/views/js/controller/runner/runner.js
@@ -228,7 +228,7 @@ define([
 
                         //FIXME this event should not be triggered on the test runner
                         .on('disablefeedbackalerts', function() {
-                            if (errorFeedback) {menuItemSelector
+                            if (errorFeedback) {
                                 errorFeedback.close();
                             }
                             preventFeedback = true;
--- a/views/js/controller/runner/runner.js
+++ b/views/js/controller/runner/runner.js
@@ -174,11 +174,6 @@ define([
 
             // adding attr for RTL languages
             $('.delivery-scope').attr({dir: locale.getLanguageDirection(context.locale)});
-            // tune classes to remove page header RTL artifacts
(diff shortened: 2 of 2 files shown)


marpesia (approved):
Looks good!

**Review Checklist**
- [x]  Commits messages are meaningful
- [x]  Commits are atomic
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

meefox:
I have this error when running your brunch with LTR language. 
 
Screenshot 2021-06-09 at 18 16 09

views/js/controller/runner/runner.js
```diff
@@ -174,6 +174,11 @@ define([
 
             // adding attr for RTL languages
             $('.delivery-scope').attr({dir: locale.getLanguageDirection(context.locale)});
+            // tune classes to remove page header RTL artifacts
+            const menuItemSelector = '.delivery-scope[dir=rtl] .settings-menu li';
+            $(menuItemSelector).addClass('sep-before');
+            $(`${menuItemSelector}:visible:last`).removeClass('sep-before');
```
jsconan, line 180: **nitpick:** The name is a bit confusing. It should rather highlight the relationship with RTL

**suggestion:** Instead of calling it `menuItemSelector`, better to name it `menuItemSelectorRTL` or something like that.

jsconan (changes requested):
It works, but I wonder if this should not be done directly in the layout upfront, instead of in the client application. With the proposed solution, we can see a visual glitch as the page is always set in LTR while loading, and it stays as it for a few moments, then it finally applies RTL.

The header bar is built by the delivery container, as part of the main layout. A solution could be to work at [this level](https://github.com/oat-sa/extension-tao-delivery/blob/4f7f9773d95bdcf259acf328a1a32169bc5fab55/views/templates/DeliveryServer/blocks/header.tpl#L12).

Another possibility would be to do it through CSS only.

jsconan:
> I have this error when running your brunch with LTR language.
> UPD: and RTL language too :)

@meefox This has been added by a previous PR, on[ tao-core-sdk](https://github.com/oat-sa/tao-core-sdk-fe/pull/109/files). If you have the error, this means you are not up to date. And this also means, the dependency didn't get updated in one of the previous PRs. @lecosson could you update accordingly the dependency to tao-core in the [composer file](https://github.com/oat-sa/extension-tao-testqti/blob/0e707cbc686e6576e5eb8abcc909f6a1849ab980/composer.json#L66)?

views/js/controller/runner/runner.js
meefox: this is smth I want to try as well to use only CSS, if possible.

jsconan: However, the CSS won't remove the visual glitch, as the RTL direction is set by the FE.
A better approach, would be definitely to set it from the BE directly on the template.

This may need some help from BE guys though. Eventually I can help adding the missing part.

composer.json
```diff
@@ -63,7 +63,7 @@
     "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
     "qtism/qtism": "~0",
     "oat-sa/generis" : ">=14.0.0",
-    "oat-sa/tao-core" : ">=47.0.0",
+    "oat-sa/tao-core" : ">=v48.11.0",
```
jsconan, line 66: **nitpick:** While this requirement should not be harmful, it is not correct. The exact version that added the required feature is this one [v48.9.0](https://github.com/oat-sa/tao-core/releases/tag/v48.9.0). I think it should be set accordingly for later reference, so that we will able to have the right picture in several months if we need to figure out the changes made recently.

jsconan (commented):
It could be ok, but I think we must be more cautious about the exact version we require. A more recent version that the one which included the required feature would be fine, but not accurate.

composer.json
meefox: Hey @jsconan should we use v# or just a # in composer json or both are acceptable?

jsconan: both are acceptable, but if possible better only use `#`

jsconan (approved):
I think you can merge this PR as soon as you get enough approvals. This one is about a technical change, so no need to wait for QA.

**Review Checklist**
- [x] Commits are atomic
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

PR #2034, opened 2021-06-10
Feat: submit default responses for non-presented items on test timeout

alexeyvy:
Ticket: https://oat-sa.atlassian.net/browse/TR-1432

Submit default responses for non-presented items on test timeout by imitating skip for all remaining items.

Relies on https://github.com/oat-sa/qti-sdk/pull/282

How to test:
1) Create a test with a global timeout value set in Authoring
2) Create a delivery with that test, you may allow Guest Mode optionally to get to the test quicker
3) Pass the test with timer violation
4) Go to the back-office -> Results -> Choose your delivery -> Check the result. All items should be there. You may also want to check all the other storages we have for test results (e.g. SQL `variable_storage`)

### DIFF
--- a/helpers/class.TestSession.php
+++ b/helpers/class.TestSession.php
@@ -530,14 +530,16 @@ class taoQtiTest_helpers_TestSession extends AssessmentTestSession
     /**
      * Skip the current item.
      *
+     * @param bool $allowUntouchedItems whether to allow skipping items in state INITIAL (not ever seen yet by the TT)
+     *
      * @throws AssessmentTestSessionException If the test session is not running or it is the last route item of the testPart but the SIMULTANEOUS submission mode is in force and not all responses were provided.
      * @qtism-test-interaction
      * @qtism-test-duration-update
      */
-    public function skip()
+    public function skip($allowUntouchedItems = false)
     {
         $sessionMemento = $this->getSessionMemento();
-        parent::skip();
+        parent::skip($allowUntouchedItems);
         $this->triggerEventChange($sessionMemento);
     }
 
--- a/models/classes/runner/QtiRunnerService.php
+++ b/models/classes/runner/QtiRunnerService.php
@@ -1133,6 +1133,18 @@ class QtiRunnerService extends ConfigurableService implements RunnerService
     }
 
     /**
+     * @param RunnerServiceContext $context
+     * @param $scope
+     * @param $ref
+     * @return boolean
+     * @throws \common_Exception
+     */
+    public function autoSkip(RunnerServiceContext $context, $scope, $ref)
+    {
+        return $this->move($context, 'autoSkip', $scope, $ref);
+    }
+
+    /**
      * Handles a test timeout
      * @param RunnerServiceContext $context
      * @param $scope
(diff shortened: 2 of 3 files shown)


models/classes/runner/QtiRunnerService.php
```diff
@@ -1502,6 +1514,11 @@ protected function onTimeout(RunnerServiceContext $context, AssessmentTestSessio
         $isLinear = $session->getCurrentNavigationMode() === NavigationMode::LINEAR;
         switch ($timeOutException->getCode()) {
             case AssessmentTestSessionException::ASSESSMENT_TEST_DURATION_OVERFLOW:
+                while (!$session->getRoute()->isLast()) {
+                    $this->autoSkip($context, 'item', null);
+                    //$this->persist($context);
```
m-lagarde, line 1519: Why this comment?

alexeyvy: that's still in progress, I'm trying to figure out if it's required to call ->persist here or not

helpers/class.TestSession.php
```diff
@@ -530,14 +530,16 @@ public function moveBack($allowTimeout = false)
     /**
      * Skip the current item.
      *
+     * @param bool $allowUntouchedItems whether to allow skipping items in state INITIAL (not ever seen yet by the TT)
```
julien-sebire, line 533: I'm not I get what it means. to skip items in initial mode... Can you clarify the comment?

models/classes/runner/QtiRunnerService.php
```diff
@@ -1132,6 +1132,18 @@ public function skip(RunnerServiceContext $context, $scope, $ref)
...
     }
 
+    /**
+     * @param RunnerServiceContext $context
+     * @param $scope
+     * @param $ref
+     * @return boolean
+     * @throws \common_Exception
+     */
+    public function autoSkip(RunnerServiceContext $context, $scope, $ref)
```
julien-sebire, line 1142: Do you mean by skipping to submit a default value response? That's the equivalent in our implementation because the implementation in CG is faulty, but in NG, skipping doesn't mean that you submit a default value response. Maybe another name is better. What we want to do with this is to submit a default value for non-presented items.

julien-sebire (changes requested):
It looks good. Tests will be welcome.
Can you clarify the name of the parameter please (first comment)?

helpers/class.TestSession.php
alexeyvy: By default, items can be submitted only after they are initialized (started). For auto-skipping, we need to also allow submitting those items not ever started

models/classes/runner/QtiRunnerService.php
alexeyvy: This will submit a default value indeed. However, I suggest we leave the name as it is as it better fits this particular abstraction layer terminology

alexeyvy: ok after thinking over that for a while I think you're right :) we should differentiate the faulty implementation from this functionality. Renamed now `autoSkipWithDefaultResponse`

helpers/class.TestSession.php
julien-sebire: Got it, thanks! :)

julien-sebire (approved):

emgolubev (commented):
@alexeyvy could you add tests?

models/classes/runner/QtiRunnerService.php
m-lagarde: Can we figure it out before the to approve and merge the PR? :)

PR #2057, opened 2021-07-08
feature: TR-892 support RTL items in LTR tests

lecosson:
**Related to:** https://oat-sa.atlassian.net/browse/TR-892

**Changes:**
- using `rtlHandler` plugin to support RTL items in LTR tests
- migration added to activate plugin

**How to check:**
- checkout necessary branch of `taoQtiTest`
- go to views dir and run `npm install` 
- go to root dir of your project and run taoUpdate script to apply migration
- register any language to check as RTL in `config/tao/client_lib_config_registry.conf.php` (en-GB  as example)
```
    'config' => array(
        'util/locale' => array(
            'decimalSeparator' => '.',
            'thousandsSeparator' => '',
            'rtl' => array(
                'ar-arb',
                'en-GB'
            )
        ),
```
- import test with pre-defined RTL language for any item
- walk trought imported test and ensure it rendered as RTL

**Requires:**
 - [ ] https://github.com/oat-sa/tao-test-runner-qti-fe/pull/417

### DIFF
--- a/migrations/Version202107081239332260_taoQtiTest.php
+++ b/migrations/Version202107081239332260_taoQtiTest.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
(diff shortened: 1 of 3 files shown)


jsconan (changes requested):
While this solution will work, I don't think this is the good approach.
Since it is related to the item itself, I should be done at the itemRunner level.

PR #2129, opened 2021-11-19
fix: TR-2486. Align design with last changes

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-2486

How to test: install extenstion and check that code from https://github.com/oat-sa/tao-test-runner-qti-fe/pull/450 is presented


olga-kulish (approved):
Changes are here.

[You may want to revert unnecessary change in 3 map files taoQtiTest.min.js.map, taoQtiTestRunner.min.js.map, taoQtiTestXMLEditor.min.js.map, but it doesn't really matter]

svoronuk:
> Changes are here.
> 
> [You may want to revert unnecessary change in 3 map files taoQtiTest.min.js.map, taoQtiTestRunner.min.js.map, taoQtiTestXMLEditor.min.js.map, but it doesn't really matter]

Actually for me is not clear when i should include rebundled sources. I looked at previous PRs and they are differently looking on this situation, so i've added all just in case.

olga-kulish:
> Actually for me is not clear when i should include rebundled sources. I looked at previous PRs and they are differently looking on this situation, so i've added all just in case.

I meant that only `views/js/loader/testPlugins.min.js` has changes in `*.js` file, so logically only it should have `views/js/loader/testPlugins.min.js.map` updated.
But there are 3 more `*.js.map` files where the only change is absolute part of a path (`"/github/workspace/tao/views/build/config-wrap-end-default.js"],` -> `"/var/www/html/mexcbt-package/tao/views/build/config-wrap-end-default.js"`)

Let's leave them)

oatymart (approved):
Update looks good.

PR #2134, opened 2021-11-29
fix: update SDK dependency

boajer:
https://oat-sa.atlassian.net/browse/TR-2459

deployed at http://test-tr-2459-timeouts.playground.kitchen.it.taocloud.org:49181/

Depends on: 

- [ ] https://github.com/oat-sa/qti-sdk/pull/303


wazelin:
This one's not really needed, as the extension already requires `^0.25.0`, which includes `0.25.1`

PR #2143, opened 2021-12-06
Feature/TR-2846/Manage assets TTL for the caching proxy

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-2846

Requires: 
 - [x] https://github.com/oat-sa/tao-test-runner-qti-fe/pull/454

**Short description**
Implement a TTL management for the cached item store.

**Context**
Assets may have a TTL set at the creation of the URL. This value is defined in the file `config/tao/websource_NNNNNNNN.conf.php`, where `NNNNNNNN` is a random identifier. However, this value is not forwarded to the client in any way.

When the user stays for too long in front of a test, and then reach out an item having assets not yet in the cache, the browser will issue a `404` error. This is especially visible in the context of pre-caching where the assets URL is part of the cached data.

**Description**
Add a config option to define a typical TTL for the item store in the context of the pre-caching feature. This value is set in the config file [`config/taoQtiTest/config/testRunner.conf.php`](https://github.com/oat-sa/extension-tao-testqti/blob/dca05a44c27e989b6d9b0291de023bdfa562a621/config/default/testRunner.conf.php#L660), and the default is `15 minutes`.

This value is forwarded to the client, then the pre-caching proxy applies it and prune the expired items from the cache.

**How to test**
- Have a test with several items, and at least one with a video player (the ticket contain samples that can be imported).
- Have the pre-caching feature activated (look at the doc in [confluence](https://oat-sa.atlassian.net/wiki/spaces/DLV/pages/1681359008/pre-caching)).
- Take the test, then before moving to the item with the player, wait a few minutes (usually 20mn with the default settings)
- Move to the item with the video, and play the video. Without the fix, the browser will issue a 404 and the player will fallback to reload mode. With the fix, the video should be played back.

**Note:** To reduce the amount of time to wait to see the error or to validate it, you can tweak the TTL in the config file `config/tao/websource_NNNNNNNN.conf.php`. And set the same value in the file `config/taoQtiTest/config/testRunner.conf.php` (when the fix is applied). This value is expressed in seconds.

**Note 2:** For the time being, another issue may rise, if the user stays in front of the item with the video, and tries to play it back after a long wait: the browser may issue a 404 if the video content is not yet in the cache.


config/default/testRunner.conf.php
```diff
@@ -652,6 +652,12 @@
      */
     'item-cache-size' => 3,
 
+    /**
+     * Defines the TimeToLive of items in the cache, in seconds
+     * This is required for caching scenarios
+     * @type integer
+     */
+    'item-store-ttl' => 15 * 60,
```
oatymart, line 660: Why must the item validity be so short? What would be the impact of choosing a greater value, e.g. 30 minutes, 2 hours, 8 hours...?

jsconan: I've updated the description.

oatymart: What I really meant was, do you know why the BE uses a short TTL for assets? What is the use case? In production at least, it seems to me that the asset you reference in a compiled test should be invariant, nobody should be changing it.

olga-kulish: Maybe it's short for security? Since if you know url, you can see asset without any auth cookies. Maybe they are afraid people will share this url with others, or just intercept it. But yes, we should ask BE.
I also still think that current TTL is too small. Why not 1 hour for urls, 30min for itemStore?

olga-kulish (approved):
Migration updates config.
Default TTL value seems too small, but since default for backend is 1440 (websource_NNNNNNNN.conf.php contains default?), I guess we can't do better now.

config/default/testRunner.conf.php
jsconan: I think the default value is aligned with the session timeout. And yes, the idea behind the short TTL is to protect the access to the resources. Increasing the value is possible, but it must be aligned with the TTL set for the websource, which is defined on another config file.

Anyway, this is outside of the scope of this ticket. The default value must be lower than the default TTL for the assets. I added recommendation in the documentation related to the pre-caching.

svoronuk (approved):
Awesome work for complex task! For me fix is working. I've tested migration and run test.

PR #2154, opened 2021-12-21
fix: TR-3068. Update depndencies

svoronuk:
Add dependencies with fix for TR-3068.

How to test: ensure, that code is present


oatymart (approved):
Correct code is installed in the extension.

olga-kulish (approved):
Changes are here. I suppose js bundle will be updated by github action

PR #2172, opened 2022-01-12
Feature/TR-2836/Set the client timer as default

jsconan:
### BREAKING CHANGE
**This is considered as a breaking change as every new installation of TAO will be configured with client-side timers.**

### Description
Related to: https://oat-sa.atlassian.net/browse/TR-2836

Change the default configuration for the timer handling to `client`.
This means from new installed TAO, the test runner will process the timer value given by the client instead of only relying on the server controlled timers.

The benefit is the timer countdown will be fine grained, taking care of the client interruptions like page refresh or the navigation between items. It will also prevent the timer to silently continue on the sever when the page is closed. As the timer is also stored locally in the client, the next time the page will open, the timer will be resumed at the latest recorded position.

With the server timer, these kind of interruptions were not taken into account. Especially, when the page is closed, the timer is not paused and continues to decrease on the background, so that the next time the page is open, the timer will have a lesser value or it may be timed out.

**Note:** this PR does not update the configuration for existing installs. To do so, a command-line script is supplied: `php index.php '\oat\taoQtiTest\scripts\cli\TimerMode'`. Without parameters, it will display the current setting.

To set a different timer mode, use the `-s` or `--set` parameter together with the new value, like: `php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' -s client`

### How to test
- Install the extension or run the CLI script to change the mode.
- Take a test having a timer
- Close the page, then reopen it later, the timer should resume at the exact place it was left (a second could have elapsed due to time rounding since the timer is internally managed in milliseconds but displayed in seconds)

### DIFF
--- a/scripts/cli/TimerMode.php
+++ b/scripts/cli/TimerMode.php
@@ -22,27 +22,85 @@ declare(strict_types=1);
 
 namespace oat\taoQtiTest\scripts\cli;
 
+use common_ext_Extension;
+use common_ext_ExtensionsManager;
+use Exception;
+use InvalidArgumentException;
 use oat\oatbox\extension\script\ScriptAction;
 use oat\oatbox\reporting\Report;
 
+/**
+ * Manage the timer-mode option for the TestRunner.
+ * For the time being these modes are supported:
+ * - 'server': The timer is managed by the server. It starts with the test and ends with it.
+ *   Any interruption occurring on the client will be ignored, unless the test is explicitly
+ *   paused by calling the pause action.
+ * - 'client': The timer is managed by the client application. It is synchronized upon each
+ *   navigation move. In case of interruption, the timer is paused and resumed when the item
+ *   is in the interacting state. Time spent outside will be discarded, like while navigating
+ *   between items or waiting for network. Between synchronisation steps the timer value is
+ *   stored in the client storage. If this storage is emptied, the duration elapsed after the
+ *   last synchronisation will be lost. This may happen if the browser's local storage is cleaned,
+ *   or if the test is resumed from a different browser or computer.
+ *
+ * Usage:
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' [options]
+ *
+ *   Optional Arguments:
+ *     -g,       --get        Displays the timer mode
+ *     -s mode, --set mode    Changes the timer mode
+ *     -h,       --help       Prints a help statement
+ *
+ * Examples:
+ * - Displays the current mode:
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' -g
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' --get
+ *
+ * - Change the timer mode to 'client'
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' -s client
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' --set client
+ *
+ * - Displays help:
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' -h
+ *     php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' --help
+ */
 class TimerMode extends ScriptAction
 {
-    public const SET_TIMER_MODE = 'set';
+    public const CONFIG_FILE = 'testRunner';
+    public const GET_TIMER_MODE = 'show';
+    public const SET_TIMER_MODE = 'mode';
     public const SUPPORTED_MODES = [
         'server',
         'client'
     ];
 
+    protected function provideUsage()
+    {
+        return [
+            'prefix' => 'h',
+            'longPrefix' => 'help',
+            'description' => 'Prints a help statement'
+        ];
+    }
+
     protected function provideOptions()
     {
         return [
+            self::GET_TIMER_MODE => [
+                'prefix' => 'g',
+                'longPrefix' => 'get',
+                'cast' => 'string',
+                'flag' => true,
+                'required' => false,
+                'description' => 'Displays the timer mode',
+            ],
             self::SET_TIMER_MODE => [
                 'prefix' => 's',
                 'longPrefix' => 'set',
                 'cast' => 'string',
                 'flag' => false,
                 'required' => false,
-                'description' => 'Set the timer mode',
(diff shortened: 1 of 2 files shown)


btamas (approved):
I cannot validate the PHP code, but it works in my environment.

**Review Checklist**
- [ ] New code is respecting code style rules
- [ ] Acceptance criteria are respected
- [ ] Pull request's target is not `master`
- [ ] Changelog is updated according to changes (if applicable)
- [ ] Documentation is updated according to changes (if applicable)

scripts/cli/TimerMode.php
```diff
@@ -0,0 +1,107 @@
...
+                );
+            }
+        }
+
+        return $report;
+    }
+
+    private function getExtensionsManagerService()
+    {
+        return $this->getServiceLocator()->get(\common_ext_ExtensionsManager::SERVICE_ID);
```
svoronuk, line 81: I'm not sure about scripts, but in my last migration i get extension like this
```
        $taoQtiTestExtension = ExtensionsManager::singleton()->getExtensionById('taoQtiTest');
```

svoronuk (approved):
For me it works. Code looks good. Installation is tested via UI and using command
```
docker exec -u www-data cg-phpfpm php index.php '\oat\taoQtiTest\scripts\cli\TimerMode' -s server
```

scripts/cli/TimerMode.php
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\scripts\cli;
+
+use oat\oatbox\extension\script\ScriptAction;
+use oat\oatbox\reporting\Report;
+
+class TimerMode extends ScriptAction
```
quintanilhar, line 28: What about?
```suggestion
class ChangeTimerMode extends ScriptAction
```

scripts/cli/TimerMode.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        return 'This command sets the test runner timer mode';
+    }
+
+    protected function run()
+    {
+        $report = Report::createInfo(sprintf('The test runner timer mode is currently set to "%s"', $this->getTimerMode()));
+
+        $mode = $this->getOption(self::SET_TIMER_MODE);
+        if ($mode) {
```
quintanilhar, line 60: Early returns are preferable as they decrease the indentation level in the code thus making it more readable.

```
$report = Report::createInfo(sprintf('The test runner timer mode is currently set to "%s"', $this->getTimerMode()));

$mode = $this->getOption(self::SET_TIMER_MODE);

try {

    $this->validate($mode);

    $this->changeMode($mode);
                
    $report->add(
        Report::createSuccess(sprintf('The test runner timer mode was successfully set to "%s"', $mode))
    );
} catch (Exception $exception) {
    $report->add(
        Report::createError(sprintf('Failed to change the timer mode: %s', $exception->getMessage()))
    );
}

return $report;
```

For the validate method:

```
private function validate($mode): void
{
    if (empty($mode)) {
        throw new InvalidArgumentException('The timer mode must be provided');
    }

    if (!in_array($mode, self::SUPPORTED_MODES, true)) {
        throw new InvalidArgumentException('The timer mode must be one of: ' . implode(', ', self::SUPPORTED_MODES));
    }
}
```

The rest for the change method can stay the same.

scripts/cli/TimerMode.php
jsconan: Thanks @quintanilhar!
Initially I set a similar name, but then though it would be more generic since it also allows to know the current setting. Actually, I see it as a TimerMode manager tool. But if you think it is clearer this way, I can update it.

quintanilhar: Perhaps we can use `dry-run` for that. We already have the config itself to see what is the current value. Not sure why do we need this retrieving part and also it's not that to guess that. Ideally, action should do only a single task. We can create another one if really required. What do you think?

scripts/cli/TimerMode.php
```diff
@@ -0,0 +1,107 @@
...
+
+    protected function provideOptions()
+    {
+        return [
+            self::SET_TIMER_MODE => [
+                'prefix' => 's',
+                'longPrefix' => 'set',
+                'cast' => 'string',
+                'flag' => false,
+                'required' => false,
```
quintanilhar, line 44: We can also make it required

scripts/cli/TimerMode.php
```diff
@@ -0,0 +1,107 @@
...
+
+    private function setTimerMode(string $mode): bool
+    {
+        if (in_array($mode, self::SUPPORTED_MODES, true)) {
+
+            /** @var \common_ext_Extension $taoQtiTestExtension */
+            $taoQtiTestExtension = $this->getExtensionsManagerService()->getExtensionById('taoQtiTest');
+            $config = $taoQtiTestExtension->getConfig('testRunner');
+            $config['timer']['target'] = $mode;
+            $taoQtiTestExtension->setConfig('testRunner', $config);
```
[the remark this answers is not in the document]
jsconan: Yes, this would be ok at the condition the `target` entry is the only one possible for the `timer` index, because doing so will overwrite the full entry.

However, `array_replace_recursive` would so the trick. Thx for the hint.

scripts/cli/TimerMode.php
jsconan: Yes, it was the case on the first version, then I decided to also give the possibility to display the current value without dong any change by calling the script without argument.

I'll add a comment to highlight this

scripts/cli/TimerMode.php
jsconan: Not sure a dry-run is needed for such a simple change. But why not.
The reason why I wanted to add a display option is because the timer config is lost in the middle of a big file.

quintanilhar (approved):
👍🏻  Great job man!

- [X] Tests are running successfully on my local machine
- [X] New code is respecting code style rules
- [X] New code is respecting best practices
- [X] Feature is working correctly on my local machine
- [X] Acceptance criteria are respected
- [X] Pull request title and description are meaningful
- [X] Pull request's target is not `master`
- [X] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] Commits messages are meaningful
- [X] Commits are atomic

PR #2185, opened 2022-01-31
Improve Lithuanian translations

alroniks:
The ticket [REL-617](https://oat-sa.atlassian.net/browse/REL-617)


augustas (approved):
Lithuanian translation is correct

PR #2186, opened 2022-02-01
fix/ADF-933: Fix disconnect from delivery session

hectoras:
**Associated Jira issue:** [ADF-933](https://oat-sa.atlassian.net/browse/ADF-933)

### DIFF
--- a/actions/class.Runner.php
+++ b/actions/class.Runner.php
@@ -1099,8 +1099,8 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
     private function setNavigationContextToCommand(NavigationContextAwareInterface $command): void
     {
         $command->setNavigationContext(
-            $this->getRequestParameter('direction') ?? '',
-            $this->getRequestParameter('scope') ?? '',
+            $this->getRequestParameter('direction'),
+            $this->getRequestParameter('scope'),
             $this->getRequestParameter('ref')
         );
     }
--- a/actions/class.Runner.php
+++ b/actions/class.Runner.php
@@ -1100,7 +1100,7 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
     {
         $command->setNavigationContext(
             $this->getRequestParameter('direction') ?? '',
-            $this->getRequestParameter('scope'),
+            $this->getRequestParameter('scope') ?? '',
             $this->getRequestParameter('ref')
         );
     }
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,8 @@
   },
   "minimum-stability": "dev",
   "require": {
+    "php": ">=7.1",
+
     "ext-dom": "*",
     "oat-sa/lib-test-cat": "2.3.7",
     "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
--- a/composer.json
+++ b/composer.json
@@ -58,8 +58,7 @@
   },
   "minimum-stability": "dev",
   "require": {
-    "php": ">=7.1",
-
+    "php": ">=7.2",
     "ext-dom": "*",
     "oat-sa/lib-test-cat": "2.3.7",
     "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
--- a/model/Service/NavigationContextAwareInterface.php
+++ b/model/Service/NavigationContextAwareInterface.php
@@ -26,5 +26,5 @@ namespace oat\taoQtiTest\model\Service;
 
 interface NavigationContextAwareInterface
 {
-    public function setNavigationContext(string $direction, string $scope, ?string $ref): void;
+    public function setNavigationContext(?string $direction, ?string $scope, ?string $ref): void;
 }
--- a/model/Service/NavigationContextAwareTrait.php
+++ b/model/Service/NavigationContextAwareTrait.php
@@ -39,7 +39,7 @@ trait NavigationContextAwareTrait
     {
         $this->direction = $direction ?? '';
         $this->scope = $scope ?? '';
-        $this->ref = $ref ?? null;
+        $this->ref = $ref;
     }
 
     public function getDirection(): string
--- a/model/Service/NavigationContextAwareTrait.php
+++ b/model/Service/NavigationContextAwareTrait.php
@@ -35,11 +35,11 @@ trait NavigationContextAwareTrait
     /** @var string|null*/
     private $ref = null;
 
-    public function setNavigationContext(string $direction, string $scope, ?string $ref): void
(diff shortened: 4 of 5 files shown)


actions/class.Runner.php
```diff
@@ -1100,7 +1100,7 @@ private function setNavigationContextToCommand(NavigationContextAwareInterface $
     {
         $command->setNavigationContext(
             $this->getRequestParameter('direction') ?? '',
-            $this->getRequestParameter('scope'),
+            $this->getRequestParameter('scope') ?? '',
```
hectoras, line 1103: Do we want to make `NavigationContextAwareTrait::setNavigationContext(string $direction, string $scope, ?string $ref): void` to allow `$scope` to be null  instead? That is, changing the signature in the trait instead of fixing the method call (I think that would need changes in the method signatures for getters in `NavigationContextAwareTrait` as well)

jsconan: **thought:** I think both proposals are good. Actually, the direction scope is not mandatory. In the current issue being fixed, the direction is `exit`, and the `scope` is useless since we are exiting from the test. Having it scoped to `item` or any other element would make no sense. With this respect, the solution of having a nullable scope in the signature would be better, of course, as it would prevent a similar issue to occur elsewhere or later. But the current fix looks good enough too.

hectoras: Then I think we can make every parameter nullable, right? And turn them into empty strings inside the trait: 6e57b0ea3f08bbe8ea7ed37b395966ba08020f52

jsconan (approved):
It works as expected. Thanks!

**Review Checklist**
- [x] Commits are atomic
- ~~[ ] Changelog is updated according to changes (if applicable)~~
- ~~[ ] Documentation is updated according to changes (if applicable)~~

actions/class.Runner.php
hectoras: As we are using [nullable parameters](https://www.php.net/manual/en/migration71.new-features.php) I'm adding a require for PHP 7.1 in `composer.json`  (btw other places  such as `ItemContextAwareTrait` were already using them too) :point_right: 5075331589e0e70cca39b6de9e3ecb805977732d

composer.json
```diff
@@ -58,6 +58,8 @@
   },
   "minimum-stability": "dev",
   "require": {
+    "php": ">=7.1",
```
gabrielfs7, line 61: We actually support >=7.2, but indeed other extensions are wrongly saying 7.1

model/Service/NavigationContextAwareInterface.php
```diff
@@ -26,5 +26,5 @@
 
 interface NavigationContextAwareInterface
 {
-    public function setNavigationContext(string $direction, string $scope, ?string $ref): void;
+    public function setNavigationContext(?string $direction, ?string $scope, ?string $ref): void;
```
gabrielfs7, line 29: This is now a breaking change, but it seems the only place implementing it is the `NavigationContextAwareTrait` trait

https://github.com/search?q=org%3Aoat-sa+setNavigationContext%28&type=code

IMO we do not need to release a breaking change in this case.

model/Service/NavigationContextAwareTrait.php
```diff
@@ -35,11 +35,11 @@ trait NavigationContextAwareTrait
...
 
-    public function setNavigationContext(string $direction, string $scope, ?string $ref): void
+    public function setNavigationContext(?string $direction, ?string $scope, ?string $ref): void
     {
-        $this->direction = $direction;
-        $this->scope = $scope;
-        $this->ref = $ref;
+        $this->direction = $direction ?? '';
+        $this->scope = $scope ?? '';
+        $this->ref = $ref ?? null;
```
gabrielfs7, line 42: I found a bit misleading that we allow null, but only `ref` receives null and only `getRef() ?string` has return type null. I think the fix is solving the issue, but may also be causing some integrity issues. We will never know whatever user passed empty or null, might be that for empty we wanted one behavior and for null another, now it will be difficult to check this.

Btw, you do not need this check for $ref.

```suggestion
        $this->direction = $direction ?? '';
        $this->scope = $scope ?? '';
        $this->ref = $ref;
```

gabrielfs7 (approved):
Good job, Héctor! 🚀 

test/unit/model/Service/NavigationContextAwareTraitTest.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\test\unit\model\Service;
+
+use oat\generis\test\TestCase;
+use oat\taoQtiTest\model\Service\NavigationContextAwareTrait;
+
+class NavigationContextAwareTraitTest extends TestCase
```
gabrielfs7, line 28: 🚀  Thanks

model/Service/NavigationContextAwareTrait.php
hectoras: I've applied your change in 7b78cf9690c5c5d116bcf2663b12301489de5ab2 to keep `$ref` as provided in the ctor, thanks!

Right now I'm setting `$scope` and `$direction` to empty strings because if we don't I think we'd need to change the return type for getters in this trait, and I don't want to change that many places in this PR :sweat_smile:

test/unit/model/Service/NavigationContextAwareTraitTest.php
hectoras: :)

composer.json
hectoras: I've raised it as part of this PR because I'm using nullable parameters. Shall I raise the version to 7.2 instead? :)

gabrielfs7: I think it makes sense https://github.com/oat-sa/tao-community/blob/master/composer.json#L6

model/Service/NavigationContextAwareInterface.php
hectoras: I agree. Since we are setting to optional parameters that previously were non-optional, I think we can keep it as non-BC if that's ok for you too

composer.json
hectoras: Done :point_right: e206ec10bac0ad8ca248d95f136e9c27c52cc3e3

AntonNeviarouski:
QA verified

hectoras:
We've reached two approvals and QA has verified the fix works as expected. Therefore, I'm merging the fix into the `develop` branch.

A separate PR has been opened to merge these changes into the February release branch for this extension: https://github.com/oat-sa/extension-tao-testqti/pull/2190

hectoras:
Released as [v44.3.1](https://github.com/oat-sa/extension-tao-testqti/releases/tag/v44.3.1).

PR #2220, opened 2022-03-10
Feature/TR 3449/fizzy review panel item tooltip

olga-kulish:
Related to https://oat-sa.atlassian.net/browse/TR-3449
See ticket comment for TAE and configuration settings

- [x] apply https://github.com/oat-sa/tao-test-runner-qti-fe/pull/472
- [x] apply https://github.com/oat-sa/tao-core/pull/3446


config/default/testRunner.conf.php
```diff
@@ -402,9 +402,16 @@
...
              * @type boolean
              */
-            'displaySectionTitles' => true
+            'displaySectionTitles' => true,
+            /**
+             * Show item title as a tooltip
+             * (Applies only to `fizzy` layout)
+             * @type boolean
+             */
+            'displayItemTooltip' => false
```
jsconan, line 414: **nitpick:** I don't see the related migration script. Was it intended?

oatymart (approved):
Approved based on the companion PRs, just update the package reference.

jsconan (commented):
It works fine.
Just one remark, about the missing migration script for adding the new config in existing envs

config/default/testRunner.conf.php
olga-kulish: There wasn't any for `reviewLayout` and `displaySectionTitles`, so I didn't do it here either. 
Also, I'm afraid that if I add a migration for these 3 settings now, it will override values set by `taoMexcbt` migration. So I think we shouldn't add it.

(Actually, I wonder how config-setter priority is decided for such cases...)

jsconan: The migration scripts are following the dependencies order. So the taoQtiTest migrations would always play before any consumer extension.
The migration is needed here, even if this is to set the default value. It helps to see all the possibilities in the deployed config file, and it also makes sure the config is properly set, with default values, or with custom values. Otherwise, it may cause weird issues (it is maybe not the case for your feature, but this is a global good practice).
It is a pity that it was forgotten by the previous stories, but please do not let any chance for the "broken window effect" to take place. Better to fix issues as soon as possible to prevent bad habits to sit. Otherwise, the next developer will argue that the previous PR did not contain it, so no need to do... :)

migrations/Version202203151417562260_taoQtiTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $extension->setConfig(self::CONFIG_FILE, $config);
+
+    }
+
+    public function down(Schema $schema): void
+    {
+        $extension = $this->getExtension();
+
+        $config = $extension->getConfig(self::CONFIG_FILE);
+        unset($config['plugins']['review']);
```
oatymart, line 41: I don't know if this full `unset` is ok for the `down` migration... the previous config state must have had `reviewLayout` and `displaySectionTitles` as defined by Siarhei, no?

I suggest unsetting only the new `displayItemTooltip` property.

olga-kulish: No, they weren't defined, they are defined in this migration.
Unless you think we shouldn't unset it because install script has set them? But then we should never revert config change in migration

migrations/Version202203151417562260_taoQtiTest.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        $extension = $this->getExtension();
+
+        $config = $extension->getConfig(self::CONFIG_FILE);
+        $config['plugins']['review']['reviewLayout'] = 'default';
+        $config['plugins']['review']['displaySectionTitles'] = true;
+        $config['plugins']['review']['displayItemTooltip'] = false;
```
jsconan, line 31: **nitpick:** Not sure this is safe. Unfortunately, the previous pull requests for the new layout did not enforce the settings. But meanwhile, other extensions might have done it, and this script will overwrite the config.
While this script can fix this miss, it would need to do it using a smarter way: if the config is not set, then the default value must be set, otherwise, it must remain untouched.

**suggestion:** The script should rather do something like:
```suggestion
        // Fix the config as the previous migrations were not adding it.
        // Set the default config if it does not exist.
        if (!array_key_exists('review', $config['plugins'])) {
            $config['plugins']['review'] = [];
        }
        if (!array_key_exists('reviewLayout', $config['plugins']['review'])) {
            $config['plugins']['review']['reviewLayout'] = 'default';
        }
        if (!array_key_exists('displaySectionTitles', $config['plugins']['review'])) {
            $config['plugins']['review']['displaySectionTitles'] = true;
        }
        
        // Set the new config entry
        $config['plugins']['review']['displayItemTooltip'] = false;
```

migrations/Version202203151417562260_taoQtiTest.php
jsconan: **nitpick:** I agree with Martin, only the `displayItemTooltip` should be reverted. The 2 other properties were set before. This migration script should only take care of the current state, and move it to the previous. As explained in the other comment, because the previous pull requests that added the base feature did not take care of it, we now have an inconsistency.

jsconan (changes requested):
The new migration script looks better. However, it is a bit too greedy. There is only one addition from this PR to the config. The migration script can fill the gaps forgotten by the previous PR, but it should not introduce inconsistencies nor reset existing config to default as other changes might have been applied in between (the first implementation is already several months old).

migrations/Version202203151417562260_taoQtiTest.php
olga-kulish: Ok, I'll do that.
(But again, then we should NEVER revert config change.)

migrations/Version202203151417562260_taoQtiTest.php
olga-kulish: Thanks, it's the best, for some reason I didn't think of it

jsconan (approved):
Good job!

**Review Checklist**
- [x] Commits are atomic

PR #2240, opened 2022-04-06
Make popup warning about End of Test translated

alroniks:
Ticket is [REL-667](https://oat-sa.atlassian.net/browse/REL-667)

The scope of the ticket is about "If a TT has "Dutch" or "Lithuanian" as the user interface language", which is actually related to all **29** languages. 

Fixes include:
- [x] updating test runner FE library with proper usage of tokens
- [x] fixing usage of tokens in some files, especially with single quotes
- [ ] updated PO files for all languages, including fixes of wrong tokens and translations for tokens used at the End of Test popup (if the translation exists somehow in the file).
    - [ ] lt-LT
    - [ ] ja-JP
    - [ ] nl-NL
- [ ] compilation JS files with tokens for all languages


alroniks:
I guess it is not actual anymore due to passed time.
New will be created when needed.

PR #2241, opened 2022-04-07
Update translations for popup

alroniks:
Ticket is [REL-667](https://oat-sa.atlassian.net/browse/REL-667)

This fix contains only minimal changes required to fix the issue with a popup about ending the test.

The proper fix requires updating all tokens for all 29 languages, including fixing some issues in the code and placed as a draft [here](https://github.com/oat-sa/extension-tao-testqti/pull/2240)[^1].

The goal of such splitting is to make backporting to existing releases much easier.

[^1]: I will rebase it and rename it because it was the original one prior to this one.


alroniks:
I failed with rebase so I created [a new one](https://github.com/oat-sa/extension-tao-testqti/pull/2243) that has a clear history.

model/Container/TestQtiServiceProvider.php
```diff
@@ -141,5 +146,15 @@ public function __invoke(ContainerConfigurator $configurator): void
                     service(ToolsStateRepositoryInterface::class),
                 ]
             );
+
+        $services
+            ->set(QtiTestRepositoryInterface::class, QtiTestRepository::class)
```
shpran, line 151: [Optional]
Are use sure it's a good idea to use interface name as a key?

Imagine:
We have multiple places, where we are using this repository.
In the future, we create a new class, which implements this interface.
But we need this new class only in one place (we need to keep previous repository for all other places).

In this case, you need to create a new key (you cannot use interface name, as this will override all the places).

```php
// Service Provider
$services->set(QtiTestRepositoryInterface::class, QtiTestRepository::class);
$services->set(AnotherQtiTestRepository::class, AnotherQtiTestRepository::class);

// Class 1
private function getQtiTestRepository(): QtiTestRepositoryInterface
{
    return $this->getPsrContainer()->get(QtiTestRepositoryInterface::class);
}

// Class 2
private function getQtiTestRepository(): QtiTestRepositoryInterface
{
    return $this->getServiceManager()->getContainer()->get(AnotherQtiTestRepository::class);
}
```

It looks a bit strange, because in first case we use interface as a key and in another - class (or something else, because we cannot use the same key).

But if you will use the same class as a key for both cases, you still can use the interface for type-hint and provide different classes which implements it.

```php
// Service Provider
$services->set(QtiTestRepository::class, QtiTestRepository::class);
$services->set(AnotherQtiTestRepository::class, AnotherQtiTestRepository::class);

// Class 1
private function getQtiTestRepository(): QtiTestRepositoryInterface
{
    return $this->getPsrContainer()->get(QtiTestRepository::class);
}

// Class 2
private function getQtiTestRepository(): QtiTestRepositoryInterface
{
    return $this->getServiceManager()->getContainer()->get(AnotherQtiTestRepository::class);
}
```

model/Domain/Model/QtiTest.php
```diff
@@ -0,0 +1,48 @@
...
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\model\Domain\Model;
+
+class QtiTest
+{
+    /** @var string */
+    private $uri;
+
+    /** @var string */
```
shpran, line 30:
```suggestion
    /** @var string|null */
```

model/Infrastructure/QtiTestRepository.php
```diff
@@ -0,0 +1,78 @@
...
+    private $testModelService;
+
+    /** @var Service */
+    private $qtiItemService;
+
+    public function __construct(Ontology $ontology, TestModelService $testModelService, Service $qtiItemService = null)
+    {
+        $this->ontology = $ontology;
+        $this->testModelService = $testModelService;
+        $this->qtiItemService = $qtiItemService ?? Service::singleton();
```
shpran, line 48: [Optional]
You can register this service in DI container via factory and provide it to constructor as a service. Example:
* https://github.com/oat-sa/tao-core/blob/master/models/classes/accessControl/AccessControlServiceProvider.php#L43
* https://github.com/oat-sa/tao-core/blob/master/models/classes/accessControl/AccessControlServiceProvider.php#L50

shpran (approved):

PR #2261, opened 2022-05-07
Backport timer mode swtitc cli tool

poyuki:
backport of https://github.com/oat-sa/extension-tao-testqti/releases/tag/v44.0.0

### DIFF
--- a/config/default/testRunner.conf.php
+++ b/config/default/testRunner.conf.php
@@ -407,9 +407,10 @@ return [
          * This config tells on which TimeLine to rely to compute the assessment test durations.
          * Caution, if the server TimeLine is always filled, the client TimeLine must be explicitly
          * provided by the implementation. If the client TimeLine is missing, the durations will be zeroed.
+         * The default QTI implementation is supplying the item duration on every move.
          * @type string
          */
-        'target' => 'server'
+        'target' => 'client'
     ],
 
     /**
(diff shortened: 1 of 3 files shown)


config/default/testRunner.conf.php
```diff
@@ -407,9 +407,10 @@
          * This config tells on which TimeLine to rely to compute the assessment test durations.
          * Caution, if the server TimeLine is always filled, the client TimeLine must be explicitly
          * provided by the implementation. If the client TimeLine is missing, the durations will be zeroed.
+         * The default QTI implementation is supplying the item duration on every move.
          * @type string
          */
-        'target' => 'server'
```
tikhanovichA, line 412: We should not change the default behaviour for NCCER. The server timer has been used for counting even if the browser or tab is closed. It was the requirement from the NCCER side.
Changing to the client timer will stop timers if a TT will close the tab, which should not happens

tikhanovichA (changes requested):

config/default/testRunner.conf.php
poyuki: That means it's wrong direction. And we need to dig in handling added extra time by server timer in expiration moment

poyuki:
https://github.com/oat-sa/extension-tao-nccer/pull/453#issuecomment-1127324439

PR #2147, opened 2021-12-14
TR-2961 textReader asset post-processing

poyuki:
# [TR-2961](https://oat-sa.atlassian.net/browse/TR-2961)

## Summary
Need to create functionality which allow `CustomInteraction`(TextReader) assets post-processing 
 
## How to test
* update extension by branch
* create `item` with `Text Reader` with image
* create `test` with created in prev step `item`
* publish this `test` as local
* run published delivery and check in browser inspector image `src`

For affecting local `env` without `cdn` need to update `taoItems/models/classes/render/NoneItemReplacement.php`
by next code 
```
class NoneItemReplacement extends ConfigurableService implements ItemAssetsReplacement
{


    public function postProcessAssets($string)
    {
        return 'replaced-' . $string;
    }
}
```
depend on https://github.com/oat-sa/extension-tao-itemqti/pull/1975

PR #2117, opened 2021-11-16
feature: added cloudfrontification for preview / review modes

boajer:
https://oat-sa.atlassian.net/browse/TR-2854

PR #2301, opened 2022-08-23
Fix/lsi 1156/cfication fix for text reader assets

uncleempty:
This PR aims the goal of getting rid of our usual workaround we are applying to each WS instance on the instance start via init scripts. The workaround itself aims at fixing the problem when images inside the textReader PCI are not signed in case CFied item assets are used.
In general, this is the cherry-picked code from #2117 and #2147.
Currently this is "deployed" to integration replica environment and seem to be working. If you want to get creds to this env and exact steps to test - please contact me in pm via slack.


bugalot (approved):
Tested functionally on https://integration-2donline.depp.taocloud.fr with the following DEPP 2D 2022 deliveries:

    - EVA6_MA
   - XEVA4MA_RESO_M25
   - 22X4CEVA6_EDC_MA02
   - 22XEDCEVA2_PRO_MA10
   - PSY10
   - 22LINU_LIT
   
   In terms of the code, I like the fact that it's build only from cherry-picks.

composer.json
```diff
@@ -65,7 +65,7 @@
     "oat-sa/generis" : ">=14.0.0",
     "oat-sa/tao-core" : ">=48.9.0",
```
wazelin, line 66: I can see that one of the original PRs [requires later versions](https://github.com/oat-sa/extension-tao-testqti/pull/2117/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R65-R66) of generis and core.

I suppose that bump was to ensure the DI container features are available.
Could you please double check those dependencies' bump?

wazelin (changes requested):

composer.json
uncleempty: @wazelin, agreed. This indeed was forgotten, as I didn't cherry-pick commits with composer files - too much to merge.
As I can see in november release we, in any case, have the needed versions.
Adding changes to this PR with needed version bump in a minute

wazelin (approved):

PR #2321, opened 2022-10-10
Create dedicated queue for state management tasks

poyuki:
# [TR-4687](https://oat-sa.atlassian.net/browse/TR-4687)

## Summary
Create dedicated queue for state offloading

## Dependency
required for https://github.com/oat-sa/extension-tao-premium/pull/79

## TAE
https://oat-sa.atlassian.net/browse/TR-4687?focusedCommentId=202348

### DIFF
--- a/models/classes/tasks/QtiStateOffload/AbstractQtiStateManipulationTask.php
+++ b/models/classes/tasks/QtiStateOffload/AbstractQtiStateManipulationTask.php
@@ -33,8 +33,7 @@ use oat\tao\model\taskQueue\Task\TaskAwareTrait;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
 
-abstract class AbstractQtiStateManipulationTask extends AbstractAction
-    implements TaskAwareInterface, QueueAssociableInterface
+abstract class AbstractQtiStateManipulationTask extends AbstractAction implements TaskAwareInterface
 {
     use TaskAwareTrait;
 
@@ -42,18 +41,11 @@ abstract class AbstractQtiStateManipulationTask extends AbstractAction
     public const PARAM_CALL_ID_KEY = 'callId';
     public const PARAM_STATE_LABEL_KEY = 'stateLabel';
 
-    public const QUEUE_NAME = 'state_management';
-
     public function __invoke($params): Report
     {
         [$userId, $callId, $stateLabel] = $this->validateParameters($params);
 
         return $this->manipulateState($userId, $callId, $stateLabel);
-    }
-
-    public function getQueueName(array $params = []): string
-    {
-        return self::QUEUE_NAME;
     }
 
     abstract protected function manipulateState(string $userId, string $callId, string $stateLabel): Report;
--- a/models/classes/tasks/QtiStateOffload/AbstractQtiStateManipulationTask.php
+++ b/models/classes/tasks/QtiStateOffload/AbstractQtiStateManipulationTask.php
@@ -33,7 +33,8 @@ use oat\tao\model\taskQueue\Task\TaskAwareTrait;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
 
-abstract class AbstractQtiStateManipulationTask extends AbstractAction implements TaskAwareInterface, QueueAssociableInterface
+abstract class AbstractQtiStateManipulationTask extends AbstractAction
+    implements TaskAwareInterface, QueueAssociableInterface
 {
     use TaskAwareTrait;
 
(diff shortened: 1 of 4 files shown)


kilatib:
@poyuki cannot see were you activate that queue or PR not final?

scripts/install/SetupStateOffloadQueue.php
```diff
@@ -0,0 +1,73 @@
...
+
+        return Report::createSuccess(
+            sprintf(
+                'Setup of %s proceed successfully',
+                AbstractQtiStateManipulationTask::QUEUE_NAME
+            )
+        );
+    }
+
+    private function buildNewQueue(): QueueInterface
```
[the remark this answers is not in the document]
poyuki: as you can see in this method used `$this->propagate` [which set many useful stuff](https://github.com/oat-sa/generis/blob/master/common/oatbox/service/ServiceManagerAwareTrait.php#L89-L112) in object but after this usage we will not have type hint and it's require use ugly inline php dock. But with such usage ugly inline dock not required and also we validate that `propagate` method still return expected object by type hint

poyuki:
@kilatib not sure what you mean under `activate queue`. Queue defined with [InMemoryBroker](https://github.com/oat-sa/extension-tao-testqti/pull/2321/files#diff-a310f4a9fd1c89bbf72dbf7e1689423c98e0278d5a185f7ea948b822dd585389R65) and [initialised](https://github.com/oat-sa/extension-tao-testqti/pull/2321/files#diff-a310f4a9fd1c89bbf72dbf7e1689423c98e0278d5a185f7ea948b822dd585389R48). If you mean running of worker, then InMemory broker doesn't need a worker, task handled 'on fly' right after dispatching. Other specification of queue, IMO, as definition of RdsBroker(or something else) for it and configuration of worker should happen on some other high level

models/classes/tasks/QtiStateOffload/AbstractQtiStateManipulationTask.php
```diff
@@ -27,26 +27,35 @@
...
 use oat\tao\model\state\StateMigration;
+use oat\tao\model\taskQueue\Task\QueueAssociableInterface;
 use oat\tao\model\taskQueue\Task\TaskAwareInterface;
 use oat\tao\model\taskQueue\Task\TaskAwareTrait;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
 
-abstract class AbstractQtiStateManipulationTask extends AbstractAction implements TaskAwareInterface
+abstract class AbstractQtiStateManipulationTask extends AbstractAction
+    implements TaskAwareInterface, QueueAssociableInterface
```
siwane, line 37: `QueueAssociableInterface` does the job here. But in case if we need to change it back, we have to update the code again.
WDYT about using taskqueue config to associate tasks to their queue? the result will be the same, but need to specifically reference targetted tasks.

siwane (commented):
It works locally 👍

models/classes/tasks/QtiStateOffload/AbstractQtiStateManipulationTask.php
poyuki: i like that approach also because it's allow to define queue name once exactly in related abstract class which used as base. But agree, that your proposition is more flexible, will apply it

scripts/install/SetupStateOffloadQueue.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return $this->propagate(
+            new Queue(
+                self::QUEUE_NAME,
+                new InMemoryQueueBroker()
```
siwane, line 74: This broker must be in RDS, purpose of the taskqueue is to run task on worker

siwane (changes requested):

scripts/install/SetupStateOffloadQueue.php
poyuki: that what i discuss with you yesterday. this extension for a moment doesn't contains `taskqueue` extension yet. only all abstractions related to the queues in `tao-core` and i don't think that this extension is right place for such specification. for that purpose i create second PR in premium extension

siwane: yep you are right, my bad 🙏

siwane (approved):

scripts/install/SetupStateOffloadQueue.php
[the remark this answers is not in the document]
wazelin: Just one nitpick here – it's more like `createNewQueue`. Building is an iterative process ;)

No biggie at all, though.

wazelin (approved):

PR #2326, opened 2022-11-03
Enable batch exporting of test metadata for the particular class

alroniks:
Related task is [REL-880](https://oat-sa.atlassian.net/browse/REL-880)

### Plan of work:

- [x] Do refactoring of an existing solution under the current branch
- [x] Fix unit tests to make it pass after refactoring
- [x] [Make batch export of metadata in separated PR, based on this refactoring](https://github.com/oat-sa/extension-tao-testqti/pull/2329)

PR #2340, opened 2022-12-13
Migration which is re-registering QtiPackageExporter helper service

alroniks:
It is related to the refactoring done under #2326 and [REL-880](https://oat-sa.atlassian.net/browse/REL-880)

Migration re-registering class, using the proper class, because it comes from the script via composer autoloading.

### DIFF
--- a/migrations/Version202212131248212260_taoQtiTest.php
+++ b/migrations/Version202212131248212260_taoQtiTest.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
+use oat\taoQtiTest\scripts\install\RegisterQtiPackageExporter;
+
+final class Version202212131248212260_taoQtiTest extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Patch QtiTestExporter config by proper test exporter class';
+    }
+
+    public function up(Schema $schema): void
+    {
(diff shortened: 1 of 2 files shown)


migrations/Version202212131248212260_taoQtiTest.php
```diff
@@ -0,0 +1,27 @@
...
+final class Version202212131248212260_taoQtiTest extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Patch QtiTestExporter config by proper test exporter class';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->runAction(new RegisterQtiPackageExporter);
```
gabrielfs7, line 20: We need also make sure that the change is applied to `config/taoQtiTest/TestModel.conf.php`.  Checking this file I cannot see the change there https://github.com/oat-sa/extension-tao-testqti/blob/master/scripts/install/RegisterQtiPackageExporter.php#L39

alroniks: Because it is already fixed, see [which class it uses](https://github.com/oat-sa/extension-tao-testqti/blob/master/scripts/install/RegisterQtiPackageExporter.php#L30)

alroniks: I would not touch Updater at all. It does not use anymore but keeps the historical data.

gabrielfs7: We should not keep deleted classes in the code base. For historical data we have git hub. We already removed this class from the code 29 days ago, so it makes sense to remove all its usage.

alroniks:
@gabrielfs7 @augustas I fixed Updater script, please, have a look

augustas (approved):

PR #2343, opened 2022-12-20
[LSI-1315] Add export-class support for metadata export

mianchenia:
https://oat-sa.atlassian.net/browse/LSI-1315

Problem: a few times per champing we have transfer all test metadata to customer. For now we have to manually export about 300+ tests and recreate tree structure.

Solution: this PR adds support for export metadata by class. It does minimal changes to get live easier with the simplest solution.

Extra: 
- this changes only for backward support for 2021 release version (DEPP). We got other ticket to make it in better way: https://oat-sa.atlassian.net/browse/REL-880

### DIFF
--- a/models/classes/export/metadata/TestMetadataByClassExportHandler.php
+++ b/models/classes/export/metadata/TestMetadataByClassExportHandler.php
@@ -22,6 +22,7 @@
 namespace oat\taoQtiTest\models\export\metadata;
 
 use common_report_Report as Report;
+use core_kernel_classes_Class;
 use oat\oatbox\event\EventManagerAwareTrait;
 use \oat\taoQtiItem\model\Export\ItemMetadataByClassExportHandler;
 use oat\taoQtiTest\models\event\QtiTestMetadataExportEvent;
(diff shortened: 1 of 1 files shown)


bugalot (approved):
It does the job @mianchenia, good job! Now waiting for a second approval :)

models/classes/export/metadata/TestMetadataByClassExportHandler.php
```diff
@@ -77,4 +86,41 @@ public function export($formValues, $destPath)
...
+        $tmpFiles = [];
+
+        foreach ($classToExport->getInstances(true) as $instance) {
+            $exporterService->setOption(
+                TestExporter::OPTION_FILE_NAME,
+                $instance->getLabel() . '_metadata_' . time() . '.csv'
+            );
+
+            $filePath = $exporterService->export($instance->getUri());
+            $zip->addFile($filePath, $instance->getLabel() . '.csv');
```
alroniks, line 110: As export is possible for any class, in the case when nested classes exist two or more tests with the same label, in the archive will be set only the last file. That might produce issues in runtime and increase the possibility of support requests. That's no good.

mianchenia: Thanks! It makes sence. 
I've updated solution. Now it works fully recursive. This case won't be an issue anymore. Please have a look.

alroniks (approved):
Looks good!

PR #2393, opened 2023-07-06
feat: Delete old assets on test overwrite

hectoras:
**Associated Jira issue:** [PISA25-411](https://oat-sa.atlassian.net/browse/PISA25-411)

When importing a test with the overwrite option, if there are media assets referenced by the items in the test, delete them as well the same way we do for items. If that operation would end up with the former asset classes empty, delete them too.

**Related PRs:**
- https://github.com/oat-sa/extension-tao-mediamanager/pull/478
- https://github.com/oat-sa/extension-tao-item/pull/632

[PISA25-411]: https://oat-sa.atlassian.net/browse/PISA25-411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


models/classes/test/TestDeleter.php
```diff
@@ -0,0 +1,237 @@
...
+
+use core_kernel_classes_Class;
+use core_kernel_classes_Resource;
+use oat\generis\model\data\Ontology;
+use oat\tao\model\media\TaoMediaException;
+use oat\tao\model\media\TaoMediaResolver;
+use oat\tao\model\resources\Exception\ClassDeletionException;
+use oat\tao\model\resources\Exception\PartialClassDeletionException;
+use oat\tao\model\resources\Service\ClassDeleter;
+use oat\taoMediaManager\model\MediaService;
```
gabrielfs7, line 33: We are creating coupling with taoMediaManager extension

models/classes/class.QtiTestService.php
```diff
@@ -15,20 +15,21 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2013-2018 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
- *
+ * Copyright (c) 2013-2023 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
gabrielfs7, line 18: I see we added many refactoring in this file and we have to keep checking code that is not affecting the solution. This makes more difficult to perform the code review.

Can we open a separate PR later for this, if you have time and we are on track with the fix?

Keep in mind we also have to backport it, so less changes the better.

models/classes/test/TestDeleter.php
hectoras: Direct references to media manager from this PR have been rolled back, now we just trigger an integration event.

models/classes/test/TestDeleter.php
```diff
@@ -0,0 +1,237 @@
...
+use oat\taoMediaManager\model\Specification\MediaClassSpecification;
+use oat\taoMediaManager\model\TaoMediaOntology;
+use oat\taoQtiItem\model\qti\parser\ElementReferencesExtractor;
+use oat\taoQtiItem\model\qti\Service;
+use Psr\Log\LoggerInterface;
+use tao_helpers_Uri;
+use taoItems_models_classes_ItemsService;
+use taoTests_models_classes_TestsService;
+
+class TestDeleter
```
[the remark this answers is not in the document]
hectoras: Done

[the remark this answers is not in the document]
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/event/QtiTestDeletedEvent.php
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright (c) 2023 (original work) Open Assessment Technologies SA.
+ */
+
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\models\event;
+
+use oat\oatbox\event\Event;
+
+class QtiTestDeletedEvent implements Event
```
gabrielfs7, line 27: I am not sure if we needed this event as we have `oat\taoTests\models\event\TestRemovedEvent` already.

hectoras: This event carries the URIs for deleted tests, deleted items and related assets. `TestRemovedEvent` only carries the URI for the deleted test.

If we use `TestRemovedEvent`, we would:
- Trigger an event per deleted test (usually a single one, but maybe more if there are more tests with the label used in the import). 
- Use the `ElementReferencesExtractor` (from QtiTest) in MediaManager to get the asset refs (current changes already add a dependency in MM pointing to QtiTest). We would need to do that before actually removing the test and the items (I think once the test is deleted the refs are gone).

The event is currently triggered by taoTest (`deleteTest()` method).

Therefore, we'd need to trigger `TestRemovedEvent` before the test is actually removed in order to get the list of assets from MM, or extend `TestRemovedEvent` to carry the new info and pass that info to taoTest method `deleteTest()` to include it in the event.

I find it easier to add this new event to hold the referenced resources, since we already had logic here to determine the referenced items (items themselves are already deleted by QtiTestService).

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
...
+                $refs = array_merge($refs, $this->deleteItemClass($subClass));
             }
         }
+
+        $this->getEventManager()->trigger(
+            new QtiTestDeletedEvent($deletedTests, $deletedItemClasses, $refs)
+        );
+    }
+
+    private function deleteItemClass(core_kernel_classes_Class $subClass): array
```
gabrielfs7, line 902: This method is not only deleting the item class, but collecting list of resources references. Would not be better to keep the previous implementation and create a method called 'getResourceReferences':

```
    private function getResourceReferences(core_kernel_classes_Resource $item): array
    {
        $qtiItem = $this->getQtiItemService()->getDataItemByRdfItem($rdfItem);
        
        return $this->getElementReferencesExtractor()->extractAll($qtiItem)->getAllReferences();
    }
```

And then you can call this method here https://github.com/oat-sa/extension-tao-testqti/pull/2393/files#diff-05b1f87fb7a793e119a1ba8079639c0bfa12edb96f8e45615c2807361951b09bR893

Isn't' it cleaner and more SOLID? Wdyt?

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
...
+                $deletedTests[] = $testInstance->getUri();
                 $testService->deleteTest($testInstance);
             }
         }
 
         foreach ($itemClass->getSubClasses() as $subClass) {
             if ($subClass->getLabel() === $itemsClassLabel) {
-                $itemTreeService->deleteClass($subClass);
+                $deletedItemClasses[] = $subClass->getUri();
+                $refs = array_merge($refs, $this->deleteItemClass($subClass));
```
gabrielfs7, line 893: Maybe we could keep the previous logic here and just add https://github.com/oat-sa/extension-tao-testqti/pull/2393/files#r1260920766

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
...
         $testService = $this->getTestService();
-        $itemTreeService = $this->getItemTreeService();
+
+        $deletedTests = [];
+        $deletedItemClasses = [];
+        $refs = [];
 
         foreach ($testClass->getInstances() as $testInstance) {
             if ($testInstance->getLabel() === $testLabel) {
+                $deletedTests[] = $testInstance->getUri();
```
rgomeztao, line 887: suspicious, shouldnt you add it to list after deletion and not before? :)

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
...
                 $testService->deleteTest($testInstance);
             }
         }
 
         foreach ($itemClass->getSubClasses() as $subClass) {
             if ($subClass->getLabel() === $itemsClassLabel) {
-                $itemTreeService->deleteClass($subClass);
+                $deletedItemClasses[] = $subClass->getUri();
```
rgomeztao: same as above here :)

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
         core_kernel_classes_Class $itemClass
     ): void {
         $testService = $this->getTestService();
-        $itemTreeService = $this->getItemTreeService();
+
+        $deletedTests = [];
+        $deletedItemClasses = [];
+        $refs = [];
```
rgomeztao: maybe it would be nice to type this arrays, i thinkthe first two are string arrays but the third looks like objects :thinking:

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
...
+            new QtiTestDeletedEvent($deletedTests, $deletedItemClasses, $refs)
+        );
+    }
+
+    private function deleteItemClass(core_kernel_classes_Class $subClass): array
+    {
+        $resourceReferences = [];
+
+        foreach ($subClass->getInstances(true) as $rdfItem) {
+            $qtiItem = $this->getQtiItemService()->getDataItemByRdfItem($rdfItem);
```
rgomeztao: if yorue goig to call the getQtiItemService() multiple times wouldnt it be better to call it just once outside the loop? :)

models/classes/class.QtiTestService.php
```diff
@@ -873,19 +875,46 @@ private function deleteTestsFromClassByLabel(
...
+            $itemReferences = $this->getElementReferencesExtractor()->extractAll($qtiItem);
```
rgomeztao: same here

models/classes/class.QtiTestService.php
```diff
@@ -1454,6 +1483,16 @@ private function getQtiPackageImportPreprocessing()
         return $this->getServiceLocator()->get(QtiPackageImportPreprocessing::SERVICE_ID);
     }
 
+    private function getQtiItemService(): QtiItemService
```
rgomeztao, line 1507: in theory with above comments this method shouldnt be necessary, neither  getElementReferencesExtractor()
As a nitpick, consider dependency injection rather than using a container if possible

models/classes/event/QtiTestDeletedEvent.php
rgomeztao: have you considered a child class from TestRemovedEvent @hectoras ?

models/classes/class.QtiTestService.php
hectoras: Changed: fc7194217

models/classes/class.QtiTestService.php
hectoras: Changed: b7adabb66

For me it feels a bit weird to delete an object and then retrieve one of its values (for example, if we would be using an ORM with a collection that is lazy-loaded, trying to read it after deletion could break). By the way, in this case the deletion does not modify the object we already have in-memory, so it works fine.

models/classes/class.QtiTestService.php
hectoras: Changed: b7adabb66

See previous comment.

models/classes/class.QtiTestService.php
hectoras: I'd too prefer to use DI here, but this being a legacy service makes it harder (and I'm trying to prevent adding to many changes to this file). Some places are already instantiating this as `QtiTestService::singleton()`.

Moreover, having these in separate methods makes it easier to use DI from tests in the future (but I'm also refraining from doing that here to prevent adding too many changes): The code calls the service getters, and in the future the getters may be rewritten to pick an injected service if available, or call the singleton otherwise .

```php
private function getQtiItemService(): QtiItemService
{
    if (!($this->qtiItemService instanceof QtiItemService)) {
        $this->qtiItemService = QtiItemService::singleton();
    }

    return $this->qtiItemService;
}
```

models/classes/event/QtiTestDeletedEvent.php
hectoras: We may do that, but this method carries an array of test URIs, so we'd need to trigger an event per test instead of a single one.

gabrielfs7: We have to keep in mind the confusion that might cause/bring on having multiple similar events for the same thing just because they are called in a different place, etc. 

If it is not a big deal, I would suggest to: 

- add missing parameters for `TestRemovedEvent` (maybe a param called `array $extendedPayload`, so this is extensible). 
- Trigger this event before deleting the test / item classes.

I am not sure adding new `duplicated` event is a solution when we already have one.

If extending the event is too complicated, then create a new event focused on mediaRemoval instead. 

```
class CleanUpMediaEvent
```

That is focused on deletion...

But I would not have 2 events for Test Deletion, cause it will be confused for developers.

models/classes/class.QtiTestService.php
gabrielfs7: @hectoras not needed, but if you are trying to improve that (I agree with it is not necessary), we can easily do that on DI container using `factory` method. Please, do not create a private property with cache :)

models/classes/class.QtiTestService.php
hectoras: Type for `$refs` is inferred from the PHPDoc of `collectResourceReferences()`, and the others from the type of data pushed to them. However, I'm adding explicit type annotations in 012a459ce to make it clearer :ok_hand:

models/classes/class.QtiTestService.php
hectoras: Changed in 87436f8d5 :ok_hand:

gabrielfs7: I could not find any container reference for this `Service`. Are you sure this is working as expected? (calling the `singleton` inside the container cache file).

models/classes/class.QtiTestService.php
hectoras: Note this has been changed since then to first loop over all item classes and tests to collect their URIs, then to trigger a `TestDeletedEvent` (sic) and finally to do the test and item class deletion per se based on the previously collected URIs.

models/classes/class.QtiTestService.php
hectoras: Already changed to get the ItemTree and Test services just once.

models/classes/class.QtiTestService.php
hectoras: Already changed to get the ItemTree and Test services just once.

models/classes/class.QtiTestService.php
hectoras: It is working fine for me

image

I was assuming the LegacyServiceGateway was doing the right thing and calling the `singleton()` method. I'll recheck and (most probably) rollback to use `QtiItemService::singleton()` again

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
         string $itemsClassLabel,
         core_kernel_classes_Resource $testClass,
         core_kernel_classes_Class $itemClass
     ): void {
-        $testService = $this->getTestService();
-        $itemTreeService = $this->getItemTreeService();
+        $itemClassesToDelete = $this->getSubclassesByLabel($itemClass, $itemsClassLabel);
+
+        /** @var string[] $refs */
+        $refs = $this->collectResourceReferences($itemClassesToDelete);
```
gabrielfs7, line 884: You do not need to type hint here with annotations, instead, add annotations to `collectResourceReferences`, otherwise every place calling this method will have to be typehinted.

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
 
+        /** @var string[] $deletedTestsUris */
+        $deletedTestsUris = [];
         foreach ($testClass->getInstances() as $testInstance) {
+            /** @var core_kernel_classes_Resource $testInstance */
```
gabrielfs7, line 889: No need for typehint, the method `>getInstances()` already does it.

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
     ): void {
-        $testService = $this->getTestService();
-        $itemTreeService = $this->getItemTreeService();
+        $itemClassesToDelete = $this->getSubclassesByLabel($itemClass, $itemsClassLabel);
+
+        /** @var string[] $refs */
+        $refs = $this->collectResourceReferences($itemClassesToDelete);
 
+        /** @var string[] $deletedTestsUris */
+        $deletedTestsUris = [];
```
gabrielfs7, line 887: Why type hint local variables? Is it sustainable to do it?

```suggestion
        $deletedTestsUris = [];
```

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
-                $testService->deleteTest($testInstance);
+                $deletedTestsUris[] = $testInstance->getUri();
             }
         }
 
-        foreach ($itemClass->getSubClasses() as $subClass) {
-            if ($subClass->getLabel() === $itemsClassLabel) {
-                $itemTreeService->deleteClass($subClass);
+        /** @var string[] $deletedItemClassesUris */
+        $deletedItemClassesUris = [];
```
gabrielfs7, line 896: Why type hint local variables? Is it sustainable to do it?

```suggestion
        $deletedItemClassesUris = [];
```

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
+        foreach ($itemClassesToDelete as $subClass) {
+            $deletedItemClassesUris[] = $subClass->getUri();
+        }
+
+        // Trigger the deletion event so extensions holding references to
+        // the deleted test or items can remove them first.
+        //
```
gabrielfs7, line 903: I do not recommend to add this type of commenting here. You are basically explaining why you took the decision of to call this method. This should be corrected by renaming the event class to something like `DeleteAssetsEvent` as I mentioned before. Anybody reading this will know this event is deleting assets :)

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
+            new QtiTestsDeletedEvent(
+                $deletedTestsUris,
+                $deletedItemClassesUris,
+                $refs
+            )
+        );
+
+        // Actually delete the items (if this crashes the tests will
+        // hold refs to items that don't exist anymore).
+        //
```
gabrielfs7, line 914: No need to explain, otherwise you have to do it for all the code in the platform, every line without an explanation.

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
+        $itemTreeService = $this->getItemTreeService();
+        foreach ($itemClassesToDelete as $subClass) {
+            $itemTreeService->deleteClass($subClass);
+        }
+
+        // Actually delete the tests, which already hold refs to
+        // non-existing items.
+        //
```
gabrielfs7, line 922: No need to explain, otherwise you have to do it for all the code in the platform, every line without an explanation.

models/classes/class.QtiTestService.php
```diff
@@ -866,24 +869,62 @@ protected function importTest(
...
+        $testService = $this->getTestService();
+        foreach ($testClass->getInstances() as $testInstance) {
+            /** @var core_kernel_classes_Resource $testInstance */
```
gabrielfs7, line 925: No need to typehint. This is already done on `getInstances` method.

gabrielfs7, line 925:
```suggestion
```

models/classes/class.QtiTestService.php
gabrielfs7, line 914:
```suggestion
```

models/classes/class.QtiTestService.php
gabrielfs7, line 922:
```suggestion
```

models/classes/event/QtiTestsDeletedEvent.php
```diff
@@ -0,0 +1,74 @@
...
+    /** @var string[] */
+    private array $testUris;
+
+    /** @var string[] */
+    private array $itemClassesUri;
+
+    /** @var string[] */
+    private array $referencedResources;
+
+    public function __construct(
```
gabrielfs7, line 38: IMO it is better to have open event due nature of communication and flexibility, so I would receive an array as constructor parameter instead. It is easy to extend.

The data keys can be constants in this very class.

```suggestion
    public function __construct(array $data)
```

models/classes/event/QtiTestsDeletedEvent.php
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright (c) 2023 (original work) Open Assessment Technologies SA.
+ */
+
+declare(strict_types=1);
+
+namespace oat\taoQtiTest\models\event;
+
+use oat\oatbox\event\Event;
+
+class QtiTestsDeletedEvent implements Event
```
gabrielfs7, line 27: We need a specific event for that to avoid confusion with existing TestRemovedEvent. 

```suggestion
class DeleteAssetsEvent implements Event
```

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks! It is all good!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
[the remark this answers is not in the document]
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
gabrielfs7: Thanks!

gabrielfs7: Thanks!

models/classes/class.QtiTestService.php
```diff
@@ -1446,21 +1453,33 @@ private function verifyItemPermissions(core_kernel_classes_Resource $oldTest, st
         $this->getSecureResourceService()->validatePermissions($ids, ['READ']);
     }
 
-    /**
-     * @return QtiPackageImportPreprocessing
-     */
-    private function getQtiPackageImportPreprocessing()
+    private function getSubclassesByLabel(core_kernel_classes_Class $root, string $label)
```
gabrielfs7, line 1456: You can **apply the deletion directly here** instead of having to iterate 2x by the resources here 
https://github.com/oat-sa/extension-tao-testqti/pull/2393/files#diff-05b1f87fb7a793e119a1ba8079639c0bfa12edb96f8e45615c2807361951b09bR881

If you do that, method name needs to be changed

models/classes/event/QtiTestsDeletedEvent.php
hectoras: This event has been already deleted

hectoras:
I'm now getting an error each time I overwrite an asset though the API:

```
[2023-07-14 11:19:11] UDEMO.INFO: Assets "["https:\/\/pisa25-411-delete-old-assets-on-upload.docker.localhost\/ontologies\/tao.rdf#i64b112dc2090d85745707ce5a81032639"]" removed after Item "https://pisa25-411-delete-old-assets-on-upload.docker.localhost/ontologies/tao.rdf#i64b112d8a702085745875cdb5ad6d12ab" using them was removed  [] {"uid":"9d0b0dd85c0861c0b0587edf"}
[2023-07-14 11:19:11] UDEMO.ERROR: core_kernel_classes_EmptyProperty exception deleting "https://pisa25-411-delete-old-assets-on-upload.docker.localhost/ontologies/tao.rdf#i64b112dc2090d85745707ce5a81032639": Property (http://www.tao.lu/Ontologies/TAOMedia.rdf#Link) of resource (https://pisa25-411-delete-old-assets-on-upload.docker.localhost/ontologies/tao.rdf#i64b112dc2090d85745707ce5a81032639) should not be empty [] {"uid":"9d0b0dd85c0861c0b0587edf"}
[2023-07-14 11:19:11] UDEMO.DEBUG: Event oat\taoItems\model\event\ItemRemovedEvent processed [] {"uid":"9d0b0dd85c0861c0b0587edf"}
```

gabrielfs7 (approved):

hectoras:
Released as [v47.2.0](https://github.com/oat-sa/extension-tao-testqti/releases/tag/v47.2.0).

PR #2394, opened 2023-07-12
fix: rename featyure flag accordingly to policy

bartlomiejmarszal:
Rename feature flag for XML edition

### DIFF
--- a/models/classes/xmlEditor/XmlEditor.php
+++ b/models/classes/xmlEditor/XmlEditor.php
@@ -30,7 +30,7 @@ use core_kernel_classes_Resource;
 
 class XmlEditor extends ConfigurableService implements XmlEditorInterface
 {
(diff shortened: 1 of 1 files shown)


models/classes/xmlEditor/XmlEditor.php
```diff
@@ -30,7 +30,7 @@
 
 class XmlEditor extends ConfigurableService implements XmlEditorInterface
 {
-    private const XML_EDITOR_ENABLED = 'XML_EDITOR_ENABLED';
+    private const XML_EDITOR_ENABLED = 'FEATURE_FLAG_XML_EDITOR_ENABLED';
```
Karol-Stelmaczonek, line 33: doesn't the const name also shouldn't b FEATURE_FLAG_XML_EDITOR_ENABLED ?

Karol-Stelmaczonek (approved):

PR #2421, opened 2023-11-27
feat: Pause previous deliveries when the test taker launches a new one

hectoras:
**Associated Jira issue:** [TR-5852](https://oat-sa.atlassian.net/browse/TR-5852), [RFE-748](https://oat-sa.atlassian.net/browse/RFE-748)

[TR-5852]: https://oat-sa.atlassian.net/browse/TR-5852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RFE-748]: https://oat-sa.atlassian.net/browse/RFE-748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


models/classes/runner/config/QtiRunnerConfig.php
```diff
@@ -147,6 +147,14 @@ protected function buildConfig()
...
         }
 
+        // Send a unique value so the frontend knows if the test has been reopened
+        // in another tab (that allows it to not modify the local storage from two
+        // different tabs)
+        //
+        $config['runNumberId'] = floor(microtime(true) * 100);
+
+        // @todo If the session is being restored, force restoreTimerFromClient to be true
+
```
hectoras, line 157: @js the FE does not need this value to be passed in the payload of the config endpoint, right?

It it's not needed, I'll remove changes in this file

hectoras, line 157:
```suggestion
```

jsconan: Correct, this is not needed anymore. The client-side implementation is taking care if this. See https://github.com/oat-sa/tao-test-runner-qti-fe/pull/522/files#diff-943b82aebd513800263e6f86920439485c546a4f7897a8bc65036d5ce0936f61R57

actions/class.Runner.php
```diff
@@ -71,6 +77,13 @@ class taoQtiTest_actions_Runner extends tao_actions_ServiceModule
 {
     use RunnerToolStates;
 
+    /**
+     * Controls whether launching a new delivery suspends other sessions by the same user.
+     *
+     * @var string
+     */
+    public const FEATURE_FLAG_PAUSE_CONCURRENT_SESSIONS = 'FEATURE_FLAG_PAUSE_CONCURRENT_SESSIONS';
```
gabrielfs7, line 85: We do not use constants for the feature flag itself, we just call its value. It is not necessary. Also, a controller/action is not a place to it. 

- [ ] Please document this feature flag on the README.md instead. Example: https://github.com/oat-sa/tao-core#environment-variables

jsconan (commented):
**Review Checklist**
- [x] New code respects code style rules
- [x] New code respects best practices
- [x] Commits are atomic

actions/class.Runner.php
gabrielfs7, line 85: Lets use it on README.md only. The constant is the value itself and we do not use for featureFlags.

```suggestion
```

model/Service/PauseService.php
```diff
@@ -30,6 +30,8 @@
 
 class PauseService
 {
+    public const PAUSE_REASON_CONCURRENT_TEST = 'PAUSE_REASON_CONCURRENT_TEST';
```
gabrielfs7, line 33: Is it still needed?

- [ ] remove if nod needed anymore - No usages in this very extension at least

actions/class.Runner.php
```diff
@@ -168,10 +181,11 @@ protected function getServiceContext()
             }
             $testDefinition = $container->getSourceTest($execution);
             $testCompilation = $container->getPrivateDirId($execution) . '|' . $container->getPublicDirId($execution);
+
             $this->serviceContext = $this->getRunnerService()->getServiceContext(
                 $testDefinition,
                 $testCompilation,
-                $testExecution
+                $execution->getIdentifier()
```
gabrielfs7, line 188: Before it was the `sessionId`, now it is the `deliveryExecution URI`. Wont it break compatibility with other places?

models/classes/runner/config/QtiRunnerConfig.php
hectoras: On it

hectoras: 3f527a61c6f881f0ff063427bd30720517f6aedf

hectoras: 3f527a61c6f881f0ff063427bd30720517f6aedf

actions/class.Runner.php
```diff
@@ -638,10 +685,10 @@ public function move()
 
             $response = $moveService($moveCommand);
 
-            common_Logger::d('Test session state : ' . $this->getServiceContext()->getTestSession()->getState());
-
             $this->returnJson($response->toArray());
         } catch (common_Exception $e) {
+            $this->checkExceptionForTestSessionConflict($e);
```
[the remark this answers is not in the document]
gabrielfs7: Humm, got it, the `returnJson` will exit the method. Forget it then :)

actions/class.Runner.php
```diff
@@ -970,6 +1018,35 @@ public function messages()
         $this->returnJson($response, $code);
     }
 
+    private function mustForceRestoreTimersFromClient(
+        QtiRunnerServiceContext $serviceContext,
+        $initResponse
+    ): bool {
```
gabrielfs7, line 1024: Typehint missing 
```suggestion
    private function mustForceRestoreTimersFromClient(
        QtiRunnerServiceContext $serviceContext,
        array $initResponse
    ): bool {
```

actions/class.Runner.php
```diff
@@ -58,7 +61,10 @@
 use oat\taoQtiTest\models\runner\QtiRunnerService;
 use oat\taoQtiTest\models\runner\QtiRunnerServiceContext;
 use oat\taoQtiTest\models\runner\RunnerToolStates;
+use oat\taoQtiTest\models\runner\session\TestSession;
 use oat\taoQtiTest\models\runner\StorageManager;
+use Psr\SimpleCache\CacheInterface;
```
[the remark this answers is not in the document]
hectoras: 75e59cb21

actions/class.Runner.php
```diff
@@ -970,6 +1018,35 @@ public function messages()
...
+            return false;
+        }
+
+        if (isset($initResponse['success']) && $initResponse['success']) {
+            $currentState = $serviceContext->getTestSession()->getState();
+
+            return ($currentState != AssessmentTestSessionState::INITIAL);
+        }
+
+        return false;
```
gabrielfs7, line 1030: Can be simplified by:

```suggestion
        return $this->isPausingConcurrentSessionsEnabled() 
            && isset($initResponse['success']) 
            && $initResponse['success'] 
            && $serviceContext->getTestSession()->getState() !== AssessmentTestSessionState::INITIAL;
```

actions/class.Runner.php
```diff
@@ -1168,4 +1245,18 @@ private function createErrorResponseFromException(Exception $exception): void
             $this->getStatusCodeFromException($exception)
         );
     }
+
+    private function checkExceptionForTestSessionConflict($exception): void
```
gabrielfs7, line 1249: Missing typehint 

```suggestion
    private function checkExceptionForTestSessionConflict(Throwable $exception): void
```

actions/class.Runner.php
```diff
@@ -1168,4 +1245,18 @@ private function createErrorResponseFromException(Exception $exception): void
...
+    {
+        if (!$exception instanceof common_exception_Unauthorized) {
+            return;
+        }
+
+        try {
+            if ($this->getServiceContext()->getTestSession()->getState() === AssessmentTestSessionState::INTERACTING) {
+                $this->setSessionAttribute('testSessionConflict', true);
+            }
+        } catch (common_Exception $exception) {
```
gabrielfs7, line 1259: Why we silent it?

actions/class.Runner.php
hectoras: Hmm.. the `FeatureFlagCheckerInterface` itself uses this approach to expose some feature flags publicly, so I used the same approach here. ltiDeliveryProvider is also using this technique to expose `FEATURE_FLAG_MAINTAIN_RESTARTED_DELIVERY_EXECUTION_STATE` ([here](https://github.com/oat-sa/extension-tao-ltideliveryprovider/blob/master/controller/DeliveryTool.php)).

I don't have a strong preference in this case, and the constant is currently just used once, so I'm removing this const and adding it to the README instead.

gabrielfs7: Yes, indeed about `FeatureFlagCheckerInterface` it uses, but the problem is that it is not sustainable as it is on tao-core and we should document it on README.md instead. As you can see, not all FeatureFlags are there and they vary from extension to extension. More visible as a config in README.md. 

Thanks!

hectoras: Added to the README: 1ed567a3ea9055015ee2f1b57076def513eb7bb9

model/Service/PauseService.php
hectoras: This is used by `ltiDeliveryProvider` to store why a delivery has been paused, and then from the feedback screen (in  the same extension) to known which message to show (just _paused_ vs _paused because a concurrent test_). Additionally, it is used similarly from the `ltiTestReview` extension (to indicate there is a concurrent session, in this case the "review session").

It could be moved to `ltiDeliveryProvider` (since the only other usage is from `ltiTestReview`, and that already declares a dependency on `ltiDeliveryProvider`), but I added this here because I considered the reason for pausing  a test part of the model of the `(Qti)Test` itself.

actions/class.Runner.php
hectoras: The docblock for `getServiceContext()` states it expects _The URI of the delivery execution_ for this parameter, so I think this is correct. However, I don't remember anymore the reason for this change (maybe there was something not working), I'll recheck.

actions/class.Runner.php
```diff
@@ -296,8 +298,17 @@ public function init()
...
-            $serviceContext = $this->getRunnerService()->initServiceContext($this->getServiceContext());
-            $this->returnJson($this->getInitResponse($serviceContext));
+            $serviceContext = $this->getRunnerService()->initServiceContext(
+                $this->getServiceContext()
+            );
+
+            $initResponse = $this->getInitResponse($serviceContext);
+
+            if ($this->mustForceRestoreTimersFromClient($serviceContext, $initResponse)) {
+                $initResponse['testData']['config']['timer']['restoreTimerFromClient'] = true;
```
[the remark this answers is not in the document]
jsconan: **issue:** Good catch! I did not read it well in the first review. Actually, the idea is to prevent the restoration from the client when the delivery is force-paused due to concurrency... The correct value for such a case would rather be `false` when a concurrency is detected. The condition is: if another delivery is executed, force-pause all existing delivery executions. Then, if the paused delivery is resumed, prevent the client from restoring the client time since it may have increased meanwhile. However, this is needed only if the delivery countdown was not frozen by the client, which would be unlikely to happen if the `preventConcurrency` plugin is active. In the case of server-side timer, this configuration would not matter that much since we don't care about client time.

actions/class.Runner.php
gabrielfs7: Indeed it was and as you can see [here](https://github.com/oat-sa/extension-tao-testqti/pull/2421/files#diff-fea95e24360d827ee2d0e75d5c4d18bc5d43134b522acb2f205ad18f7be19a67L152) it uses the sessionId to retrieve the execution, so it seems they are the same thing. 

Although, if this change is not required, I would not touch it to avoid possible issues.

hectoras: Indeed, both take the same value:

```
[2023-11-27 15:50:06] tao.CRITICAL: testExecution: http://backoffice.docker.localhost/ontologies/tao.rdf#i6564ac934f79397b0c319ed0b1335e execution.identifier: http://backoffice.docker.localhost/ontologies/tao.rdf#i6564ac934f79397b0c319ed0b1335e [] []
```

I'll rollback to get the value as we did previously.

hectoras: Rolled back

model/Service/PauseService.php
gabrielfs7: Got it. The thing is that what we pause is a delivery execution, not the `QTI test`. I think the boundaries here are already overlapping a bit.

Also what is there on test review, probably could go to another extension (or this one) https://github.com/oat-sa/extension-lti-test-review/pull/135#discussion_r1406106289

I do not have a strong opinion on this as ltiDeliveryProvider, taoDelivery, taoDeliveryRdf, testQti, ltiTestReview, etc are kind of talking about similar thing but segregated. 

So, I accept suggestions on this one. If it is already in use, we can keep it. And if this is part of the parameters used by `PauseService`, it is ok to be here.

actions/class.Runner.php
gabrielfs7: Thanks!

actions/class.Runner.php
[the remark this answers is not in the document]
gabrielfs7: cc: @hectoras

gabrielfs7 (approved):

PR #2432, opened 2024-01-02
bugfix: move event to proper delivery finish

Karol-Stelmaczonek:
Ticket: https://oat-sa.atlassian.net/browse/INF-248

## What's Changed

Moving the event to the end of Delivery Execution gives us access to Outcomes and DE Finish time at the same time. We need both of these values to cover the case when a query is sent via the AGS API without changing the result and when the test is set with Outcomes is set to none

## Dependencies PRs

- https://github.com/oat-sa/extension-tao-ltideliveryprovider/pull/371
- https://github.com/oat-sa/extension-tao-outcome/pull/273

PR #2435, opened 2024-01-08
Bugfix/inf 248 not send ags with outcome processing set to none

Karol-Stelmaczonek:
Proper PR to develop, old reverted PR https://github.com/oat-sa/extension-tao-testqti/pull/2432

Ticket: https://oat-sa.atlassian.net/browse/INF-248

## What's Changed

Moving the event to the end of Delivery Execution gives us access to Outcomes and DE Finish time at the same time. We need both of these values to cover the case when a query is sent via the AGS API without changing the result and when the test is set with Outcomes is set to none

## Dependencies PRs

- https://github.com/oat-sa/extension-tao-ltideliveryprovider/pull/371
- https://github.com/oat-sa/extension-tao-outcome/pull/273

### DIFF
--- a/helpers/class.TestSession.php
+++ b/helpers/class.TestSession.php
@@ -819,12 +819,11 @@ class taoQtiTest_helpers_TestSession extends AssessmentTestSession
             $this->getSessionId(),
             $variables,
             $this->getSessionId(),
-            $testUri,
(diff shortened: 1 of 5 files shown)


pribi (approved):
LGTM, previous pr was already checked.

PR #2442, opened 2024-02-02
fix: transform method after lodash update

taopkorczak:
Related: https://oat-sa.atlassian.net/browse/ADF-1579

Fix transform method


oatymart (approved):
Yes, this version passes tests and can also be bundled.

PR #2456, opened 2024-04-12
fix: Improve opening another tab with new session with the same user

tikhanovichA:
https://oat-sa.atlassian.net/browse/INF-266

For cases when one of sessions was finished and then reopened, the pausing for another tabs worked not like expected.

How to test:

1. Start new test A (new LTI link)
2. Start new test B (new LTI link)
3. A gets the “concurrent test” popup. Do not click on it
4. Finish test B
5. Check that you have “Thank you” page
6. Go back to test A & Accept the popup
7. Check that you have “Test suspended” page
8. Start the previous test A again (reusing the previous LTI link). This unpauses A
9. Start the previous test B again (reusing the previous LTI link). This pauses A
10. Go back to test A & Accept the popup
11. You should receive the “Test suspended” page

https://github.com/oat-sa/extension-tao-testqti/assets/1053022/de422b1d-26e6-40a0-b680-42acfddf6863

### DIFF
--- a/model/Service/ConcurringSessionService.php
+++ b/model/Service/ConcurringSessionService.php
@@ -24,10 +24,12 @@ namespace oat\taoQtiTest\model\Service;
 
 use common_Exception;
 use oat\oatbox\service\ServiceManager;
+use oat\tao\model\featureFlag\FeatureFlagChecker;
 use oat\tao\model\featureFlag\FeatureFlagCheckerInterface;
 use oat\taoDelivery\model\execution\DeliveryExecution;
 use oat\taoDelivery\model\execution\DeliveryExecutionInterface;
 use oat\taoDelivery\model\execution\DeliveryExecutionService;
+use oat\taoDelivery\model\execution\StateServiceInterface;
 use oat\taoDelivery\model\RuntimeService;
 use oat\taoQtiTest\models\container\QtiTestDeliveryContainer;
 use oat\taoQtiTest\models\runner\QtiRunnerService;
@@ -35,6 +37,7 @@ use oat\taoQtiTest\models\runner\QtiRunnerServiceContext;
 use oat\taoQtiTest\models\runner\time\TimerAdjustmentService;
 use oat\taoQtiTest\models\TestSessionService;
 use PHPSession;
+use Psr\Container\ContainerInterface;
 use Psr\Log\LoggerInterface;
 use qtism\common\datatypes\QtiDuration;
 use qtism\data\AssessmentItemRef;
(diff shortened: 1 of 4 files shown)


model/Service/ConcurringSessionService.php
```diff
@@ -67,6 +77,21 @@ public function __construct(
...
+            $this->getConcurringSessionService()->pauseConcurrentSessions($activeExecution);
+
+            if ($activeExecution->getState()->getUri() === DeliveryExecution::STATE_PAUSED) {
+                $this->getConcurringSessionService()->adjustTimers($activeExecution);
+            }
+
+            $this->getConcurringSessionService()->clearConcurringSession($activeExecution);
+        }
+
+        $this->resetDeliveryExecutionState($activeExecution);
```
hectoras, line 92: `resetDeliveryExecutionState` does nothing if `$activeExecution` is `null`, moving that method call inside the conditional body may be a bit more clear.

```suggestion
        if ($activeExecution instanceof DeliveryExecution) {
            $this->getConcurringSessionService()->pauseConcurrentSessions($activeExecution);

            if ($activeExecution->getState()->getUri() === DeliveryExecution::STATE_PAUSED) {
                $this->getConcurringSessionService()->adjustTimers($activeExecution);
            }

            $this->getConcurringSessionService()->clearConcurringSession($activeExecution);
            $this->resetDeliveryExecutionState($activeExecution);
        }
```

model/Service/ConcurringSessionService.php
```diff
@@ -67,6 +77,20 @@ public function __construct(
         $this->currentSession = $currentSession ?? PHPSession::singleton();
     }
 
+    public function pauseActiveDeliveryExecution($activeExecution): void
```
hectoras, line 80: suggestion: I think this actually pauses _all_ delivery executions for a given user

- `pauseConcurrentSessions()` pauses all executions other than `$activeExecution` in `STATE_ACTIVE` for the current user (sets them to `STATE_PAUSED`)
- `resetDeliveryExecutionState()` pauses `$activeExecution` (calling [StateService::pause()](https://github.com/oat-sa/extension-tao-delivery/blob/master/model/execution/StateService.php#L66) at line 231 of this file).

Then, when `$activeExecution` starts, `DeliveryServer::runDeliveryExecution()` calls ([here](https://github.com/oat-sa/extension-tao-delivery/blob/master/controller/DeliveryServer.php#L268)) [StateService::run()](https://github.com/oat-sa/extension-tao-delivery/blob/master/model/execution/StateService.php#L56), which sets the state back to `STATE_ACTIVE`.

I think it would make sense to rename this to something like `pauseActiveDeliveryExecutionsForUser()` to make it clear that it does not pause just the execution passed as an arg but also all other concurrent executions.

hectoras (approved):

gabrielfs7 (approved):
Thanks @tikhanovichA 

If possible, can we add a test case with the new code implement?

PR #2461, opened 2024-04-16
Feature/AUT-3414/Align categories that activate the TextToSpeech tool

jsconan:
### Related to: https://oat-sa.atlassian.net/browse/AUT-3414

Requires: 
 - [ ] https://github.com/oat-sa/extension-tao-texthelp/pull/70

### Summary

Add support for multiple categories to activate presets in the test authoring.
The concepts is as follows:
- a preset can be activated using any of the declared categories (support for legacy categories)
- when saving the test, only the main declared category of the preset is persisted

The main example for this is the text-to-speech tool which used to be activated using the category `x-tao-option-textToSpeech`. However, TAO Advance expects it to be `x-tao-option-tts`. Both categories will be supported, but only the later will be kept when updating a test.

### How to test
Prerequisite: have the extension taoTextHelp installed with the branch `feature/AUT-3414/align-tts-category`

- create/open a test in Authoring
- activate the test taker tool TextToSpeech
- save the changes
- export the test, then open the archive an check that the category `x-tao-option-textToSpeech` is set appropriately
- check out the branch on both taoQtiTest and taoTextHelp: `git checkout -t origin/feature/AUT-3414/align-tts-category`
- update TAO
- open the same test in Authoring
- check that the taker tool TextToSpeech is still activated
- deactivate the tool, then activate it again
- save the changes
- export the test, then open the archive an check that the category `x-tao-option-tts` is set appropriately

### DIFF
--- a/models/classes/TestCategoryPreset.php
+++ b/models/classes/TestCategoryPreset.php
@@ -49,7 +49,7 @@ class TestCategoryPreset implements JsonSerializable
     /**
      * @var string[] $altCategories - the other possible qti categories that would activate the preset
      */
-    private $altCategories;
+    private $altCategories = [];
 
     /**
      * @var string $description - what is the category purpose
@@ -97,8 +97,6 @@ class TestCategoryPreset implements JsonSerializable
         $this->label        = (string) $label;
         $this->qtiCategory  = (string) $qtiCategory;
 
-        $this->altCategories = [];
-
         if (isset($data['description'])) {
             $this->description = (string) $data['description'];
         }
@@ -136,7 +134,7 @@ class TestCategoryPreset implements JsonSerializable
         return $this->altCategories;
     }
 
-    public function getDescription(): ?string
+    public function getDescription(): string
     {
         return $this->description;
     }
--- a/models/classes/TestCategoryPreset.php
+++ b/models/classes/TestCategoryPreset.php
@@ -31,57 +31,45 @@ use common_exception_InconsistentData;
  */
 class TestCategoryPreset implements JsonSerializable
 {
-    /**
-     * @var string $id
-     */
-    private $id;
+    private string $id;
 
     /**
-     * @var string $label - short preset name
+     * Short preset name
      */
-    private $label;
+    private string $label;
 
     /**
-     * @var string $qtiCategory - the actual qti category that will end up in the QTI markup
+     * The actual qti category that will end up in the QTI markup
      */
-    private $qtiCategory;
+    private string $qtiCategory;
 
     /**
-     * @var string[] $altCategories - the other possible qti categories that would activate the preset
+     * The other possible qti categories that would activate the preset
      */
-    private $altCategories = [];
+    private array $altCategories = [];
 
     /**
-     * @var string $description - what is the category purpose
+     * What is the category purpose
      */
-    private $description = '';
+    private string $description = '';
 
     /**
-     * @var string $order - to sort the categories
+     * To sort the categories
      */
-    private $order = 0;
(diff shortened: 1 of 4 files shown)


models/classes/TestCategoryPreset.php
```diff
@@ -46,6 +46,11 @@ class TestCategoryPreset implements JsonSerializable
      */
     private $qtiCategory;
 
+    /**
+     * @var string[] $altCategories - the other possible qti categories that would activate the preset
+     */
+    private $altCategories;
```
gabrielfs7, line 52: chore: There is no need to initialize it inside constructor https://github.com/oat-sa/extension-tao-testqti/pull/2461/files#diff-c5577ec1fd5d99becf2d82ed6ffc21e5f0cd54abfe2be7f978c690139257a473R100

```suggestion
    private $altCategories = [];
```

gabrielfs7: Not necessary to refactor, but since PHP 7.4 we can declare properties with types.

models/classes/TestCategoryPreset.php
```diff
@@ -172,7 +188,7 @@ public function toArray()
      * @return TestCategoryPreset the new instance
      * @throws common_exception_InconsistentData
      */
-    public static function fromArray(array $data)
+    public static function fromArray(array $data): TestCategoryPreset
```
gabrielfs7, line 189: chore: totally optional and just sharing information. Since we have typehint + return type, there is no need for PHPDocs like `@param` or `@return`. It would also not break PHP Doc generation if you would remove it. We generally remove it as it is redundant and can make the code smaller. 

I good reason to keep is if you want to document these like `@param array $data Some info here`, but if not, you can remove safely

models/classes/TestCategoryPreset.php
```diff
@@ -104,47 +111,55 @@ public function __construct($id, $label, $qtiCategory, $data)
...
     }
 
-    public function getPluginId()
+    public function getPluginId(): string
     {
         return $this->pluginId;
     }
 
-    public function getFeatureFlag()
+    public function getFeatureFlag(): string
```
gabrielfs7, line 152: @jsconan as far as I see in the constructor this is optional, so it needs to be a bit different https://github.com/oat-sa/extension-tao-testqti/pull/2461/files#diff-c5577ec1fd5d99becf2d82ed6ffc21e5f0cd54abfe2be7f978c690139257a473R111

```suggestion
    public function getFeatureFlag(): ?string
```

jsconan: Yes, I saw it, but the type is forced to string so that it always return with a string. Or do you think I should change the return instruction as well? If possible I'd prefer to not change too much the code I was not supposed to touch unless the change is harmless, and for this one I don't now the impact.

gabrielfs7: IMO if it is optional, it must be null in case of not provided. It helps on validation and it is more consistent IMO.

gabrielfs7 (approved):
Thanks for improving the code, @jsconan 

I did a check on PHP code level only and looks good. We can have a demo to check functionality if possible.

- [x] New code is covered by tests (if applicable) JS code is

viktar-dzmitryieu-tao (approved):
- [X] New code is covered by tests (if applicable)
- [X] New code is respecting code style rules
- [X] New code is respecting best practices
- [X] Acceptance criteria are respected
- [X] Pull request title and description are meaningful
- [X] Pull request's target is not `master`

shaveko (approved):
Looks good!

PR #2469, opened 2024-05-16
fix: Don't fetch the current item for a null test session

hectoras:
**Associated Jira issue:** [INF-268](https://oat-sa.atlassian.net/browse/INF-268)

Fixes a race condition that happens when the same test taker launches two executions concurrently, causing the logic to pause the timers for the first one that got processed to try to fetch a test session that may not have been persisted yet.

**Changelog:**
- Don't call `getCurrentAssessmentItemRef()` on `null` when an execution does not have a test session assigned (yet)
  - If there's no test session, there's no current item nor item duration to adjust
- Don't try to `suspend()` the same delivery execution twice

This fixes a customer-reported error, and has already been validated on their side. The error was:

```
/var/www/html/taoQtiTest/model/Service/ConcurringSessionService.php@146:
   Uncaught Error: Call to a member function getCurrentAssessmentItemRef() on null in
     /var/www/html/taoQtiTest/model/Service/ConcurringSessionService.php:146

Stack trace:
#0 /var/www/html/ltiDeliveryProvider/controller/DeliveryTool.php(145):
    oat\taoQtiTest\model\Service\ConcurringSessionService->adjustTimers(
        Object(oat\taoDelivery\model\execution\DeliveryExecution)
    )
#1 [internal function]: oat\ltiDeliveryProvider\controller\DeliveryTool->run() 
...
```

[INF-268]: https://oat-sa.atlassian.net/browse/INF-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/model/Service/ConcurringSessionService.php
+++ b/model/Service/ConcurringSessionService.php
@@ -23,7 +23,6 @@ declare(strict_types=1);
 namespace oat\taoQtiTest\model\Service;
 
 use common_Exception;
-use oat\oatbox\service\ServiceManager;
 use oat\tao\model\featureFlag\FeatureFlagCheckerInterface;
 use oat\taoDelivery\model\execution\DeliveryExecution;
 use oat\taoDelivery\model\execution\DeliveryExecutionInterface;
@@ -44,6 +43,7 @@ use Throwable;
 class ConcurringSessionService
 {
     private const PAUSE_REASON_CONCURRENT_TEST = 'PAUSE_REASON_CONCURRENT_TEST';
+
     /**
      * @var string Controls whether a delivery execution state should be kept as is or reset each time it starts.
      *             `false` – the state will be reset on each restart.
@@ -58,9 +58,9 @@ class ConcurringSessionService
     private DeliveryExecutionService $deliveryExecutionService;
     private FeatureFlagCheckerInterface $featureFlagChecker;
     private StateServiceInterface $stateService;
+    private TestSessionService $testSessionService;
+    private TimerAdjustmentServiceInterface $timerAdjustmentService;
     private ?PHPSession $currentSession;
-    private ?TestSessionService $testSessionService;
-    private ?TimerAdjustmentServiceInterface $timerAdjustmentService;
 
     public function __construct(
         LoggerInterface $logger,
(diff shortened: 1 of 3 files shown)


model/Service/ConcurringSessionService.php
```diff
@@ -298,20 +308,27 @@ private function getContextByDeliveryExecution(DeliveryExecutionInterface $execu
...
+
+        return $this->timerAdjustmentService;
     }
 
     private function getTestSessionService(): TestSessionService
     {
-        /** @noinspection PhpIncompatibleReturnTypeInspection */
-        return $this->getServiceManager()->get(TestSessionService::SERVICE_ID);
-    }
+        if (!$this->testSessionService instanceof TestSessionService) {
```
gabrielfs7, line 325: Why cant we rely on DI container? I know that `$this->testSessionService` is nullable, but this is decided in the DI Container definition files, not here. I would remove completely the `ServiceManager` usage from this class. Wdyt?

gabrielfs7: I know we did it here, but maybe we can avoid that hybrid approach? 

https://github.com/oat-sa/extension-tao-testqti/commit/d66b9e501cedf0fd7978278e9787ae7e250202ed#diff-be1605fd62af343d118c742ba9c27603c98978aeff03b929a54110c9ba97cda6R256-R266

hectoras: Makes sense, changed here: 024d5dd01

model/Service/ConcurringSessionService.php
```diff
@@ -67,15 +69,19 @@ public function __construct(
         DeliveryExecutionService $deliveryExecutionService,
         FeatureFlagCheckerInterface $featureFlagChecker,
         StateServiceInterface $stateService,
+        TestSessionService $testSessionService,
```
gabrielfs7, line 72: - [ ] Can you please make sure there is no other DI Service provider instantiating this class apart of `TestQtiServiceProvider`? Thanks!

model/Service/ConcurringSessionService.php
gabrielfs7: Thanks!

gabrielfs7 (approved):
LGTM, Code looks good, but I cannot check it. 

shpran (approved):

model/Service/ConcurringSessionService.php
hectoras: I've just checked and it is only used in the `TestQtiServiceProvider` from this extension, and all instantiations at runtime are done though the PSR container.

hectoras: image

hectoras:
These changes have two approvals and the fix was validated by QA, therefore I'm merging this PR.

PR #2516, opened 2024-10-16
Feature/ADF-1789/Forward the translation parameters

jsconan:
### Summary

Forward the translation parameters to the authoring.

### Details

This is a technical change only. It adds support for the translation mode, forwarding:
- `translation`: `bool` - A flag telling if we are translating a test. Can be `null`.
- `originResourceUri`: `string` - The URI of the original test being translated. Can be `null`.

From the client app, the data is available at the root of the received config from the controller.

### DIFF
--- a/actions/class.Creator.php
+++ b/actions/class.Creator.php
@@ -1,22 +1,22 @@
 <?php
 
 /**
-* This program is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


shaveko (approved):
Looks good!

PR #2519, opened 2024-10-29
fix: support test items in subsections for translations

gabrielfs7:
# Goal 

Make the items in nested subsections to be translated


jsconan (approved):
Issue fixed

shpran (approved):

PR #2525, opened 2024-11-06
fix: unique identifiers

shpran:
Fix for unique identifiers + removing unnecessary classes


models/classes/UniqueId/Service/UniqueIdSetter.php
```diff
@@ -0,0 +1,88 @@
...
+use oat\generis\model\data\Ontology;
+use oat\tao\model\featureFlag\FeatureFlagCheckerInterface;
+use oat\tao\model\TaoOntology;
+use oat\tao\model\Translation\Service\ResourceMetadataRetriever;
+use oat\tao\model\UniqueId\Service\ResourceUniqueIdRetriever;
+use Psr\Log\LoggerInterface;
+use taoQtiTest_models_classes_QtiTestService;
+use Throwable;
+
+class UniqueIdSetter
```
gabrielfs7, line 35: - [ ] This one seems exactly the same as for Items. Why did not we create it in tao-core

PR #2606, opened 2025-05-13
Feat/aut 4186/load scale remote list to creator

bartlomiejmarszal:
Load Remote List Scale in Authoring


actions/class.Creator.php
```diff
@@ -86,8 +89,24 @@ public function index()
...
+            $this->setData(
+                'scalesPresets',
+                json_encode(
+                    iterator_to_array(
+                        $this->getRemoteListService()->getListElements(
+                            $this->getClass(RemoteScaleListService::SCALES_URI)
+                        )
+                    )
+                )
+            );
```
Karol-Stelmaczonek, line 105: Ok this can fail on so many levels, can you split it to part where you build `$scalePresets` and handle all possible problems like empty list elements? Then do` json_encode` and also handle json errors?

PR #2609, opened 2025-05-22
feat: backend pass scale presets

bartlomiejmarszal:
# What changed?
Pass Scale Presets from remote lists service to frontend

### DIFF
--- a/actions/class.Creator.php
+++ b/actions/class.Creator.php
@@ -18,6 +18,9 @@
  * Copyright (c) 2013-2024 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
  */
 
+use oat\generis\model\OntologyAwareTrait;
+use oat\taoBackOffice\model\lists\ListService;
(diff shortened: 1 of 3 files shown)


actions/class.Creator.php
```diff
@@ -215,4 +228,46 @@ protected function getRuntimeConfig()
...
     }
+
+    private function getRemoteListService(): ListService
+    {
+        return $this->getServiceManager()->getContainer()->get(ListService::class);
+    }
+
+    private function isScaleEnabled(): bool
+    {
+        return $this->getRemoteScaleListService()->isRemoteListEnabled();
```
Karol-Stelmaczonek, line 239: It will not be hidden behind FF? There will be no situation where we would want to set the remote list and still not show scales? Just curious.

Karol-Stelmaczonek (approved):

actions/class.Creator.php
bartlomiejmarszal: This is a feature flag. If we will not set it up remote list should not be created.

PR #2786, opened 2026-05-21
fix/INV-1203: regenerate text reader images when asset updated

jebersole:
## Ticket:
https://oat-sa.atlassian.net/browse/INV-1203

## What's Changed
Fixes an issue where text reader images wouldn't update after their corresponding assets were changed.

## Dependencies PRs
- https://github.com/oat-sa/extension-tao-itemqti/pull/3005
- https://github.com/oat-sa/tao-deliver-be/pull/1689
- https://github.com/oat-sa/extension-tao-mediamanager/pull/567

## How to test
1. Upload an image to the Assets section in backoffice.
2. Create a Text Reader interaction from Custom Interactions and insert this image asset from the toolbar.
3. Create a test with this item, and publish the test.
4. Click Replace Asset and upload a different image for this image in Assets.
5. The image should now be replaced in item and test previews, and in deliveries after republishing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Text reader pages now correctly replace embedded image sources with their resolved content URLs so linked images display as expected.
  * Improved HTML parsing and re-serialization to handle edge cases, preserve page content structure, and avoid broken or missing images.
  * Increased reliability of page content rendering when pages include mixed or serialized content formats, reducing display inconsistencies.

### DIFF
--- a/models/classes/render/CustomInteraction/PostProcessor/TextReaderPostProcessor.php
+++ b/models/classes/render/CustomInteraction/PostProcessor/TextReaderPostProcessor.php
@@ -97,17 +97,11 @@ class TextReaderPostProcessor implements CustomInteractionPostProcessorInterface
 
     private function normalizePages($pages): array
     {
-        if (is_array($pages)) {
-            return $pages;
+        try {
+            return json_decode($pages, true, 512, JSON_THROW_ON_ERROR);
+        } catch (\JsonException $_) {
+            return is_array($pages) ? $pages : [];
         }
-
-        if (!is_string($pages) || $pages === '') {
-            return [];
-        }
-
-        $decoded = json_decode($pages, true);
-
-        return is_array($decoded) ? $decoded : [];
     }
 
     private function extractContentUrls(array $properties): array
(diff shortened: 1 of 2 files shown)


models/classes/render/CustomInteraction/PostProcessor/TextReaderPostProcessor.php
```diff
@@ -45,7 +46,119 @@ public function postProcess(array $element): array
...
+        }
+
+        $properties['pages'] = is_string($properties['pages'])
+            ? json_encode($updatedPages)
+            : $updatedPages;
+
+        return $properties;
+    }
+
+    private function normalizePages($pages): array
```
kilatib, line 98: This method looks for me like an extra one for delivering here may we simplify it in one string and use at line 62? 

Something like 
try {
   return json_decode($pages, true, 512, JSON_THROW_ON_ERROR);
} catch (\JsonException $e) {
     return is_array($pages) ? $pages : [];
}

kilatib (changes requested):
The realization that and on delivery seems to me like a little havy for those simple things. Maybe we could trust a few variables there and make it simple WDYT?

kilatib (approved):
LGTM

PR #2792, opened 2026-06-17
fix: exclude checking outcomes from saving translations

tikhanovichA:
## Ticket:

https://oat-sa.atlassian.net/browse/AUT-4617

## What's Changed
- Skip scoring-change when saving a translation if only translatable metadata changed (section titles, rubrics) - translations shouldn’t need outcome reprocessing.
- Fixed the overlapping header buttons in translation mode.

Before the fix:

https://github.com/user-attachments/assets/7b135f17-676b-4ae4-9759-a4329ba696cb

After the fix:

https://github.com/user-attachments/assets/85f43019-2d43-4e07-b2f4-54b1691e37e2

## How to test
- Create a test with needed outcome processing
- Create a translation
- Press save on the transaltion page 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scale recognition when a persisted scale file reference is already present.
  * Save operations now emit `scoring-change` only in non-translation scenarios.

* **Style**
  * Updated the side-by-side authoring editor menu layout to use a flexbox row.

* **Tests**
  * Added unit coverage to ensure the scales directory is preserved when a saved scale reference is used.
  * Added QUnit coverage for the conditional `scoring-change` behavior (with a dedicated test runner page).


bartlomiejmarszal (approved):

KirylHatalski (approved):

views/js/controller/creator/creator.js
```diff
@@ -333,7 +333,9 @@ define([
                                 creatorContext.on('save', function () {
                                     if (!$saver.hasClass('disabled')) {
                                         $saver.prop('disabled', true).addClass('disabled');
-                                        modelOverseer.trigger('scoring-change');
+                                        if (!options.translation) {
```
Karol-Stelmaczonek: What if this flag is true but we also have scoring changes? Is that possible? With translation and scoring needed to be save we will end with only translations persisted.

tikhanovichA: It's not even possible on the translation page to update outcomes

gabrielfs7: Today there should not be any scoring changes possible to be applied in translations. It was never supported. 

In the translations we can only change labels and titles, nothing more. The rest must remain an snapshot of the original test (in its original state)

gabrielfs7: - [x] @tikhanovichA can we add a comment here with a big `warning` stating that translation cannot have any outcome variable changed? I think this will help future developers to not do it by mistake. 

Thanks!

Karol-Stelmaczonek (approved):

views/js/controller/creator/creator.js
tikhanovichA: will do

tikhanovichA: Done 
https://github.com/oat-sa/extension-tao-testqti/pull/2792/changes#diff-6434a6e92f888f70da6de22dfc258dce6bb45976b2b2538cc9c7ea8105b847faR339-R345

gabrielfs7 (approved):

